Issue #408, opened 2017-02-28
Semantic Checker Throws Errors On Loading a Graph


NAnguiano:
I didn't see any issue that this would fit into, so I'm just making a new one. 

The semantic checker is having some issues right now and currently will not run on the beta server in its current state. When I went to refresh the beta server this morning, the semantic checker crashed the server. @dondi and I hotfixed a few random issues that were found, but upon loading a file, this happened:
![screen shot 2017-02-27 at 6 39 51 pm](https://cloud.githubusercontent.com/assets/6444156/23389326/35487de4-fd1c-11e6-9b4e-665e7b9cf21c.png)

I haven't tracked down why this is occurring, nor whether or not it will occur when SIF and GraphML are added to the check, as per issue #386. 

As a result, the semantic checker is currently turned off on beta. 

Another random note:
The version of node being run on the beta server is extremely old: 0.10. @dondi is going to upgrade it eventually (issue #407), but for now, there's many things that simply just don't work when they're pushed to the server. Most notably, `for of` (ex: `for (var potato of potatoArray) {`), will not work on the current version. This would not show up in testing, and @dondi and I were not aware of it until now. If you're using `for of`, it's safest to remove it for now or replace it with `for in`, as it will cause the server to crash when a file is run.

kdahlquist:
@NAnguiano, is your "random note" above related to this new issue or is it something separate?  If it is separate, please make a new issue for it so that it doesn't get buried.  Thanks.

NAnguiano:
It's related insofar that it was significantly contributing to the initial problems we had getting beta running. I'll make a new issue for it anyways though, since it is separate overall.

yshin4:
Changed for of loop, which was causing the problem, into a traditional for loop. Also fixed the typo in the semantic checker. @NAnguiano and @yshin4 can confirm that the demo server is running without the problems proposed above. All the errors passed the test as well.

kdahlquist:
I verified that v2.0.2 is running on the beta page this morning.  If that's all I need to do to check this one, it's ready for closing.  If there's something else I should do, let me know, otherwise go ahead and close this one.

NAnguiano:
Alright. Going ahead and closing this one then!

Issue #397, opened 2017-02-13
Implement preference to change viewport size between small, medium, large


kdahlquist:
There will be three size options: 

- small is for 13-15 " laptop size
- medium is for best fit on @kdahlquist's screen (~24" monitor)
- large is larger than that

kdahlquist:
Medium, according to @kdahlquist's screen using Firefox:

Current beta is 14.5 inches wide, should go to 15 inches.
Current beta is 8.25 inches wide, should go to 9 inches.

kdahlquist:
small is a little too big.  The height is OK, but the width needs to shrink by 1/2 inch.

kdahlquist:
We noted at our meeting that final tweaks to the sizes will be made after UI changes are made.  So, make these changes now and we'll revisit it later.

NAnguiano:
I tweaked the sizes as requested.

kdahlquist:
I can't check the sizes specifically right now, but I can verify that choosing the radio buttons changes the size.  I also note that changing the size causes a reload, which is something we don't want to have happen.

NAnguiano:
It does currently trigger a reload. This should be fixed once I get the adaptive drafting board fully implemented (#398), as at that point the code will allow for the graph to change size without reloading. Until I get that completed, a reload will be required. Without a reload, the svg doesn't see the new size of the drafting board, and will continue to operate as if the size had not changed.

NAnguiano:
The reloading issue is fixed on my fork.

kdahlquist:
Making this a review requested and assigning to me.  Since this is related to #399, #396, #398, as soon as pull request #415 has been taken care of and @NAnguiano can make this live on beta, I will review them as a batch and close.

kdahlquist:
I am going to close this one since the buttons are there and they do indeed switch the viewport between small, medium, and large.

I think we need to fully explore all the interactions between choosing viewport size and allowing adaptive vs. fixed drafting board size, but let's open specific issues for that as they arise.

Issue #395, opened 2017-02-13
Implement automatic discovery of initial drafting board and viewport size


kdahlquist:
GRNsight web app should set its initial drafting board and viewport size to the best fit (among our three fixed sizes - Issue #397) based on the current iframe dimensions and zoom factor (scale is at 1). I.e., world starts with drafting board and viewport aligned. The iFrame should determine its size based on the amount of available space on the user's screen.

 Note that the viewport does not change after this point.

kdahlquist:
I just wanted to note that if the _automatic_ part of the automatic discovery gets us bogged down, I would be OK with it not being automatic and moving on to the meat of the functionality.  I can imagine that this might be browser-dependent and get hairy.

NAnguiano:
Looking at this, it looks like the web page and the iFrame are going to need to be able to communicate with each other in order for this to become automatic. They can't right now, but I'm working on getting this... well, working. 

I'm going off of this example on stackoverflow currently. 
http://stackoverflow.com/questions/9153445/how-to-communicate-between-iframe-and-the-parent-site

NAnguiano:
@dondi pointed out that I was probably overthinking this.

He was right.

This is now working properly. I've made a few changes to the CSS of the beta page so that it looks better when viewing the automatic sizing. The most notable change is that I significantly increased the height of the iframe. This allows for even the largest size to be viewable in one piece without vertical scrolling. I would like to make it so that the containing div expands to the height of its iframe's content, but I can't seem to figure out how to do that at this time. This is a temporary workaround.

kdahlquist:
Note that on @kdahlquist's laptop, using Firefox at 100%, it chose "medium" and not "small" as expected.

NAnguiano:
This is working with my current method, but the method I'm using isn't the best. I'll be switching it to use [media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries), which [Bootstrap](https://v4-alpha.getbootstrap.com/layout/overview/#responsive-breakpoints) also uses.

When it comes to expanding the div to be the height of the iFrame, I'll use the [iframe resizer](https://github.com/davidjbradshaw/iframe-resizer).

NAnguiano:
Ok. I switched this to using media queries. I also fixed the issue that was causing it to choose medium instead of small.

kdahlquist:
[...]

PR #409, opened 2017-02-28
Fixes from issues #397 and #395

NAnguiano:
I made the size changes requested in #397, and also fixed some bugs with automatic detection from #395. I also fixed the issue causing it to sometimes choose medium instead of small on smaller computers. 

There were some notes here on bugs and stuff, but they've been fixed as of the fixing of issue #408.

If @anuvarsh , @mihirsamdarshi , @eileenchoe , @yshin4 or @ebachoura can confirm that everything is working right, this one should be pretty easy to test.


yshin4:
npm test runs without any error and all the demo files function without any problem. Merging.

Issue #414, opened 2017-03-01
Why would running both node processes simultaneously cause no graphs to load? (npm start)


NAnguiano:
@dondi 
I was trying to find a way that enables the developers to only run "npm start" in a single place rather than needing to open two terminals (or terminal tabs), cding one into web-client and the other into server, and then running node app.js in both places.

I found two npm options: concurrently and parallelshell, both of which do pretty much the same thing and allow for the user to run two node processes at once. However... Both of them prevent any graphs from being able to load. It seems like the xlsx.parse step of the spreadsheet controller just fails to complete. Running them separately the normal way works just fine, but together they don't. 

Any ideas?

NAnguiano:
I figured out what was happening. It has nothing to do with them being run concurrently - Instead, it's due to the paths on the demo files. Because the paths are hardcoded, when the server is run from the parent directory, all of the paths are wrong. As a result, it's trying to grab a graph that doesn't exist. Fixing that seems to have fixed the problem.

As a result though, the old method of going into both places now doesn't work. For me, it's more convenient to just run `npm start` than it is to do it the old way. I've also integrated nodemon so that I never need to reload the server when a file is changed - nodemon just reloads it automatically. For now, this is something I'll just keep locally. 

Closing this up.

kdahlquist:
@NAnguiano, would you please update the documentation page on the wiki with what you have discovered here.  https://github.com/dondi/GRNsight/wiki/Running-the-Applications

One @NAnguiano has updated the documentation, would everyone on the team please acknowledge on this check list that he/she is able to run it this way:

- [ ] @mihirsamdarshi 
- [ ] @anuvarsh 
- [ ] @ebachoura 
- [ ] @eileenchoe 
- [ ] @yshin4

NAnguiano:
@everyone (why isn't this a thing you can do?), right now this is just on my fork. Everyone can test it on my fork if they want to see how it works. Just load it up, `npm install`, then run `npm start` and prepare to be amazed. 

The implications of doing this extend a little further than just changing the way that the apps are launched. Namely, the beta server (and master, once they are merged) would need to be reconfigured so that the new paths to the demo files are recognized. This wouldn't change much - it would just be a slight change to each of the existing commands, as well as a change in the directory that the commands are run. When the pull request is accepted, I'll make the new commands in the screens and ensure everything works. 

[...]

PR #415, opened 2017-03-06
Enable npm start and incorporate nodemon

NAnguiano:
As stated in issue #414, I've created a method that allows us to just run `npm start` on the parent directory instead of needing to do the current method. As also stated there, you'd need to make sure to run `npm install` before running `npm start`, as you need to get the new dependencies nodemon and parallelshell.

@ebachoura and @yshin4, I'm not sure whether this method works in windows. It should, but you never know. If one of you could check it out and make sure it works, that would be ideal. Just pull my fork, switch to branch beta, and test it out :)

I'll attempt a beta refresh once this is approved and merged.


anuvarsh:
@ebachoura and @yshin4, if either/both of you could confirm that "npm start" works on Windows, I can go ahead and complete this pull request

Also, thank you a million times @NAnguiano! This is very useful and helpful and great!

ebachoura:
I pulled Nicole's Beta and then did "npm install" and when I "npm start" it seems to be running the server at port 5001, but when I localhost:5001, it says that it refuses to connect.

I'm sure that I'm doing something wrong, but I'll confirm with you guys today at the meeting.

NAnguiano:
@yshin4, what is the error you are getting when you try to run npm start?

Issue #392, opened 2017-02-13
Make the arrowheads larger for unweighted and thinnest weighted edges: fix tips going behind nodes


kdahlquist:
Since the GRNmap team is using GRNsight extensively, one thing I've noticed is that the arrowheads for the unweighted and thinnest weighted edges are actually pretty small and hard to see.  We have trouble distinguishing whether an edge is incoming or outgoing from a node.  The arrowheads will need to be larger for these cases; I don't think we will need to enlarge the ones for thicker weighted edges.

yshin4:
@kdahlquist 

I made a change so that the small arrowheads look bigger for: 
1. the thinnest ones for weighted graphs and
2. all the unweighted graph
and the result looks as below. First, two images are how it was before, and last two are after the change I made is applied. Please tell me if there should be any other modifications regarding this issue. 


Before
![image](https://cloud.githubusercontent.com/assets/16246811/24937653/ffb182b0-1ee6-11e7-90ca-a7684639f5c9.png)
![image](https://cloud.githubusercontent.com/assets/16246811/24937657/08361784-1ee7-11e7-80aa-452b5dc4a11b.png)

After
![image](https://cloud.githubusercontent.com/assets/16246811/24937666/14901e12-1ee7-11e7-855f-a2ac7749ec61.png)
![image](https://cloud.githubusercontent.com/assets/16246811/24937669/1e1a4ef8-1ee7-11e7-832d-a93145d622f7.png)

kdahlquist:
This looks good.  The final test will be when I can see this "live" on beta.  I note that in the screenshot above that the self-regulatory arrowheads are no longer centered, though.

There are some test files that were set up to display all the possible edge widths for a weighted graph.  You should use those when making adjustments so that you can see how what you are doing affects all of the possible end markers.

These files are actually a little too complicated to hone in on the specific issue involved, so I edited the one called "check-self-referential-widths.xlsx" to show how to construct a test file that facilitates checking just the widths.

I'm attaching the modified file here, with a screenshot of its output
[check-self-referential-widths.xlsx](https://github.com/dondi/GRNsight/files/917890/check-self-referential-widths.xlsx)

![check-self-referential-widths-output-v1 18 1](https://cloud.githubusercontent.com/assets/6463867/24980204/1ef651a0-1f8c-11e7-9d9e-dd4f8bab9325.jpg)


 plus one I made for labels of all possible lengths of I's and W's.

[...]

Issue #188, opened 2015-04-14
Blunt markers are not centered especially for thick edges


kdahlquist:
Blunt markers do not appear centered depending on orientation; effect is pronounced when the edge is thicker.

anuvarsh:
Updates were made on the blunt markers of especially thick edges.

anuvarsh:
During the last lab meeting we discovered that centering the edges when they are curved uncenters the edges when they are straight. The new plan of action is to create separate coordinates for straight edges and curved edges.

dondi:
I’ve made some adjustments to the blunt-edge positioning code. It visually looks more balanced now, though varying thicknesses still have small differences. But for now, worth a review.

kdahlquist:
Before the beta-server crashed, I was looking at this test file which is in the graph-tests folder:

all-negative-weights.xlsx

Some of the thicker edges in this file are not centered.  If I could pull it up again, I could be more specific, but it's fairly easy to see if you arrange them in a box layout so all the edges are straight.

There are two issues, the blunt marker is not centered on the edge.  I've also noticed that the blunt marker is also not centered on the node box.

yshin4:
Trying to fix the issue by changing the refX value of the marker attribute when it is referring to itself. When value is too low the arrow is hard to see and when the value is too high it gets off the center of the edge. For now adjusted the numbers so it has a mutual distance away from the node and away from the center (in order to minimize the visual offset).

kdahlquist:
Maybe this should be a new issue, but it seems to be a side effect of the changes we have been making to the edges/end markers either on this issue or #318 or #392 or #333.

When I load the test file "all-negative-weights.xlsx", the blunt end marker is not a uniform distance away from the node for all thicknesses.  As the edge gets thicker, the end marker gets progressively closer to the node until it is touching for medium-weights and above.
![blunt-end-markers-too-close](https://cloud.githubusercontent.com/assets/6463867/25251969/27c5d16c-25d0-11e7-934b-c7bc60a47838.jpg)

kdahlquist:
As for the centering, it looks to me like they mostly look OK, except for E-5 to E-6, E-6 to E-7 and E-7 to E-8 in the screenshot in the comment above.

kdahlquist:
After playing with it a bit more, I realize that it's a left vs. right thing and not the thickness.  If the blunt end marker is on the right side of a node, it's too close to the node; if it's on the left, it's OK.  Top and bottom are OK.

[...]

Issue #445, opened 2017-04-19
Protocol for handling pull requests that are not ready to be merged?


NAnguiano:
I think we should have some sort of protocol for pull requests that are not ready to be merged or shouldn't be merged yet - whether it's because they are bugged or because there is additional functionality that's going to be added to them. This would just ensure that incomplete or bugged pull requests don't get added by accident (though hopefully the pull request checklist from #440 would allow bugs to be caught first? :wink:).

Perhaps some sort of label indicating a pull request is not ready for evaluation could be created and added to the pull request. Or, the pull request can be closed and reopened when it is ready for evaluation. Or maybe both - the first being used for small bug fixed and the second being used for more major changes and fixed. 

This could just be an issue that bothers me, but I'm not certain. What do you guys think?

kdahlquist:
I'm not perhaps the best person to comment on this, but I think that a pull request implies that it is ready to be merged, and that if it isn't for some reason, then it shouldn't be there.  @dondi definitely needs to weigh in on this one.

dondi:
It used to be that the act of merging was synonymous with approval—i.e., a pull request remained unmerged until it was deemed ready—but GitHub has since added an explicit "approval" feature (without being merged). This is available when a team member opts to “review” a pull request. When reviewing a request, they can now mark it as approved, or request changes. The coder can then submit additional commits, and the cycle can go on until a pull request is approved. An approved pull request can then be merged whenever.

Of course this feature is optional and need not be used, but its presence does allow for a cycle like I described, where a pull request is first issued, then we can establish that it needs to explicitly get approved before it can be merged.

kdahlquist:
@NAnguiano, does what @dondi suggest fit your needs?  Initially, I was thinking that we could get away with a less formal review process, but if you think it's necessary, then it seems like the process @dondi suggests will work.

NAnguiano:
The initial motivation for creating this issue was two-fold. The first being that I keep seeing we have an open pull request and checking it, only to realize it's something I've already checked. And second being that yesterday, during my work session with @yshin4, I checked and reviewed her pull request (the now closed #436) and was about to merge before she let me know that the PR wasn't actually ready for review because there was more work that needed to be done. The first is a minor annoyance, but I'd like to avoid situations like the second one. 

In this case, it wouldn't necessarily have affected anything should I have just merged it before finding that out. But that may not always be the case. And it's that possible case that worries me.

[...]

Issue #364, opened 2017-01-18
Slider Display Values do not change after loading a file


NAnguiano:
The slider values on beta stop updating once you load a file. Before loading, they work just fine, but after, the display values don't change, though the slider still has an effect on the graph.

NAnguiano:
I've noticed that this actually has broken undo reset. Reset works, but undo reset is no longer functional on beta.

NAnguiano:
This is still an issue. I don't feel comfortable with a master release (#422) unless this is fixed, as this is a pretty major bug with our force graph. I'll assign this to myself for now, but if anyone else wants to take it over and get it done, they can do it if they want.

NAnguiano:
@ebachoura @eileenchoe I'm assigning this to the two of you. I'm also upping this to priority 0. This is too big of a bug in my opinion to ignore in a release.

eileenchoe:
@ebachoura @NAnguiano and @eileenchoe collaborated on this issue and it is now fixed. Issuing the pull request now.

kdahlquist:
I'm sorry to say that this is not completely fixed.  The slider values do change before I load a network.  However, once I load one of the demo's, the values no longer change when I move the sliders.  It looks like the opposite is happening from what @NAnguiano originally reported.

I tested both on Firefox and Chrome with beta v2.0.7.

NAnguiano:
@kdahlquist Beta hadn't been updated with the most recent pull request that contained this fix, which is why it didn't seem to be working. It's updated now.

kdahlquist:
Ah.  I refreshed and can see that it's fixed.  Closing.

PR #446, opened 2017-04-19
graph bug fixes

### LINKED ISSUE #333: Reduce white space one either side of gene label
I've noticed this a few times now, but I think we need to reduce the white space on either side of the label in a node rectangle.  As the labels get wider, the white space does as well and it's using up too much real estate.  This is related to another issue #318.

yshin4:
#333 Reduced white space, minimized the box size, 
#392 changed minimum arrowhead size for thin and unweighted edges,
#188 modified number values to center arrowheads

@NAnguiano, if you have time please review and merge. Thank you!


NAnguiano:
Looks great! I'm not actually sure what the protocol on this would be - since @kdahlquist mentioned wanting a few more changes, I'm not sure this should be merged or not. Since there's not really anything wrong with it, I'll merge it for now, but I think the proper behavior for something like this will be determined in #445.

Pull Request Checklist
- [x] All Tests Pass
The number does not change, but it does have an effect. See issue #364.
Locking, unlocking, and resetting work. Undo reset does not. See issue #364.

kdahlquist:
To answer @NAnguiano, in this case, since tests are passing, no new apparent bugs were introduced, and it looks better than it did before, it's OK (more than OK) to merge and make it live on beta.  Remember, beta doesn't have to be perfect, but it should preserve the basic functionality of the master.

Issue #447, opened 2017-04-19
Hide "Data" menu in beta


kdahlquist:
The "Data" menu got hidden on Master because there is no functionality there and we didn't want to have any placeholder menu items visible.  This is a reminder that we need to hide this menu option on beta, too, so that when it becomes Master, it suddenly doesn't become visible again.

I'm not sure who to assign this to.

NAnguiano:
I went ahead and did this on my fork. It will be in my next PR, which should come tonight.

NAnguiano:
This is now live on beta as per PR #448.

kdahlquist:
Verified complete on beta v2.0.7.

PR #448, opened 2017-04-19
Random Tweaks

NAnguiano:
#447 Hide data menu option
#395 reduce screen size requirement for large screen to be selected
Also remove phantom beta folder


eileenchoe:
Verified that data menu is now hidden (#447) and that large viewport is automatically selected on a 2550 x 1440 monitor (#395)

Pull Request Checklist
- [x] All Tests Pass

PR #449, opened 2017-04-20
slider values

eileenchoe:
#364 fixed. The issue was that slider event handler was only set up to update the force, not the displayed value. As a result, this broke the reset. Also updated the style sheet to fix small bug on the media width.


ebachoura:
Pull Request Checklist
- [ ] All Tests Pass

ebachoura:
Hey mom, I did it!

PR #452, opened 2017-04-20
SIF importer fix

### LINKED ISSUE #451: SIF importer crashing due to "entry" being null
Looks like something brought the beta server down while @kdahlquist was testing #362. Here's the error message.
screen shot 2017-04-19 at 10 30 59 pm
It got a little cut off at the bottom, but there's nothing below that. Last line just says `at Object.oncomplete (fs.js:107:15)`. 

@eileenchoe, I'm sending this your way since SIF is your domain. 

The responsible file is posted in #362.

eileenchoe:
#451 Fixed source of error that caused server to crash.

### DIFF
--- a/server/controllers/importers/sif.js
+++ b/server/controllers/importers/sif.js
@@ -77,7 +77,9 @@ module.exports = function (sif) {
   var links = [];
   var networkType = "unweighted";
 
-  if (entries.includes(null)) {
+  var nullEntries = entries.filter(function (entry) { return entry === null; });
+
+  if (nullEntries.length > 0) {
     errors.push(constants.errors.SIF_STRAY_DATA_ERROR);
   } else {
     entries.forEach(function (entry) {
--- a/server/controllers/importers/sif.js
+++ b/server/controllers/importers/sif.js
@@ -74,45 +74,52 @@ module.exports = function (sif) {
   });
 
   var genes = [];
-  entries.forEach(function (entry) {
-    if (entry.length && entry[GENE_NAME] && genes.indexOf(entry[GENE_NAME]) == constants.NOT_FOUND) {
-      genes.push(entry[GENE_NAME]);
-    }
-  });
-
-  var networkType = sifNetworkType(entries);
-  if (networkType.warnings) {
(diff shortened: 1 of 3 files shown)


server/controllers/importers/sif.js
```diff
@@ -74,45 +74,52 @@ module.exports = function (sif) {
   });
 
   var genes = [];
-  entries.forEach(function (entry) {
-    if (entry.length && entry[GENE_NAME] && genes.indexOf(entry[GENE_NAME]) == constants.NOT_FOUND) {
-      genes.push(entry[GENE_NAME]);
+  var links = [];
+  var networkType = "unweighted";
+
+  if (entries.includes(null)) {
```
dondi: Watch out—`Array.includes` is super new (defined in 2016): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

You can do something like `entries.filter(function (entry) { return entry === null; })` then check if the resulting array is not empty. Not as concise but more backwards-compatible.

server/controllers/spreadsheet-controller.js
```diff
@@ -391,6 +392,15 @@ var errorList = {
     "pasting just your adjacency matrix into a fresh Excel Workbook and saving it."
   },
 
+  unknownFileError: {
+    errorCode: "unknownFileError",
+    possibleCause: "An unexpected error occured originating in the file.",
+    suggestedFix: "Please check the format of your spreadsheet with the guidlines outlined on the" +
```
dondi: Typo: “guidlines”

server/controllers/importers/sif.js
dondi: `indexOf` may also work, but because `null` is special I would check on whether there are gotchas with it when used with `indexOf`.

server/controllers/spreadsheet-controller.js
eileenchoe: Whoops this actually was initial work for another issue I was working on before I took up fixing the SIF.

eileenchoe:
Made changes as suggested by @dondi 's review!

Issue #443, opened 2017-04-19
Weights on weighted graphs congregate in the upper left hand corner when "always show weights" is selected


NAnguiano:
While writing issue #442, I noticed this weirdness going on.

![screen shot 2017-04-18 at 7 08 26 pm](https://cloud.githubusercontent.com/assets/6444156/25160355/736e0920-246a-11e7-81d3-8708f3574af9.png)
![screen shot 2017-04-18 at 7 09 51 pm](https://cloud.githubusercontent.com/assets/6444156/25160389/a2e6caac-246a-11e7-967b-3ddaf1c2134c.png)
![screen shot 2017-04-18 at 7 09 00 pm](https://cloud.githubusercontent.com/assets/6444156/25160369/815901b6-246a-11e7-8fb6-eb9802d309fa.png)

I'm pretty sure this isn't supposed to be happening.  It seems like the weights are just having a little party up in the upper left hand edge of the bounding box. It happens on both fixed and adaptive bounding boxes, on all sizes. It also happens on weighted graphs drawn the typical way - this particular graph was the source behind issue #442.

From what I know about d3, when something is drawn but not used, it gets placed in the upper left hand corner. But since the thing that is drawn isn't used, the user won't notice it. Well, for whatever reason, the text element containing the weight is drawn twice. The first text element goes on the link, and the second goes unused - so it sits in the top left hand corner as something drawn but unused. This is fine when mouse over is selected, because the unused texts are just sitting invisible in the top left. But when "always show" is selected, everything containing the class "weight" is also given the class "visible". So those unused text elements in the top left hand corner are made visible.

Assigning this to myself and adding it to my tasks.

eileenchoe:
Problem was that weights for display were being searched over the .svg class selector instead of specifically the .link selector. Submitted PR #457

NAnguiano:
@eileenchoe fixed this. This is up on the live beta now for testing :)

kdahlquist:
Looks fixed to me.  Closing.

PR #457, opened 2017-04-27
fixed bug #443

eileenchoe:
#443 : Bug fix, weights now do not congregate at the upper left corner when "always show weights" is selected.


NAnguiano:
@eileenchoe ily ❤️ 

Pull Request Checklist
- [x] All Tests Pass
- [x] Force Sliders change the number above them and have an effect on the graph
- [x] Locking/Unlocking/Resetting/Undoing the force graph parameters works as expected

Issue #355, opened 2016-11-22
Network loads with black edges and cyan blunt end markers


kdahlquist:
Here is the spreadsheet from @Nwilli31.  When it loads in GRNsight it has black edges and cyan blunt end markers on some of the edges.

The sheet name is "network" but has real number values for the adjacency matrix (instead of just 1, 0).  I guess we don't have a case testing for this.

If we really wanted to help the user, we would detect that there is a sheet called "network" and that it had numbers other than 1 or 0 in the matrix and then issue a warning that it will just use black edges and pointed arrowheads and if the user wants colored edges s/he would have to change the name of the sheet to network_optimized_weights.
[dCIN5_network_14-nodes_25-edges.xlsx](https://github.com/dondi/GRNsight/files/605195/dCIN5_network_14-nodes_25-edges.xlsx)

kdahlquist:
Maybe this one should be tackled by our syntactic error team of @mihirsamdarshi , @ebachoura , and @eileenchoe .

kdahlquist:
Assigning this to @mihirsamdarshi.  As noted above, a syntactic test needs to be written for the Excel importer to take care of this case.  Work with @ebachoura and @eileenchoe to define what the equivalent syntactic test would be for GraphML and SIF, respectively

kdahlquist:
This is priority 0.5, after @mihirsamdarshi finishes with #386.

kdahlquist:
@anuvarsh says that these tests exist, but since further work has been done since then, something must have changed.

anuvarsh:
These files technically exist in test-files>sheet-modifications>wrong-sheet-name-input/output, but they are a several random and unnecessary sheets in the file. Perhaps these files should be reworked/simplified so only network or network-weighted sheets exist so we can confirm that the changes in those sheets are the only things being tested.

dondi:
@mihirsamdarshi has just started and has been able to replicate the issue. He will experiment with some changes or additions to the test files and @dondi suggested that he looked at references to the `stroke` property in _graph.js_.

mihirsamdarshi:
it seems that the issue originates in the actually _graph.js_ file, in the lines that depend on the normalized weights for the ends of the edges. I'm working on it still.

kdahlquist:
What is the update on progress on this issue?

mihirsamdarshi:
[...]

PR #458, opened 2017-04-28
Fixed issue #355

mihirsamdarshi:
Added/changed code to properly do every use case specified in issue #355, and also added tests for this.


NAnguiano:
Pull Request Checklist
- [ ] **All Tests Pass**
I'm getting 5 test failures. 1 of them is a timeout (which I'm relatively sure would have passed), but the remaining 4 are outright failures.
screen shot 2017-04-27 at 7 59 43 pm

Not sure what's going on here.

- [ ] **Hide/Show Weights works as expected** 
The bug from issue #443 seems to be present, though I suspect this is from the fact that the current beta hasn't been merged into this branch.

- [x] Force Sliders change the number above them and have an effect on the graph
- [x] Locking/Unlocking/Resetting/Undoing the force graph parameters works as expected

NAnguiano:
I loaded the file that initially triggered issue #355 and I found this bug:
screen shot 2017-04-27 at 8 07 23 pm

It's listing 25 warnings, but nothing is showing up. I suspect this is due to them not being added the upload.js, but I'm not sure. Outside of this bug, the file looks totally fine.

The second file in the issue - the one with a network_optimized_weights sheets with just 0s and 1s - doesn't seem to be triggering any warnings at all.

kdahlquist:
So, it looks like additional bugs need to be fixed before this PR can be accepted.

So, @NAnguiano, please use the "request changes" option on the "files changed" tab or maybe we need to close this PR and @mihirsamdarshi will submit another one when the bugs are fixed.

Just want to get us using the workflow...

test/import-tests.js
```diff
@@ -377,3 +377,10 @@ describe("Import from GraphML", function () {
     });
   });
 });
+
+describe('sheet-named-incorrectly', function () {
+  it('1 wrong name warning per file', function () {
```
NAnguiano: To follow the structure of the other warnings, this should likely be written as "should return one wrong name warning". Also, I think this test would fit a little better in errors-sheet-modification rather than import tests. Import tests is for SIF/ML.

web-client/public/js/upload.js
```diff
@@ -308,6 +309,8 @@ $(function () {
     createWarningsString(edgesWithoutWeightsCount,7);
     createWarningsString(edgeDefaultNotDirectedCount,8);
     createWarningsString(sifFormatWarningCount, 9);
+    createWarningsString(incorrectlyNamedSheetWarningCount, 10);
+
```
NAnguiano: Extraneous space can be removed.

server/controllers/spreadsheet-controller.js
```diff
@@ -401,79 +422,158 @@ var errorList = {
...
       warningCode: "INVALID_NETWORK_SIZE",
       errorDescription: "Your network has " + genesLength + " genes, and " + edgesLength + " edges. Please note that networks are recommended to have less than 50 genes and 100 edges."
     }
-  }
+  },
+
+  incorrectlyNamedSheetWarning: function() {
+    return {
+      warningCode: "INCORRECTLY_NAMED_SHEET",
+      errorDescription: "The uploaded spreadsheet was named incorrectly. Please check if the sheet(s) in the uploaded spreadsheet have been named properly."
```
NAnguiano: I think this could be written a bit more clearly. I think something like "The uploaded file appears to contain a weighted network, but contains no 'network_optimized_weights' sheet. A weighted network must be contained in a sheet called 'network_optimized_weights' in order to be drawn as a weighted graph. Please check if the sheet(s) in the uploaded spreadsheet have been named properly."

mihirsamdarshi:
All tests now pass, requesting review on pull request.

server/controllers/spreadsheet-controller.js
```diff
@@ -464,6 +465,13 @@ var warningsList = {
       warningCode: "INVALID_NETWORK_SIZE",
       errorDescription: "Your network has " + genesLength + " genes, and " + edgesLength + " edges. Please note that networks are recommended to have less than 50 genes and 100 edges."
     }
+  },
+
+  incorrectlyNamedSheetWarning: function() {
+    return {
+      warningCode: "INCORRECTLY_NAMED_SHEET",
+      errorDescription: "The uploaded file appears to contain a weighted network, but contains no 'network_optimized_weights' sheet. A weighted network must be contained in a sheet called 'network_optimized_weights' in order to be drawn as a weighted graph. Please check if the sheet(s) in the uploaded spreadsheet have been named properly."
```
dondi: Strictly speaking, we do not assume that code editors or viewers perform automatic line breaks in the same way, so the preference for long strings like this is to break them up as a concatenation, thus permitting line breaks. A natural indentation also emerges from this:

    errorDescription: "The uploaded file appears to contain a weighted network, but contains no " +
        "'network_optimized_weights' sheet. A weighted network must be contained in a sheet called " +
        "'network_optimized_weights' in order to be drawn as a weighted graph. Please check if the " +
        "sheet(s) in the uploaded spreadsheet have been named properly."

test/test.js
```diff
@@ -186,15 +187,15 @@ function checkForGene(test, frequency, input) {
 function warningsCountError(input, frequency) {
   var sheet = xlsx.parse(input),
       network = spreadsheetController.parseSheet(sheet),
-      warningsCountErrorArray = network.errors.filter(function(x){return x.errorCode == "WARNINGS_OVERLOAD";});
+      warningsCountErrorArray = network.errors.filter(function(x){return x.errorCode === "WARNINGS_OVERLOAD";});
```
dondi: One best practice in collaborative coding: when the group has adopted a particular style (and it has), as you evolve the existing code and see opportunities to improve style adherence, go do it. This is one such opportunity: we have decided upon individualized variable declarations, so this block of code should become:

    var sheet = xlsx.parse(input);
    var network = spreadsheetController.parseSheet(sheet);
    var warningsCountErrorArray = network.errors.filter(function (x) { return x.errorCode === "WARNINGS_OVERLOAD"; });

test/test.js
```diff
@@ -186,15 +187,15 @@ function checkForGene(test, frequency, input) {
...
 
   assert.equal(frequency, warningsCountErrorArray.length);
 }
 
 function errorsCountError(input, frequency) {
   var sheet = xlsx.parse(input),
       network = spreadsheetController.parseSheet(sheet),
-      errorsCountErrorArray = network.errors.filter(function(x){return x.errorCode == "ERRORS_OVERLOAD"});
+      errorsCountErrorArray = network.errors.filter(function(x){return x.errorCode === "ERRORS_OVERLOAD"});
```
dondi: Ditto one-`var`-per-variable.

test/test.js
```diff
@@ -225,59 +226,67 @@ function noWarnings(input) {
 function missingSourceWarning(input, frequency) {
   var sheet = xlsx.parse(input),
       network = spreadsheetController.parseSheet(sheet);
-  var missingSourceCount = network.warnings.filter(function(x){return x.warningCode=="MISSING_SOURCE"});
+  var missingSourceCount = network.warnings.filter(function(x){return x.warningCode === "MISSING_SOURCE"});
```
dondi: Note the dangers now of not keeping your peripheral vision open for style adherence opportunities—this block, in some sense, is even worse than before because it uses inconsistent `var` styles within a line of each other.

test/test.js
```diff
@@ -225,59 +226,67 @@ function noWarnings(input) {
...
 }
 
+function incorrectlyNamedSheetWarning(input, frequency){
+  var sheet = xlsx.parse(input),
+    network = spreadsheetController.parseSheet(sheet);
+  var incorrectlyNamedSheetWarning = network.warnings.filter(function(x){return x.warningCode === "INCORRECTLY_NAMED_SHEET"});
+
+  assert.equal(frequency, incorrectlyNamedSheetWarning.length);
+}
+
```
dondi: There might be other places where we can separate variable declarations into separate `var` statements, but we are not obligated to revise those for now. We want to revise _these_ because your work happens to encounter them. These little improvements, if done consistently, will allow the code to evolve toward a cleaner, more consistent, and less error-prone style.

server/controllers/spreadsheet-controller.js
```diff
@@ -135,7 +135,7 @@ var parseSheet = function(sheet) {
               if (currentSheet.data[row][column] === undefined) {
                 // SHOULD BE: addError(network, errorList.missingValueError(row, column));
                 addWarning(network, warningsList.invalidMatrixDataWarning(row, column));
-              } else if (isNaN(+("" + currentSheet.data[row][column]))) {
+            } else if (isNaN(+("" + currentSheet.data[row][column])) || typeof currentSheet.data[row][column] !== "number") {
```
dondi: Indentation is incorrect here.

kdahlquist:
I see that there are merge conflicts with "spreadsheet-controller.js"; I am making the executive decision to put a hold on this pull request until the research period in the summer so that we don't introduce more bugs in the next 24 hours.  @mihirsamdarshi, this will be the first thing we take up on May 15.

mihirsamdarshi:
The pull request was merged, but Eileen doesn't have the SSH token on the computer she has right now so we will refresh beta as soon as she has access to it.

Issue #474, opened 2017-05-03
Run the Graph Statistics Report After completing semantic analysis


NAnguiano:
I just managed to freeze beta for a solid few minutes because I uploaded a large graph multiple times. 

Right now, we run the graph statistics report after we finish parsing a network, but before the semantic analysis. This means that we haven't checked to see whether there are too many nodes or edges.

A fully connected graph with 150 nodes takes about 1.5-2 minutes to finish loading when graph statistics is enabled, and about a second with them off - nevermind the fact that one can actually just upload it over and over and over without limit (because they may think that it didn't load correctly - at least that's why I did it...), potentially bringing GRNsight down until they all finish generating.

So I propose that we run the graph statistics report only after the semantic analysis is complete. There's no real reason to run it beforehand that I can think of anyways. 

This would also have the benefit of removing the cytoscape errors spam from the tests when we run them.

Thoughts?

kdahlquist:
Yes, absolutely!

Question: why are you trying to run things with cytoscape turned on?  It's not actually computing the graph statistics correctly as per #296, #298, and #307.  We can't turn it back on until we've addressed those issues, anything we display from it right now will be giving wrong information to the user.

NAnguiano:
I'm actually not fully sure why we have it on right now. It was turned on for issue #343.  I think there are some tests in our test suite that rely on it being on not to fail, but they could be turned off.

kdahlquist:
OK, I see that we probably forgot to turn them off again.

For now we just need to turn cytoscape off.  The GRNmap data analysis team has been working all semester to try and interpret graph stats from Gephi and we're not actually done with that yet.  Until we have an independent way of verifying that what cytoscape is returning is valid, we can't turn it on.

I'm moving this to a 1; it's not in the critical path for release.

NAnguiano:
Sounds good to me! I'll turn it off.

NAnguiano:
It's off. Just so you all know, now that it's off, there are 29 tests that will be returned as "pending". That's because I've set it to just skip them. All of those tests are the graph stats tests.

kdahlquist:
Backlogging because we are putting a freeze on new features for now.  Can re-open in future when we are ready to tackle graph statistics again.

Issue #470, opened 2017-05-02
Allowable zooming/scrolling actions when "Restrict Graph to Viewport" is selected


kdahlquist:
Carried over from #396, these are the allowable zooming/scrolling actions for when "Restrict Graph to Viewport" is selected:

The way things work now under the adaptive setting is that scrolling or grabbing the graph moves the entire bounding box to a different relative location within the viewport.

Thus, when "restrict graph to viewport" is selected, to be true to this label, we can't allow scrolling or grabbing of the graph because that would bring the bounding box out of alignment with the viewport.
* At 100% zoom, no scrolling/grabbing is allowed
* Remember, we are not allowed to make graph bigger anyway (zoom in)
* If graph is smaller (zoomed out), then you can scroll/grab, but you can't move the bounding box beyond the edge of the viewport (top, bottom, left, right).
    * if subsequently you do zoom in, the bounding box is going to get pushed so that it never goes outside the viewport, so that it will end up at 100% flush with all four sides of the viewport.

dondi:
Agreed—“restrict graph to viewport” is essentially the current behavior, with the third bullet being the way this mode handles zooming. @NAnguiano and I actually talked about how to implement this today, and I gave her a sketch of how to potentially do it.

kdahlquist:
This is complete as of v2.0.16, but there one remaining small issue that will become its own task for later.

Issue #469, opened 2017-05-02
Use hand icon for mouse pointer when grabbing (scrolling) is allowed


kdahlquist:
As discussed in issue #396, change the mouse pointer to a hand icon when grabbing (scrolling) is allowed.

NAnguiano:
This one is fully done. I tested a bunch of different zoom/scroll/pan combinations and the hand seemed to always behave as intended for the most part. There were a few times it got stuck in the "grabbed" position, but that seemed to only occur when I'd grab the graph and shake it back and forth as quickly as I could. Clicking again fixes it.

kdahlquist:
Sounds good--looking forward to seeing it.

dondi:
Rapidly shaking back and forth is a good test for the robustness of your mouse handling. If things get stuck or awry when this is happening, that is usually an indication that the mouse cursor has left the mouse-listening area without giving that area a chance to clean up.

If the mouse listener is the individual gene node, the mouse may move faster than the gene node can move, and thus one might be left in an inconsistent state. The fix for this would be to move some mouse-listening duties to the overall container.

But I would deal with that later. Keep this one focused on just the hand icon and someone else can deal with rapid shaking some other time.

kdahlquist:
This is confirmed in beta v2.0.16.

Issue #467, opened 2017-05-02
Mouse pointer should be arrowhead, not vertical cursor bar when hovering over scrolling arrows


kdahlquist:
The mouse pointer should be an arrowhead, not vertical cursor bar when hovering over scrolling arrows.

Only, if there is time.

kdahlquist:
This is confirmed fixed in beta v2.0.16.

PR #476, opened 2017-05-03
PR for a Bunch o' Issues

NAnguiano:
6:15 am is the perfect time for a pull request! 

To avoid comment spam, I'm going to list all of the completed issues and any comments here:
#461 - The iFrame fits to the window now. This one is actually live already! I set a minimum height requirement, so unfortunately for small screens, it's not going to be a perfect fit at the moment. I didn't actually end up using iFrame resizer after all because I found that it didn't actually seem to be doing what I wanted it to do and I took a bit of a performance hit while using it. I found a nice 4-line implementation that seems to do the same thing. There's a few tweaks that could be made, but it's way better than it was.
#466 - Some initial tweaks to the arrows. Nothing major.
#467 - Mouse cursor is now a pointer above the arrows and slider.
#469 - The pointer is now a hand on graphs that can be scrolled. I see @dondi 's comment on the mouse handling being a little flimsy. The entire viewport serves as the mouse handler, so I'm not sure it's being caused by the cursor leaving the mouse handling area, but eh... 
#470 - The bounding box now stays within the viewport, no matter where it's dragged. There is one behavior I'm not sure is wanted or not, but I spent a good bit trying to find a good solution before decided to move on. Excuse my incredibly sleep-deprived brain as I try to explain this. The edge of the bounding box sometimes gets a bit "sticky" when you zoom close and then far away. That is, when you pull the bounding box over and push it against the edge of the viewport, zoom in, and then go to zoom out, the bounding box kinda.. sticks to it. It eventually detaches at the point at which you initially started zooming in. I hope that makes sense. It's not really bad or anything, but I tried a bit to fix it and encountered some frustrating D3 behavior that I couldn't immediately think of a clean way to handle. So it's just there for now, but I can think of a solution if one is wanted.
#474 - I turned off the stats report, as noted. BTW guys, the tests now take like 20 seconds to run.. it's fantastic.
#396 - I think @kdahlquist made a comment here about force re-application once a graph has completely gone to rest... And stating that a graph at rest didn't start expanding when switched from fixed to adaptive. I found the root cause and fixed it during my efforts in all of these issues, so just giving that a mention.

#465 - So this is the one that I haven't finished yet. All the pieces are there, I just need to connect them. I gave the zoom slider more space, but everything else is coming soon. Though I can say that the log 2 transform idea did help! It didn't work exactly as I hoped, but it did give me an idea that i've 50% or so implemented at this point and seems to work well.

So there ya go. Now I'm going to go take a nap.

### DIFF
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -128,8 +128,8 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
   var previousMax = [width, height];
   var previousScale = zoom.scale();
   function zoomed(manual = false) {
-    let scaleRange = maximumScale - minimumScale;
-    let percentZoom = zoom.scale() / scaleRange * 100;
+    var scaleRange = maximumScale - minimumScale;
+    var percentZoom = zoom.scale() / scaleRange * 100;
     if (!MANUAL_ZOOM) {
       $(".zoomSlider").val(d3.event.scale.toFixed(2)); // This doesn't work using d3 selection for some reason
     }
@@ -167,7 +167,7 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
   }
 
   d3.selectAll(".scrollBtn").on("click", null); // Remove event handlers, if there were any.
-  let arrowMovememnt = [ "Up", "Left", "Right", "Down" ];
+  var arrowMovememnt = [ "Up", "Left", "Right", "Down" ];
   arrowMovememnt.forEach((direction) => {
     d3.select(".scroll" + direction).on("click", function () {
       move(direction.toLowerCase());
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -15,6 +15,8 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
       gridWidth = 300,
       colorOptimal = true;
 
+  var CURSOR_CLASSES = "cursorGrab cursorGrabbing";
+
   $('#warningMessage').html(warnings.length != 0 ? "Click here in order to view warnings." : "");
 
   var getNodeWidth = function (node) {
@@ -29,12 +31,12 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
   var adaptive = !$("input[name='viewport']").prop("checked");
   var scrolling = adaptive;
 
-  var MIN_SCALE = 0.25
+  var MIN_SCALE = 0.25;
   var FIXED_MAX_SCALE = 1;
   var ADAPTIVE_MAX_SCALE = 4;
 
   var minimumScale = MIN_SCALE;
-  var maximumScale = (adaptive) ? ADAPTIVE_MAX_SCALE : FIXED_MAX_SCALE;
+  var maximumScale = adaptive ? ADAPTIVE_MAX_SCALE : FIXED_MAX_SCALE;
   d3.select(".zoomSlider").attr("max", maximumScale);
 
   var BORDER_OFFSET = 4;
@@ -94,7 +96,7 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
       .origin(function(d) { return d; })
       .on("dragstart", dragstart);
 
-  var MANUAL_ZOOM = false;
+  var manualZoom = false;
   var zoom = d3.behavior.zoom()
     .center([width / 2, height / 2])
     .scaleExtent([minimumScale, maximumScale])
@@ -127,10 +129,10 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
   var previousValue = zoom.translate().slice();
   var previousMax = [width, height];
   var previousScale = zoom.scale();
-  function zoomed(manual = false) {
+  function zoomed() {
     var scaleRange = maximumScale - minimumScale;
     var percentZoom = zoom.scale() / scaleRange * 100;
-    if (!MANUAL_ZOOM) {
+    if (!manualZoom) {
       $(".zoomSlider").val(d3.event.scale.toFixed(2)); // This doesn't work using d3 selection for some reason
     }
     if (!adaptive) { // Limit to viewport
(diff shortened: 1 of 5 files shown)


web-client/public/js/graph.js
```diff
@@ -112,46 +124,71 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
                      .append("g")
 
 
+  var previousValue = zoom.translate().slice();
+  var previousMax = [width, height];
+  var previousScale = zoom.scale();
   function zoomed(manual = false) {
```
dondi: I know this was not changed in this PR, but note that default parameters (i.e., `manual = false`) are an ES6-only feature. This has probably gone unnoticed because this is browser-side code and most browsers are now ES6-capable, but still worth noting because everywhere else, we are still at ES5.

web-client/public/js/graph.js
```diff
@@ -112,46 +124,71 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
+      }
 
-  d3.select(".scrollLeft").on("click", function () {
-    move("left");
-  });
+      if (d3.event.translate[1] < 0) {
+        d3.event.translate[1] = 0;
+      } else if (d3.event.translate[1] > maxY) {
+        d3.event.translate[1] = maxY;
+      }
```
dondi: Won't these `if` blocks be more compact with `Math.min` and `Math.max`? These are actually something like:

    d3.event.translate[0] = Math.min(Math.max(d3.event.translate[0], 0), maxX);
    d3.event.translate[1] = Math.min(Math.max(d3.event.translate[1], 0), maxY);

web-client/public/js/graph.js
```diff
@@ -112,46 +124,71 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
-    move("right");
-  });
+      zoom.translate([d3.event.translate[0], d3.event.translate[1]]);
+    }
+    if (!scrolling && d3.event.scale < 1) {
+      $container.addClass("cursorGrab");
+      scrolling = true;
+    } else if (!adaptive && scrolling && d3.event.scale >= 1) {
+      if ($container.hasClass("cursorGrab")) $container.removeClass("cursorGrab");
+      if ($container.hasClass("cursorGrabbing")) $container.removeClass("cursorGrabbing");
```
dondi: Stick to our style—brace these out.

    if ($container.hasClass("cursorGrab")) {
        $container.removeClass("cursorGrab");
    }

web-client/public/js/graph.js
```diff
@@ -112,46 +124,71 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
+      scrolling = false;
+    }
+    // $("#zoomPercent").html(percentZoom.toFixed(1) + "%");
```
dondi: This commented-out line can be taken out completely, yes?

web-client/public/js/graph.js
```diff
@@ -112,46 +124,71 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
+    svg.attr("transform", "translate(" + zoom.translate() + ")scale(" + d3.event.scale + ")");
+  }
 
-  d3.select(".scrollDown").on("click", function () {
-    move("down");
+  d3.selectAll(".scrollBtn").on("click", null); // Remove event handlers, if there were any.
+  var arrowMovememnt = [ "Up", "Left", "Right", "Down" ];
+  arrowMovememnt.forEach((direction) => {
```
dondi: Arrow functions are ES6-only. I don’t strictly mind this especially because this is browser code, but I don't recall our declaring ES6 as a requirement yet. If we do move to ES6, we need to specify the appropriate minimum browser versions.

dondi: (and if we do move to ES6, we should do so whole-hog—thus I am more inclined now to hold back on the ES6 and make it a full future task in itself)

dondi: I mean, if we were doing this, we should be using `const` and `let` and all of the others, too.

web-client/public/js/graph.js
```diff
@@ -112,46 +124,71 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
 
   d3.select(".zoomSlider").on("input", function () {
     var newScale = this.value;
-    scale(newScale);
+    zoom.scale(newScale);
+    svg.transition().call(zoom.event);
   }).on("mousedown", function () {
     MANUAL_ZOOM = true;
   }).on("mouseup", function () {
     MANUAL_ZOOM = false;
```
dondi: Why is `MANUAL_ZOOM` capitalized when its value might change? That is in consistent with our style.

web-client/public/js/graph.js
```diff
@@ -170,44 +207,65 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
-      var newHeight = $(".grnsight-container").css("height");
+    } else if (fixed) {
+      adaptive = false;
+      maximumScale = 1;
+      zoom.scaleExtent([minimumScale, maximumScale]);
+      if (zoom.scale() >= 1) {
+        zoom.scale(1);
+        scrolling = false;
+        if ($container.hasClass("cursorGrabbing")) $container.removeClass("cursorGrabbing");
+        if ($container.hasClass("cursorGrab")) $container.removeClass("cursorGrab");
```
dondi: More `if`s that need to be broken up with braces added.

web-client/public/js/graph.js
```diff
@@ -170,44 +207,65 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
 
+  $container.on("mousedown", function () {
+     if (scrolling) {
+       if ($container.hasClass("cursorGrab")) $container.removeClass("cursorGrab");
+       if (!$container.hasClass("cursorGrabbing")) $container.addClass("cursorGrabbing");
+     }
+  }).on("mouseup", function () {
+    if (scrolling) {
+      if ($container.hasClass("cursorGrabbing")) $container.removeClass("cursorGrabbing");
+      if (!$container.hasClass("cursorGrab")) $container.addClass("cursorGrab");
```
dondi: More line break needs; but actually, having now seen this multiple times, it sounds like you need two helper functions:

    removeClassIfPresent($element, className)
    addClassIfAbsent($element, className)

.....in fact, I think that jQuery is actually pretty robust here. You might not need the condition check if removing or adding classes. Removing a non-existent class should have no effect, and neither should adding an already-existent class.

I suggest that you do a quick experiment to suss out the behavior of `removeClass` and `addClass` under these conditions, then adjust the code accordingly.

web-client/views/upload.jade
```diff
@@ -186,32 +186,33 @@ html
...
+              td(class='scrollBtn scrollDown')
+                p(class='movementArrow') &#8595;
+              td 
+          table(class='zoomTable')
+            tr 
+              td
+                input(class="zoomSlider" type="range" min="0" max="0" value="1" step="0.1")
+            //- tr
+            //-   td
+            //-     span(style="color:#AAA" id="zoomPercent") 100%
```
dondi: More code to just toast, presumably?

web-client/public/js/graph.js
NAnguiano: Woah, really? I didn't know that that was ES6 exclusive syntax! I'll change that to be ES5 then and do a manual workaround for it for now.

web-client/public/js/graph.js
NAnguiano: The commented out lines all mentioned here are for the implementations of #465 that haven't been put in yet. I'll remove them in the PR and add them back later.

web-client/public/js/graph.js
NAnguiano: Oops, I've been doing so many for-eaches in other classes using the es6 syntax that I forgot that wasn't the way you were supposed to do it.

web-client/public/js/graph.js
NAnguiano: Yep, those both seem to work!

NAnguiano:
@dondi changes are done and pushed.

web-client/public/js/graph.js
```diff
@@ -140,35 +142,23 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
       zoom.translate([d3.event.translate[0], d3.event.translate[1]]);
     }
     if (!scrolling && d3.event.scale < 1) {
-      $container.addClass("cursorGrab");
+      $container.removeClass(CURSOR_CLASSES).addClass("cursorGrab");
       scrolling = true;
     } else if (!adaptive && scrolling && d3.event.scale >= 1) {
-      if ($container.hasClass("cursorGrab")) $container.removeClass("cursorGrab");
-      if ($container.hasClass("cursorGrabbing")) $container.removeClass("cursorGrabbing");
+      $container.removeClass(CURSOR_CLASSES);
```
dondi: Ah just look at that code reduction 😎

web-client/public/js/graph.js
```diff
@@ -251,13 +240,11 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
      if (scrolling) {
-       if ($container.hasClass("cursorGrab")) $container.removeClass("cursorGrab");
-       if (!$container.hasClass("cursorGrabbing")) $container.addClass("cursorGrabbing");
+       $container.removeClass(CURSOR_CLASSES).addClass("cursorGrabbing");
      }
   }).on("mouseup", function () {
     if (scrolling) {
-      if ($container.hasClass("cursorGrabbing")) $container.removeClass("cursorGrabbing");
-      if (!$container.hasClass("cursorGrab")) $container.addClass("cursorGrab");
+      $container.removeClass(CURSOR_CLASSES).addClass("cursorGrab");
```
dondi: _And_ you remembered to chain—this is as concise as it gets, while remaining readable.

Issue #363, opened 2017-01-18
Add "syntactic" tests for graphml format


kdahlquist:
Need to add more tests for graphml format input files. @ebachoura and @eileenchoe can either split SIF and graphml or both work on both. Discuss with @dondi this week.

kdahlquist:
Assigning this solely to @ebachoura (@mihirsamdarshi has been given other tasks).

Here's how to approach this one:
1. Read the documentation on the Documentation page of the website to fully understand the specification for this format in GRNsight (which only uses certain aspects of GraphML).
2. Look in the test suite to see what tests are already there from @dondi's initial implementation
3. Enumerate the cases for which we don't have test coverage, looking at the cases for the Excel format may be helpful here.  Perhaps we could have a meeting with just @kdahlquist, @eileenchoe, and @ebachoura to go over likely edge cases to look for.  The errors and warnings glossaries on the wiki will also be helpful; it would be a good place to add any new cases.
4. Then write the tests.

kdahlquist:
Warnings that have already been written are now showing up in the warnings box on GRNsight when use uploads bad file.

ebachoura:
@kdahlquist I am free to meet during these times this week. I am also hoping to meet before Wednesday.
- Tuesday 2:15 - 4:00
- Wednesday 11:20 - 2:25

kdahlquist:
@ebachoura I can meet on Tuesday at 2:15; please come to my office, LSB 289.

kdahlquist:
@ebachoura and @kdahlquist just met and discussed strategy for how to approach these tests.
1. Use simple test files that are "bare bones" and hone in on the thing that needs to be tested.  A simple graph of four nodes called A, B, C, D with four edges covering the different edge types was suggested.
2. Think "outside the box", i.e., don't approach the program in the way that a user is expected to do by following the directions.  Instead, think of all the dumb things that could be done to break the program.  Think about extreme values and edge cases.
3. The first test is whether it is a .graphml file and not some other filetype.
4. Then go line by line through the sample graphml file shown on the documentation page of the website and ask the questions:
    * Does this line exist?
    * is it formatted properly?
    * does it have the required information
5. We are ignoring a lot of what graphml can do.  Question is whether to warn the user that this is happening.  @dondi should comment at Wednesday's meeting.
6. Keep in mind that some of the things that are in the GRNsight-exported graphml are there for compatibility with Cytoscape and yED and are not actually needed by GRNsight.
[...]

Issue #428, opened 2017-04-03
Need to return error when user uploads blank file


kdahlquist:
At the meeting Thursday, we discovered that GRNsight does not throw an error when a user uploads a blank file, i.e., a .xlsx, .sif, or .graphml file with no content.  Instead, GRNsight loads a blank graph, which correctly says 0 nodes, 0 edges at the top right.  However, we think that the user should be alerted that the file is blank instead, otherwise s/he might get frustrated trying to load the same file multiple times with only a very subtle cue that something is wrong.

Discussion ensued as to whether this was a semantic error that should be caught at the level of the semantic checker or a syntactic error that should be caught individually.

@kdahlquist suggested that the team should see what JSON is returned in each case of an empty file.  If it is the same, then it seems that it can be handled by Semantic Checker.  If not, then there would need to be individual syntactic tests.  @eileenchoe mentioned this in her post to #362.

dondi:
Yes, I agree that this is a semantic error. Technically, an empty file is "syntactically correct"—i.e., it represents an absence of data, which is considered legal at the structural level. We are the ones deciding that for the particular context of loading a graph, putting up a zero-node, zero-edge graph is most likely an indication that something is not going as intended.

This is analogous to how we don’t consider zero, the empty string, or null/nil to be an erroneous expression per se, but depending on the context, we might interpret it as incorrect.

This choice also reduces code duplication. The semantic checker can look for a zero-node, zero-edge network and issue a warning or error, thus placing this check in one place and avoiding the implementation of this check at the level of every format that we support.

eileenchoe:
Wrote the semantic test for this issue, which returns an EMPTY_NETWORK_ERROR when it detects that the network object has an empty genes array. I interpreted a "blank file" to mean an empty SIF file, an Excel file with a network tab which contains no data, or a GraphML file with no nodes and edges. Semantic test has been tested for all 3 file types and displays the error correctly. The files I generated can be found in test-files > semantic-error-test-files. Submitting the pull req now.

kdahlquist:
For GraphML, the test should be like SIF, an empty file.  Since GraphML is a text format, that is how I would interpret an empty file.  The test you wrote where there were no nodes or edges in the file (but presumably the header info was intact) is a separate test.

For Excel, if there is a blank worksheet called "sheet1", which is what a new workbook file looks like in Excel, presumably triggers a different error where there is no "network" or "network_optimized_weights" sheet.

eileenchoe:
[...]

PR #478, opened 2017-05-04
graphML Syntactic Checker: General graphML Syntax Error

ebachoura:
#363 As discussed on this issue, before this pull request, out servers crash whenever we input a graphML file with any sort of major syntactic error that is caught by the parser that we are using. I have created an error called the General GraphML Syntax Error and made it so that the very first thing that we check for on import is to make sure that there are no errors spewed out by the parser, and if so, then instantly return a network with a General GraphML Syntax Error. Let it be noted that not every syntactic error is caught by this. There is a full list of errors that we need to check for individually and some that are not caught by the parser at all. Find this list under #363

### DIFF
--- a/server/controllers/importers/graphml.js
+++ b/server/controllers/importers/graphml.js
@@ -1,18 +1,83 @@
+
 var constants = require(__dirname + "/../constants");
 var parseString = require("xml2js").parseString;
 var semanticChecker = require(__dirname + "/../semantic-checker");
 
+var graphmlWarnings = {
+  EDGES_WITHOUT_WEIGHTS: {
+    warningCode: "EDGES_WITHOUT_WEIGHTS",
+    errorDescription: "GRNsight has detected that one or more edges in your network are missing numerical weight" +
+    " values. Because the algorithm GRNsight uses for determining the arrowhead type and the color and thickness of" +
+    " the edges requires numerical weight values, your graph will display as an unweighted graph with black edges" +
+    " and pointed arrowheads. If you want to display the network as a weighted graph, please modify your input file" +
+    " to include weight values for all edges."
+  },
+
+  EDGE_DEFAULT_NOT_DIRECTED: {
+    warningCode: "EDGE_DEFAULT_NOT_DIRECTED",
+    errorDescription: "GRNsight interprets the graph as directed unconditionally."
+  }
+};
+
+var graphmlErrors = {
+  GRAPHML_GENERAL_SYNTAX_ERROR: function (error) {
+    return {
+      errorCode: "GRAPHML_GENERAL_SYNTAX_ERROR",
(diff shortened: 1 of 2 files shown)


server/controllers/importers/graphml.js
```diff
@@ -1,19 +1,84 @@
...
 var semanticChecker = require(__dirname + "/../semantic-checker");
 
+var graphmlWarnings = {
+  EDGES_WITHOUT_WEIGHTS: {
+    warningCode: "EDGES_WITHOUT_WEIGHTS",
+    errorDescription: "GRNsight has detected that one or more edges in your network are missing numerical weight" +
+    " values. Because the algorithm GRNsight uses for determining the arrowhead type and the color and thickness of" +
+    " the edges requires numerical weight values, your graph will display as an unweighted graph with black edges" +
+    " and pointed arrowheads. If you want to display the network as a weighted graph, please modify your input file" +
+    " to include weight values for all edges."
```
dondi: Indent unfinished expressions, to show that those lines are “under” the `errorDescription` property name.

server/controllers/importers/graphml.js
```diff
@@ -1,19 +1,84 @@
...
-    graph = result.graphml && result.graphml.graph && result.graphml.graph[0];
+    if (err) {
+      pushRelevantError(err);
+    } else {
+      key = result.graphml && result.graphml.key;
+      graph = result.graphml && result.graphml.graph && result.graphml.graph[0];
+    }
   });
 
+  if (network.errors.length > 0) { return semanticChecker(network); }
```
dondi: Our style dictates that we break this into multiple lines:

    if (network.errors.length > 0) {
        return semanticChecker(network);
    }

server/controllers/importers/graphml.js
```diff
@@ -1,19 +1,84 @@
...
+  function pushRelevantError(err) {
+    var parseError = readErrorFromErr(err);
+    network.errors.push(graphmlErrors.GRAPHML_GENERAL_SYNTAX_ERROR(parseError));
+    // switch(parseError) {
+    //   case "Invalid attribute name":
+    //     network.errors.push(graphmlErrors.UNKNOWN_ERROR);
+    //     break;
+    //   default:
+    //     network.errors.push(graphmlErrors.GRAPHML_GENERAL_SYNTAX_ERROR);
+    // }
```
dondi: Add a comment here for next semester:

    // TODO Ask Dondi about how to do this in a more data-driven manner.

ebachoura:
@dondi Alright I made the changes, let me know if there is anything else.

kdahlquist:
As noted on issue #428, an empty .graphml file crashes the server.  Because @ebachoura is not around this summer, I was wondering if there is a quick fix to adding a test case for this instance since we know that it has a bad outcome.

PR #482, opened 2017-05-05
normalization adjustments courtesy of dondi

anuvarsh:


dondi:
Mysteriously, the file comparison made it look like the entire file changed—not sure why. So I ended up committing this directly so that the proper record of what changed will show up. Thanks, things look good now!

Issue #465, opened 2017-05-02
Changes to zoom slider


kdahlquist:
Now working on making individual tasks from #399.

1. We need a vertical bar on the zoom slider to indicate where 100% is (scale factor 1).
2. We need a label that shows what the current zoom factor is (in terms of percent).
3. The positioning of the zoom slider needs to come up and to the right a little bit so there is some white space surrounding it and it's not flush with the viewport boundary.
4. It was suggested in #399 that using a log transform on the zoom slider might make it behave better.  Was this tried and was it successful?

I know from discussions with @NAnguiano that the zoom functionality is behaving a little erratically, which might make these labels also behave erratically.  I still think we need to have them, however.  As it currently stands, it's not ultra obvious that the slider is referring to zoom.  Adding the current % will clue the user in.

dondi:
Nothing to add here for now; let’s see what emerges then we can tweak from there.

NAnguiano:
Just about done with this. The zoom slider works way better than before, and a vertical bar is set at the scale 1. Just need to have the slider update properly when you're zooming with the mouse wheel, then this is done.

kdahlquist:
After discussion with @NAnguiano, Bostock says himself that zoom is not functioning properly in version 3 of D3, which we are using.  It is supposed to be fixed in version 4, but moving to version 4 will require a complete overhaul of GRNsight.  For example, v4 of D3 is not a single library but a set of smaller libraries that need to be installed independently.  As a test, @NAnguiano tried just using the v4 D3 zoom code in GRNsight and it doesn't work.

Creating the vertical bar for 100% that automatically returns the graph to 100% and showing a numerical zoom factor are sticking points right now that are causing problems.

@kdahlquist says that we can let go of those features for the moment, in favor of simply having a zoom bar that functions without crashing.

kdahlquist:
I've played around with the zoom and it seems to be working OK, with the known issue of it originating with the upper left of the bounding box.  However, using the sun button to center the graph after zooming is not too bad.  I'll go ahead and close this one.  I'm going to open separate issues for creating the 100% bar on the zoom (#484)  and for annotating the zoom factor (#485).  But as agreed, these are not required for the release.

Issue #422, opened 2017-03-29
Checklist for v2.0.x release


kdahlquist:
I occurred to me as I closed a bunch of issues tonight, that we should start documenting the features and bug fixes going into the new release.

Maybe it's time to start a draft release for this (or maybe not).  In either case, our aim is to do a release by the end of the last week of class, i.e., April 28.

I'm assigning this one to @NAnguiano and @anuvarsh (our graduating seniors) to do the honors, but everyone on the team will contribute his/her new features/bug fixes/tests to this.  At our next meeting on Thursday, I will ask each team member to commit to what s/he will have done for the end-of-semester release.

Within this issue, let's keep a tally of all the resolved issues that are going into this release, organized by feature/bug fix/test suite.

This will aid us in getting the Documentation page of the website ready for the new release as well because we are changing the way the GRNsight behaves in some ways and adding new features.

kdahlquist:
I realized that we can't really start the draft release until we are ready because one of the first steps is tagging the release.  So, I renamed the issue.  I'm pasting the release checklist here so we can check items off when it does come time to release.

However, since we have not had a master release since August, there are a lot of new features/bug fixes to document.  I'm going to start the release notes as comments to this issue.  We also need to be thinking about what needs to be changed in the documentation.

- [ ] Merge the branch(es) under development into `master`
    - [ ] Resolve any version control conflicts
    - [ ] Test the final merged version
    - [ ] Perform hotfixes on `master` as needed
- [ ] Tag the final, tested version as `v#.#.#` (all three numbers are required, even if `0`; see _npm_ step below)
    - [ ] Remember that the tag needs to be pushed to GitHub: `git push origin <tag-name>`
- [ ] Create draft release item on Github repository (name it exactly like the tag)
    - [ ] Make release notes: a list of new features/bug fixes with issue numbers covered by this release
- [ ] Publish GitHub release item
- [ ] Publish release to GRNsight home page
    - [ ] Update the version number and modification date on `gh-pages`
- [ ] Increment beta version
    - [ ] Merge released `master` with `beta`, especially if hotfixes/merges were made on `master`
    - [ ] Update version number and modified date on `gh-pages`
- [ ] Publish to _npm_
    - [ ] Update version numbers in all _package.json_ files
    - [ ] _npm_ requires all `.0`s in the semantic version
    - [ ] Check to see that BioJS updates its GRNsight entry
- [ ] Run Istanbul and note new test coverages (optional)
[...]

PR #483, opened 2017-05-09
Final Features for Release

### LINKED ISSUE #461: "Fit to window" doesn't fit to window
I thought it might be easier if we start making these issues more granular to individual behaviors.

When I select "Fit to window", it doesn't.

I expect that the viewport will "snap" to the lower right corner of my browser.

Instead, I actually get a really long viewport with scroll bars.

Since I have my browser window maximized, "fit to window" should not really make a big change, actually.  It would seem that it would come into play when my browser window was smaller.

### LINKED ISSUE #479: Moving the zoom slider and then hitting the center button crashes the graph
See title. Looks like d3 doesn't like it much.

NAnguiano:
Fix #479 
Fix #461 - well, attempted. It's still got some buggy permutations where scrollbars appear when a graph is loaded while "fit to window" is selected, but I couldn't consistently find them and I got tired of fighting with it. They generally don't show up. 
#465 - Zoom slider now reflects accurate zoom level regardless of what's being used to zoom. It's still not the most pleasant thing to use, but it functions as intended. Clicking on it and then using the arrow keys making using it way more pleasant.

### DIFF
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -9,8 +9,9 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
   var $container = $(".grnsight-container");
   d3.selectAll("svg").remove();
 
-  var width = $container.width(),
-      height = $container.height(),
+  var BORDER_OFFSET = 4;
+  var width = $container.width() - BORDER_OFFSET,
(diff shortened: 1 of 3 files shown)


NAnguiano:
Oh, I also removed the normalization ui stuff, as per @kdahlquist's comment on #422.

Issue #417, opened 2017-03-22
Missing Value Error is not being caught when there are missing values in the matrix


NAnguiano:
If there are any missing values in the matrix, a missing value error should be returned. Right now, a missing value error is not being returned. It also seems like the missing value error is being caught in the wrong spot - almost as if it's a catch-all for all of the problems that could arise in the matrix. In the spot that it looks like a missing value error should be caught, it seems as though an invalidMatrixData warning is being returned instead. However, in all missing value test files, there are no errors or warnings being returned at all.

NAnguiano:
@eileenchoe @ebachoura, pinging you guys for this one.

kdahlquist:
@eileenchoe, since you are in a stable state with #362, focus on fixing this bug first.

NAnguiano:
I'm not sure if this is related to this or not, but I managed to get this error to show up while testing pull request #458. I loaded file adjacency-matrix-modifications > extra-column-end-of-sheet-input.xlsx.

![screen shot 2017-04-27 at 7 53 48 pm](https://cloud.githubusercontent.com/assets/6444156/25512536/6308e35a-2b83-11e7-9a74-0819e6a3c939.png)

This isn't a huge bug, but "undefined2" doesn't make much sense.

anuvarsh:
Errors and warnings present the letter and number of the row and column. It looks like it's only catching the number and not the letter.

NAnguiano:
@eileenchoe Weirdness is going on with the test suite. There's an error in the errors array that is straight up just half finished and doesn't actually seem to be used anywhere. It's called "outsideCellError". I'm worried that this testing problem may run a bit deeper than it appears even now...

![screen shot 2017-04-28 at 5 48 31 pm](https://cloud.githubusercontent.com/assets/6444156/25551526/eb92a00a-2c3a-11e7-8eab-f1271dc2ea60.png)
.... I have no idea how this happened, but git blame attributes it to me, so I suppose I have no one to blame but myself :scream:

NAnguiano:
@eileenchoe I found more test weirdness! Namely this little gem in warnings-adjacency-matrix-modifications...
![screen shot 2017-05-02 at 5 12 28 pm](https://cloud.githubusercontent.com/assets/6444156/25644433/c0b5b00c-2f5a-11e7-9c8d-6a258f058476.png)

[...]

PR #486, opened 2017-05-15
ESLint Integration, Code Cleanup + Bug Fixes

NAnguiano:
This is sort of a last hoorah for me I guess!

As promised, the stylistic changes. The code is now consistent style-wise in every single one of the 32-ish javascript files in this project (well excluding some parts of graph.js but I tried). And upon doing this, I discovered some major bugs and patched them, as well as some minor ones.

This is what I started with at 11pm last night:
screen shot 2017-05-14 at 10 58 12 pm

It's now down to a sparkling 0 errors. You can now run "npm run lint" (after npm installing) to see how your code is doing style-wise. If you have an editor like Atom, it will yell at you if you go against the rules of ESLint. I added to our rules list the ones we voted on, as well as some other ones that I felt integrated well with them. 

Ok, so changes:
1. Found and fixed bug noted in #417 where the warnings modal was listing "undefined".
* This bug still exists, but I solved the initial reason I was getting the error - there was an object being mentioned in the error string that wasn't defined in the file referencing it. The object gave the corresponding column letter for whatever the number of the erroring column was (1 = A, 2 = B, etc). With it there, this displays normally most of the time, unless the error is way outside the bounds of what the object can list.
2. Again as noted in #417, I fixed a bug with the warnings error being caused by a typo. It was actually impossible to get a warnings error from an xlsx file beforehand. I've added it in, and corrected a test that was failing because it wasn't there. 
3. Based on that bug fix, I was able to make the test that reads "should return warnings count error" actually return a warnings count error (or 49,062 warnings count errors, to be exact).
4. Fixed a typo in export-tests that was keeping a test from running at all. We now have one additional test as a result, so there are 202 passing tests.
5. Found and fixed a bug in graph.js that made it so that the graph would get weirdly cut off on the right-hand side of the bounding box immediately upon load. 

Also fixed some less important typos, cleaned up some parts of the code, condensed a few things, and tried to make things generally more pleasant to work with. Graphjs is still a bit of a mess, but... I made an attempt.

So yeah. I don't think much should have changed at all. I didn't really do much other than reformat the code so that it's following some consistent guidelines. Everything seems to run just as expected, and all tests pass.  Then again, it's like 5am and I've been coding for the past 6 hours so I'm probably not the best person to ask. I really just wanted to get this done and ready before you all started on the summer stuff. Even if I'm cutting it super close. 😅 

Let me know if I need to change anything. 

Have a great summer, guys :heart: I'll be around


eileenchoe:
Pull Request Checklist
- [ ] All Tests Pass

eileenchoe:
Thanks so much Nicole! The code is looking great, and loads cleaner and more readable. @mihirsamdarshi is fixing some merge conflicts that resulted from the merge of his pull request yesterday but other than that, it's passing all the tests!

Issue #491, opened 2017-05-16
Random tests not passing, as a result of unresolved merge conflicts


eileenchoe:
The following 5 tests are not passing as a unintended result of merge conflicts that @mihirsamdarshi and @eileenchoe were sorting out this morning with @NAnguiano 's pull request. Bookmarking them here so we can return to addressing them separately. They are currently commented out while we are focusing on other issues for the rest of the day. 
- errors-adjacency-matrix-modifications extra-column-end-of-sheet should return warnings count error
- errors-adjacency-matrix-modifications extra-row-end-of-sheet should return errors count error
- errors-adjacency-matrix-modifications extra-row-end-of-sheet-output should return errors count error
- warnings-adjacency-matrix-modifications missing-column-end no warnings
- warnings-adjacency-matrix-modifications sheet-named-incorrectly should return a wrong name warning

eileenchoe:
Another adjacency matrix test issue brought up by the merge. Commented out with intention to fix along with this issue
- value-replaced-with-spaces

kdahlquist:
Elevating this now that the zoom work is finished.

eileenchoe:
Fixed the broken tests. The tests were failing because of a few random values that got changed from the merge. After cross checking with the last working beta pre-merge it seems that only those tests listed above were affected.

kdahlquist:
This hasn't been pulled in via a PR yet, though, right?

eileenchoe:
Working on the PR now, had merge conflicts I'm trying to fix.

eileenchoe:
Live on beta 2.1.8

kdahlquist:
I'm going to assign this review to @dondi.

dondi:
I looked at these errors and they look mostly right except for a redundant test that I found in both _errors-adjacency-matrix-modifications.js_ and _warnings-adjacency-matrix-modifications.js_ for `missing-column-end`. Because the test pertained to warnings, I removed it from the _errors_ file and kept just the one in the _warnings_ file. This was committed and pushed directly to `beta`.

The change reduces the test count from 207 to 206, but that is as it should be because it was the same test being performed twice.

Review completed; removing `review requested` tag.

kdahlquist:
I'm also closing because this is done.

Issue #386, opened 2017-02-02
Semantic checker needs to be integrated into SIF and GraphML importers


dondi:
The SIF and GraphML import routes (_import-controller.js_) do not run their output through the semantic checker. This should be done so that our semantic checks our consistent across all formats. The best way to integrate this is to refactor the import helper function in _import-controller.js_.

kdahlquist:
Assigning this to @mihirsamdarshi and bumping it up to priority 0.

kdahlquist:
@mihirsamdarshi is working on this right now.  Says will be done by next week.

kdahlquist:
@ebachoura and @eileenchoe are in fact working on this exact issue because they thought that this issue was a sub-issue of #362 / #363. @mihirsamdarshi needs talk to them.

kdahlquist:
@ebachoura and @eileenchoe, I need clarification as to the status of this one.  Is it done? or is there more work needed?

eileenchoe:
@mihirsamdarshi is reviewing the semantic checker integration for SIF an GraphML and will update status when finished

mihirsamdarshi:
reviewed, it works

kdahlquist:
@dondi has not had time to review this one, so I did some quick checks on what I think should be semantic errors.  
* I made a .sif file with a node name that was too long; that brought up the error as expected.
* I made a .sif file with a duplicated row, which should trigger the duplicate gene name error, but it still loaded.  So, for whatever reason, this is not being caught by the semantic checker.  I'm attaching the file I made here.
[duplicate-gene-name.zip](https://github.com/dondi/GRNsight/files/945554/duplicate-gene-name.zip)

To thoroughly check this, we probably need to try making the .sif and .graphml versions of an assortment of .xlsx test files and load 'em up and see what happens.

kdahlquist:
Merged with PR #492 and refreshed on beta v 2.1.2

dondi:
I reviewed this and the flow looks good; semantic check is bypassed if any syntax errors are found.

Issue #362, opened 2017-01-18
Add "syntactic" tests for SIF format


kdahlquist:
Need to add more tests for SIF format input files.  @ebachoura and @eileenchoe can either split SIF and graphml or both work on both.  Discuss with @dondi this week.

dondi:
@dondi gave the group an overview/tour of the existing code and pointed out where syntactic errors can be detected then reported. @ebachoura and @eileenchoe will take a look at the code and decide how to tackle this.

dondi:
(the above note also covers #363)

eileenchoe:
Decided that I will focus on SIF tests while @ebachoura will focus on GraphML tests #363 for now. Ran some initial tests using existing SIF files in test-files to understand code structure of the importer.

kdahlquist:
Here's how to approach this one:
1. Read the documentation on the Documentation page of the website to fully understand the specification for this format in GRNsight (which is slightly more specific than the general specification for this format).
2. Look in the test suite to see what tests are already there from @dondi's initial implementation
3. Enumerate the cases for which we don't have test coverage, looking at the cases for the Excel format may be helpful here.  Perhaps we could have a meeting with just @kdahlquist, @eileenchoe, and @ebachoura to go over likely edge cases to look for.  The errors and warnings glossaries on the wiki will also be helpful; it would be a good place to add any new cases.
4. Then write the tests.

dondi:
@eileenchoe has successfully implemented one syntactic warning and will just keep proceeding with more. At the meeting, @dondi showed her how she can include unit tests for this in the codebase.

kdahlquist:
Warnings that have already been written are now showing up in the warnings box on GRNsight when use uploads bad file.

eileenchoe:
@kdahlquist here are times that I am available this week to meet about this issue. I'm hoping we can meet before my research block on Wednesday, so I can get a significant amount of work done for this issue this week.
- Mon 11:20AM - 12:30PM,  2:20PM-4:20PM
- Tues 4:10PM-5:30PM 
- Thurs 8:30AM - 9:30AM, 4:10PM-5:30PM

kdahlquist:
@eileenchoe, I can do Tuesday, 4:10-5:15.  Please come meet me in my office, LSB 289.

eileenchoe:
@kdahlquist Great, see you then.

kdahlquist:
@eileenchoe and @kdahlquist met to talk about tests for SIF.

[...]

Issue #326, opened 2016-08-18
Update state diagram/automatic client-side test generator


kdahlquist:
I am thinking that our state diagram might need to be updated due to the changes we made for importing/exporting SIF and GraphML.

kdahlquist:
@mihirsamdarshi is going to consult with @NAnguiano to work on this.

mihirsamdarshi:
Looked over current diagram, now trying to figure out where to put the new functionality, and how it would likely fit into the current diagram.

kdahlquist:
This is something @NAnguiano could possibly help with because she is the main architect of the diagram.

kdahlquist:
@mihirsamdarshi is redoing the diagram right now in Illustrator because the original program LucidChart @NAnguiano used maxed out the free version.

We need to look into other options like yED or Gephi for a diagramming tool that is scalable.  @mihirsamdarshi should take 15-20 minutes to evaluate each of those to see if they are viable in terms of ease of use and how the diagram will look.

mihirsamdarshi:
@yshin4 can you text me please 4087056761. imma delete this as soon as i get that.

kdahlquist:
@mihirsamdarshi will use yED and aims to have all the old stuff done by next week.

kdahlquist:
Almost done... @mihirsamdarshi, can you get this wrapped up this week?

mihirsamdarshi:
done.

kdahlquist:
reassigned to @ebachoura and @eileenchoe to add feature to show/hide all weights.  They need to talk to @mihirsamdarshi to get a walkthrough on the diagram and to @anuvarsh to get a walkthrough on the feature.

dondi:
This conversation has taken place and the addition of the new states is now in progress on the yEd file.

eileenchoe:
After analyzing the current state diagram, and getting a walkthrough on the new feature (edge weight settings) @ebachoura and I have realized that adding this feature (and any future new feature) will increase the complexity of the state diagram by a factor of two or more. Additionally we've noted that, this feature only affects weighted networks but in its current implementation, is also available for unweighted and unloaded networks (but doesn't induce a change). Waiting to discuss for further direction on how to most effectively implement this new feature in the state diagram.

kdahlquist:
I think we will need @dondi to weigh in on dealing with the complexity of the state diagram.

[...]

PR #492, opened 2017-05-17
general syntactic and semantic test bug fixes and integration

eileenchoe:
#428 Semantic test written for blank file.

#362 SIF syntactic checker fixes
- No duplicate errors thrown
- Wrote up corresponding weighted SIF files to test new syntax checking

#386 Modified integration of semantic checker to only run if no syntactic errors were detected

Other:
#491 5 tests are not passing as a result of merge conflicts. I've commented these tests out for now, they are related to #417 
#326 Initial work on defining states for testing script generator


server/controllers/importers/graphml.js
```diff
@@ -2,7 +2,7 @@ var constants = require(__dirname + "/../constants");
 var parseString = require("xml2js").parseString;
 var semanticChecker = require(__dirname + "/../semantic-checker");
 
-/* 
+/*
```
dondi: Per our emerging style conventions, can we just delete this block of code rather than commenting it out? We can always retrieve from `git` in case we want this back.

server/controllers/importers/graphml.js
```diff
@@ -208,5 +208,10 @@ module.exports = function (graphml) {
         });
     }
 
-    return semanticChecker(network);
+
+  if (network.errors.length === 0) {
+      return semanticChecker(network);
+  }
+  return network;
+
```
dondi: More compact, and clearer on the intent that we are returning something for sure:

    return (network.errors.length === 0) ? semanticChecker(network) : network;

server/controllers/importers/sif.js
```diff
@@ -11,26 +11,21 @@ module.exports = function (sif) {
     var warnings = [];
     var errors = [];
 
-    // Empty SIF files must return a network, thus must contain some data.
-    if (!sif) {
-        sif = " ";
-    }
+    // Workaround for empty SIF file handling
+    var emptySifFile = sif === "";
+    sif =  sif ? sif : " ";
```
dondi: This type of conditional expression is the same as the shorter:

    sif = sif || " ";

server/controllers/importers/sif.js
```diff
@@ -11,26 +11,21 @@ module.exports = function (sif) {
...
-    if (sif.match(/(?=.*[\t]{2,}?[^\n\t]).*/g)) {
-        errors.push(constants.errors.SIF_STRAY_DATA_ERROR);
-    }
-
-    // Stray data detected when there are 2 or more consecutive new lines
-    if (sif.match(/[\n]{2,}/g)) {
+    // OR Stray data detected when there are 2 or more consecutive new lines
+    if (sif.match(/(?=.*[\t]{2,}?[^\n\t]).*/g) || sif.match(/[\n]{2,}/g)) {
         errors.push(constants.errors.SIF_STRAY_DATA_ERROR);
     }
```
dondi: Good catch for code consolidation! 👍

server/controllers/importers/sif.js
```diff
@@ -122,20 +122,23 @@ module.exports = function (sif) {
                 });
             }
         });
+  }
+
+  var network = {
+    genes: emptySifFile ? [] : genes.map(function (geneName) {
+      return { name: geneName };
+    }),
```
dondi: Is this the only use for `emptySifFile`? If this is its sole purpose, why can’t `genes` just be left as `[]` if the SIF is empty? An empty list will `map` to an empty list.

server/controllers/importers/sif.js
```diff
@@ -122,20 +122,23 @@ module.exports = function (sif) {
...
+    warnings: warnings,
+    sheetType: networkType.sheetType,
+    positiveWeights: [],
+    negativeWeights: []
+  };
+
+  if (network.errors.length === 0) {
+      return semanticChecker(network);
+  }
+  return network;
```
dondi: Same conditional expression as suggested previously.

test/errors-adjacency-matrix-modifications.js
```diff
@@ -35,13 +35,13 @@ describe("errors-adjacency-matrix-modifications", function () {
...
-            test.warningsCountError("test-files/adjacency-matrix-modifications/extra-column-end-of-sheet-output.xlsx", 49062);
-        });
-    });
+    // ISSUE #491
+    // describe("extra-column-end-of-sheet", function () {
+    //     it("should return warnings count error", function () {
+    //         test.warningsCountError("test-files/adjacency-matrix-modifications/extra-column-end-of-sheet-input.xlsx", 49062);
+    //         test.warningsCountError("test-files/adjacency-matrix-modifications/extra-column-end-of-sheet-output.xlsx", 49062);
+    //     });
+    // });
```
dondi: OK, so this is one of the deferred failing tests. I’m OK with leaving this commented out if they are to be addressed immediately after.

test/errors-adjacency-matrix-modifications.js
```diff
@@ -64,19 +64,20 @@ describe("errors-adjacency-matrix-modifications", function () {
...
-            this.timeout(10000);
-            test.errorsCountError("test-files/adjacency-matrix-modifications/extra-row-end-of-sheet-output.xlsx", 1);
-        });
-    });
+    // describe("extra-row-end-of-sheet-output", function () {
+    //     it("should return errors count error", function () {
+    //         this.timeout(10000);
+    //         test.errorsCountError("test-files/adjacency-matrix-modifications/extra-row-end-of-sheet-output.xlsx", 1);
+    //     });
+    // });
```
dondi: Ditto commented-out #491 tests.

kdahlquist:
I would rather not delete the commented out tests because I'm expecting that they will get fixed soon.  I don't want these to drop off our radar screen.

dondi:
Right; it wasn’t the tests that were commented out under #451 that I requested be deleted. Those can stay, understandably. The commented-out block to be deleted isn’t a unit test but a block of code that seems to be permanently gone or otherwise moved elsewhere.

server/controllers/importers/graphml.js
eileenchoe: It looks like Eddie has a few blocks of code commented out here so that he can return to where he left off on GraphML syntactic tests after the summer.

server/controllers/importers/sif.js
eileenchoe: Yes, it's just used for emptySifFile. I set sif = " " (for lack of something better to set it to so it doesn't throw errors in the parsing) when the importer detects that the file is completely empty. Without this, genes was returning [ " " ]. Sort of a workaround I'll see if I can come up with a cleaner way to do this.

mihirsamdarshi:
- [ ] All Tests Pass

server/controllers/importers/graphml.js
dondi: Ah OK…it wasn’t apparent that this was something he was going to come back to. Let’s try to establish a convention that when we are commenting out code rather than deleting it, we explain why. (like you did with the #451-related commented-out tests)

server/controllers/importers/sif.js
dondi: Ah I didn’t realize that we get a `[" "]`. Seems like a bug in how we parse then; a lone space should produce nothing. But that can be a subsequent task then.

PR #494, opened 2017-05-17
Normalization

mihirsamdarshi:


dondi:
This was reverted, and although this may well have been an honest mistake, I can‘t help but observing: that *the time that the pull request was issued and the time that the pull request was merged _are virtually identical_*.

This tells me that there was virtually no thought/review done between issuing the pull request and merging it. _This is not how pull requests are supposed to be done._ At the very least, _review the pull request in its entirety_, ideally _by someone else_, before pulling the Merge trigger.

The very point of pull requests is to allow some thought before they are made permanent. They provide a safe staging area for changes without touching the codebase yet. If pull requests are issued and immediately merged, then their purpose is undermined: one may as well commit and push directly if that’s what was going to be done anyway.

dondi:
^^^^^ @mihirsamdarshi

PR #495, opened 2017-05-17
Revert "Normalization"

mihirsamdarshi:
Reverts dondi/GRNsight#494


dondi:
@mihirsamdarshi This sequence of PRs looks like something was done by mistake (#494, #495). When something like this is done, please add a more explicit description to the pull request. To a casual reviewer of PR notifications, it looks like a bunch of files that we actually need are being removed. It’s concerning when there is no explanation for something that looks pretty drastic.

mihirsamdarshi:
Sorry @dondi, I was doing some minor addition to my branch from normalization, and didn't properly pay attention to what the head and base forks were. It was such a minor change I didn't feel it necessary to go through a full pull request checklist, at least not on my fork.

dondi:
OK @mihirsamdarshi, that's fine, but do include an explanation like that in the future, right when describing the pull request. To someone who doesn't know the sequence of events, something that says "133 files changed" will be cause for concern. The good news is that `git` doesn't lose anything, but that doesn't mean we shouldn't still be watchful for unintended changes.

Issue #508, opened 2017-05-25
bug when changing viewport size when fixed


eileenchoe:
While working on #468 I came across this bug, and realized that it also happens on beta. The sequence is: Load Graph > check "Restrict Graph To Viewport" > Change Viewport Size.

It's pretty serious since a whole bunch of these errors come up, and the graph disappears from the viewport except for a singe node at the upper left hand corner. 

I'll be trying to fix this as I work on #468.

<img width="341" alt="screen shot 2017-05-25 at 2 31 36 pm" src="https://cloud.githubusercontent.com/assets/16827883/26471550/ef7edc2a-4156-11e7-8332-48694fe0aefb.png">

eileenchoe:
Fixed. Will merge with PR for #468 since I'm still working on other changes for that issue

eileenchoe:
Fix live on Beta v2.1.6

kdahlquist:
I followed the instructions described in the bug report and didn't see a problem.  One thing to note for later, the last instruction "Change viewport size" is a little ambiguous; it would be better to say "Change viewport size from small to large", for example because that is a specific sequence.  For example, my monitor defaulted to medium, so I would change from medium to small or medium to large.

Issue #485, opened 2017-05-10
Annotate zoom factor on zoom slider


kdahlquist:
As noted in issue #465, this is broken out as a separate issue to be assigned to someone in the future.  It would be good to indicate the actual zoom factor near the slider like what is done for the force graph parameter sliders.

eileenchoe:
Tackled this one this morning to start figuring out how zoom was implemented. Here's how it looks when a graph is initially uploaded. @kdahlquist when we meet later today I have some questions about how this percentage changes when toggling restrict graph to viewport
<img width="344" alt="screen shot 2017-05-19 at 10 12 59 am" src="https://cloud.githubusercontent.com/assets/16827883/26258963/c40d54ac-3c7b-11e7-9f25-640d92d7c050.png">

eileenchoe:
Zoom percentages are now consistent between fixed and adaptive viewport. The zoom percentage goes from 0% to 200%, starting at the default of 100%. For adaptive viewport, the user is allowed to maximally zoom to 200%, however when the viewport is fixed, zoom is capped at 100%, with the viewport and bounding box the same. 

There are two ways to zoom: the mouse scroll and the zoom slider. 

At this point, it's working fine with the mouse scroll zoom, but there's an issue with the zoom slider when the user tries to zoom to 100% with the *fixed* viewport. The way I coded it simulates disabling the slider by not allowing the zoom event to be triggered if it detects that the viewport is fixed AND the slider zoom value is greater than ADAPTIVE_MAX_SCALE, which is equal to 100% zoom. The zoom event is fired using the *on input* event listener on the zoom slider. What's weird is that when the user zooms using the slider slowly, it correctly renders the 100% view, however when the zooming is done quickly, the graph does not quite get to the 100% zoom. I think that the event listener is not detecting the change fast enough, but have not come across a solution for this. @dondi is there anything that you suggest?

kdahlquist:
@dondi willl look into the aforementioned behavior to see if something can be done to keep things in sync.

eileenchoe:
A solution I found was to make the slider less sensitive by only allowing the discrete options 0%, 25%, 50%, 75% ... 200%. @kdahlquist is this ok or do you think the user should have the full range of zoom values?

eileenchoe:
Actually this doesn't make too much sense because the mouse wheel slider is still more sensitive, so the zoom percentage won't actually capture the true zoom percentage when the user is using the mouse wheel instead.

[...]

Issue #480, opened 2017-05-04
Bounding box sticks to edges of viewport on zoom


kdahlquist:
Bounding box views to edges of viewport on zoom.

Conditions that cause this:
* fix graph to viewport
* zoomed out so you see a grey bounding box
* move the bounding box to the edge of the viewport
* zoom in and then zoom out and you will see that the bounding box sticks to the viewport edge.

@NAnguiano says that this is happening because she is manually setting the translate and D3 doesn't like it.  D3 remembers the point at which the user started zooming back in and it doesn't detach the graph from the viewport until you reach that point.

eileenchoe:
Just discussed this with @kdahlquist. The change to the zoom slider which zooms with respect to the center of the viewport fixed this problem. However, grabbing and moving the bounding box when the viewport is fixed and zoom level is <100% is disabled because we are only allowing the bounding box to "grow" into the size of the viewport radially, which we've agreed is ok for now.

eileenchoe:
Also, removing cursor grab so that the user knows dragging isn't allowed when graph is restricted to viewport.

eileenchoe:
Live on Beta v2.1.6

kdahlquist:
Confirmed and closing.

Issue #473, opened 2017-05-02
Disable mousewheel zoom after zoom slider is fully functional


kdahlquist:
This has been noted by @kdahlquist and @NAnguiano.  Immediately upon graph load, the mousewheel function switches over to controlling the zoom level, which is disconcerting if the user meant instead to scroll the browser window.

This should be changed somehow, but it is not in the critical path to release.

kdahlquist:
I think at the meeting today, the consensus was that we should disable the mousewheel zoom entirely.  Maybe not until we get the zoom slider working the way we want, but soon after.

eileenchoe:
Did you know that double click in the viewport also zooms GRNsight? I just came across this right now. It zooms with respect to the mouse pointer as the center. I am also disabling this one along with the mousewheel zoom because it interferes with the way that I implemented the slider zoom.

kdahlquist:
Yikes!  Was just writing this comment when @eileenchoe came in.  Definitely disable this!

eileenchoe:
Live on Beta v2.1.6

kdahlquist:
This was easy to test (and quite a relief)!  Closing!

PR #509, opened 2017-05-26
Zoom Slider + Bug Fixes

### LINKED ISSUE #468: In D3, center function on zoom only works with mousewheel, not slider
This has been noted in a couple of different contexts:

When you zoom in with the range slider, it doesn't zoom to the center. It instead zooms to the top right corner. This is because the center function on zoom only works on mousewheel scrolling, not regular scrolling. @NAnguiano is trying to figure out a way to imitate that behavior for the slider, without needing to do any overly complicated math that may slow it down.

This bug is not in the critical path and can be solved later.

eileenchoe:
#468 Implemented center zoom for slider
#473 Disabled mouse wheel and double click zoom
#480 Bounding box sticking to edges bug got fixed as a result of center zoom
#471 Modified adaptive bounding box expansion algorithm to prevent the "Rapunzel effect"
#485 Small change for zoom slider annotation
#508 Fixed viewport size changing bug

#428, #363 Empty network error for graphml blank file

### DIFF
--- a/server/controllers/importers/graphml.js
+++ b/server/controllers/importers/graphml.js
@@ -79,6 +79,10 @@ module.exports = function (graphml) {
         if (err) {
             pushRelevantError(err);
         } else {
+            // Quick fix to handle completely empty GraphML file. #428
+            if (!result) {
+                return semanticChecker(network);
+            }
             key = result.graphml && result.graphml.key;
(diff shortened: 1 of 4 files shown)


mihirsamdarshi:
Pull Request Checklist
- [ ] All Tests Pass

Issue #318, opened 2016-08-05
Long node labels mess up self-regulatory arrows


kdahlquist:
Long node labels mess up self-regulatory arrows in terms of their placement and the arrowhead alignment on the edge.
![long-node-label-autoregulatory-arrow](https://cloud.githubusercontent.com/assets/6463867/17447143/b4539d70-5b01-11e6-99bd-5ffe43f7ef70.png)

Probably need to also reduce whitespace on edges of label.

Low priority.

kdahlquist:
Assigning this one to @yshin4 to work on after she finishes with #333.

yshin4:
Noticed even shorter gene name nodes (length 4) has an arrow that is off as well when it is bold, will look more into it once the #333 is finished.

yshin4:
@kdahlquist 

Does this mean you would like the fixed graph to look like the top node, bottom node, or in some other form?

![image](https://cloud.githubusercontent.com/assets/16246811/24938489/59c53652-1eec-11e7-8069-94cb1dc02c0a.png)

kdahlquist:
@yshin4, not really either of those drawings.  The self-regulatory arrows should look like they do now for smaller nodes, like this:
![self-regulatory-edge-long](https://cloud.githubusercontent.com/assets/6463867/24978434/89a13ff8-1f85-11e7-8e9b-0a7db0410532.jpg)

kdahlquist:
@yshin4, this is a higher priority than either #392 or #188.  Both of those are in a better state than this one.  See if you can make some headway on this one for the release and we can come back to the other ones later, if need be.

yshin4:
Got an error and can't upload any files to the website on my fork server. #455

kdahlquist:
I'm de-labeling this "for next release" because, presumably, @yshin4 won't be able to do any further work on this for the semester.

yshin4:
Yes, sorry! I got hit by five final exams and one project. My apologies for not being able to make any progress.

yshin4:
Taking a look at graph.js where getSelfReferringEdge is being set. My original plan was to change the x value of the self referring edge when the node is long (larger than 150px box/text width) but can't find a way to move the x position. Currently investigating why SELF_REFERRING_Y_OFFSET works but changing values for selfReferringEdgeWidth doesn't work.

yshin4:
[...]

PR #529, opened 2017-09-13
Arrowheads Issues

yshin4:
Changed values and made changes to fix the issues: #392  #318.

 On chrome it looks like: 
grnsightchrome

and on Firefox it looks like:
grnsightff

it looks fine on edge as well.


eileenchoe:
@yshin4 please change the base branch to `beta`

yshin4:
Sorry, i changed it. Thank you for telling me.

web-client/public/js/graph.js
```diff
@@ -1049,11 +1049,12 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
                     var largeArc = 0;  // 1 or 0
                     var sweep = 1;     // 1 or 0
                     var offset = parseFloat(d.strokeWidth);
+                    var nodeShift = 1.033;
```
dondi: Can the initial value of `nodeShift`, `1.033`, be given a name? Even something like `INITIAL_NODE_SHIFT` or `DEFAULT_NODE_SHIFT` sounds appropriate. We can change to using `const` when/if we move to ES6.

web-client/public/js/graph.js
```diff
@@ -1070,7 +1071,10 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
 
             // For whatever reason the arc collapses to a point if the beginning
             // and ending points of the arc are the same, so kludge it.
-                        x2 = d.source.x + d.source.textWidth / 1.2;
+                        if(d.source.textWidth > 135){
```
dondi: Are you running the linter? (`npm run lint`) This line should be flagged by it. If it isn't, then we should look into whatever setting in the linter will ensure that this is properly spaced. i.e.,
```javascript
if (d.source.textWidth > 135) {
```

web-client/public/js/graph.js
```diff
@@ -1070,7 +1071,10 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
+                            nodeShift += 0.07;
+                        }
+                        x2 = d.source.x + d.source.textWidth / 1.2 * nodeShift;
```
dondi: Can the values `135`, `.0.07`, and `1.2` be given names? We’re in tech debt with regard to magic numbers so it would be nice if we can avoid adding more…

web-client/public/js/graph.js
yshin4: No problem, I changed it. I tried to match the other variables under the comment // Defaults for normal edge. when I first named it.

web-client/public/js/graph.js
yshin4: That was my bad! Changed it.

web-client/public/js/graph.js
yshin4: You are right. I assigned names for each values and changed the code accordingly. I will look out for magic numbers in the future!

web-client/public/js/graph.js
yshin4: On the side note, there are 325 errors when I run lint.

eileenchoe:
Pull Request Checklist

eileenchoe:
@yshin4 The demo files aren't loading right when I ran your code locally. Will have to investigate during our meeting tonight.

yshin4:
Fixed it :)

eileenchoe:
@yshin4 cool, everything looks good. Merging PR now.

eileenchoe:
Live on beta v 2.2.2

PR #530, opened 2017-09-13
client side testing script generator

eileenchoe:
#326 
- Completed client side testing script generator
- Removed unused files
- Added pdf export capability with `markdown-pdf`
- Updated `data.json` with normalization & grey slider UI additions


dondi:
Looks good to me from the code standpoint.

eileenchoe:
None of the changes affect web-client or server, but you'll have to npm install again to get the markdown-pdf dependency to test this @yshin4 or @ebachoura

eileenchoe:
@dondi how do we deal with merge conflicts with the package-lock.json file?

dondi:
@eileenchoe Yes, I noticed that…they mean well (https://docs.npmjs.com/files/package-lock.json) but might not be accommodating for differences in package installations that are due to running on different platforms.

Let’s play ball with their directive for now. This branch will have to be merged using the command line instructions that they mention above. Maybe, after sufficient runs, we will all converge on the same package-lock? Based on the link above, it sounds like this is supposed to eventually happen.

PR #542, opened 2017-10-17
Formula applied for the distance issue for the blunt markers

### LINKED ISSUE #318: Long node labels mess up self-regulatory arrows

yshin4:
I fixed #318 with a better solution, instead of using a condition dondi and I discussed a math formula to figure out which direction the target node and the source node is connected. Unlike the condition I used, which made the vertical spacing too wide sometimes, this formula allows me to add additional distance only when the source is to the right of the target node.


web-client/public/js/graph.js
```diff
@@ -726,12 +726,19 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
-        var MINIMUM_DISTANCE_LEFT = d.strokeWidth > 10 ? 18 : 14;
-        var MINIMUM_DISTANCE = d.target.centerX < d.source.newX ? MINIMUM_DISTANCE_LEFT : 8;
     // For arrowheads when target node is to the left of source node
         var LEFT_ADJUSTMENT = 7;
+        var MINIMUM_DISTANCE = 8;
+
+        var targetStartX = d.target.centerX + d.target.textWidth / 2;
+        var currentPointX = (targetStartX - d.target.centerX) / (d.source.newX - d.target.centerX);
+        var currentPointY = (1 - currentPointX) * d.target.centerY + currentPointX * d.source.newY;
+        var nodeHeight = 30;
```
dondi: Looks like `nodeHeight` is functionally a constant, so can be moved up. The full `nodeHeight` is also not actually used, but is divided by two, so perhaps the constant itself can simply be:
```javascript
    var NODE_HALF_HEIGHT = 30 / 2;
```
(we keep the division here to emphasize that the actual height is 30 and we are dividing by 2)

web-client/public/js/graph.js
```diff
@@ -517,12 +517,12 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
...
                      xOffsets = {
                          2 : 14, 3 : 15, 4 : 15, 5 : 15, 6 : 16, 7 : 16.5, 8 : 16.5,
-                         9 : 17, 10 : 17.5, 11 : 18, 12 : 19, 13 : 19.5, 14 : 16
+                         9 : 17, 10 : 17.5, 11 : 18, 12 : 19, 13 : 19.5, 14 : 20.5
                      };
                  } else {
                      xOffsets = {
                          2 : 13, 3 : 13, 4 : 13.5, 5 : 14, 6 : 15.5, 7 : 16.5, 8 : 17,
-                         9 : 16, 10 : 17, 11 : 17, 12 : 16, 13 : 18, 14 : 18
+                         9 : 16, 10 : 17, 11 : 17, 12 : 18, 13 : 18, 14 : 19
```
dondi: Could this be the answer to addressing my question in https://github.com/dondi/GRNsight/issues/318#issuecomment-337346596? The numbers don’t seem to bear out though because it looks like there is only a 1-pixel difference between going from 8 to 9 vs. 9 to 10. Still, I noticed this change so it got me wondering.

yshin4: This was to center the edges.

web-client/public/js/graph.js
yshin4: Fixed, thank you.

Issue #241, opened 2015-10-09
Data structure design for reading expression data from GRNmap input/output sheet


kdahlquist:
Our first task for being able to color nodes with expression data is to read the data from the GRNmap input/output workbook.

We need to design the data structure.  We will make a small test workbook with just 3-4 genes with dummy data to use in test-driven design.

kdahlquist:
Reassigned this to @NAnguiano because she is doing this for her 402 project.  Look at the GRNmap project for small test files that can be used for this.  @kdahlquist will confirm the documentation of the new format that GRNmap is using so that there won't be further changes we have to deal with.

NAnguiano:
@kdahlquist : Has the documentation of the new format been confirmed?

kdahlquist:
I haven't had a chance to look at it yet, but I think it is "close enough" for you to start your work.  The relevant wiki pages are:

https://github.com/kdahlquist/GRNmap/wiki/How-to-format-the-input-file-for-GRNmap

https://github.com/kdahlquist/GRNmap/wiki/How-to-interpret-the-output-file-for-GRNmap

https://github.com/kdahlquist/GRNmap/wiki/New-Naming-Convention

NAnguiano:
Alright, I'll start working adding this into the spreadsheet controller then

anuvarsh:
In anticipation of the discussion we were planning to have at our next meeting regarding the data structure design, I have reviewed the previously mentioned wiki pages about GRNmap's input and output sheets.

kdahlquist:
@eileenchoe and I met this morning to discuss this.  One conclusion we drew was that it might be best to design a data structure to read and store the entire output workbook from GRNmap, even if we are not going to use all the data right away.

kdahlquist:
This will be the next priority for @eileenchoe after the release (#557).  @eileenchoe and @kdahlquist will probably need to meet again to document the requirements.

dondi:
@eileenchoe and @kdahlquist will meet on Friday 1/26 to discuss the data structure, which will be a JSON extraction of a GRNmap workbook.

kdahlquist:
Here is an example of a GRNmap output workbook: 

https://github.com/kdahlquist/DahlquistLab/blob/master/data/Spring2018/MO.LK%20Edge%20Deletion%20Data/Control%20Experiments/Missing%20Data%20beta/15-genes_28-edges_db5-MO-LK_Sigmoid_estimation_missing-values_output.xlsx

[...]

Issue #566, opened 2018-02-04
Node Coloring


eileenchoe:
Nodes should be colored to visualize the gene expression data extracted from the GRNmap input or output workbook with #241.

Some initial design considerations:
- Color scheme should not conflict with the cyan / magenta edges for weighted networks
- Coloring should not obscure the gene name inside the node
- Corresponding UI menu should be added to toggle/adjust node coloring options
  - User should be able to chose which strain (e.g. wt, dgln3, etc.) to use for expression viz
- If experimental AND simulated expression data are present in the output worksheet, visualization should facilitate visual comparison. Experimental data are in `*_expression` sheets, while simulated expression data is stored in `*_optimized_expression` sheets.

eileenchoe:
<img src="https://user-images.githubusercontent.com/16827883/35782997-96b7e014-09b5-11e8-8398-1bcb6991a0b2.JPG" width="200">
Here's a sketch of ideas for node coloring methods before I start prototyping with d3.

(1) and (4) are options for visualizing experimental data only
(2) (3) and (5) are options for visualizing both experimental and simulated data.

I'm leaning toward (4) and (5) as they offer good readability of the gene name.

eileenchoe:
Bookmarking this link for [handy built in methods for data structures in D3-collections](https://github.com/d3/d3-collection).

eileenchoe:
Looked into the [d3-chromatic library](https://github.com/d3/d3-scale-chromatic#diverging) which offers diverging color schemes (like red -> blue with range [0,1]) which can be continuously interpolated or discretized between 3-11 color "bins." 

Normalization:
- Expression values will have to be normalized such that 0.5 on the color scale corresponds to 0 for expression value.
- The `[-/+]abs(max(expression_values))` should "map to" the 0 and 1 endpoints of the color scale.
- Thus, if expression values were between [-1.34, .45] it's corresponding color mapping should normalize and scale the "positive range" [0 - 1.34) => [0.5 - 1]

kdahlquist:
Need to run the colors by some color blind users.  @kdahlquist thinks that we will choose either pink/green, purple/teal, or red/blue.

Option numbers 1 and 3 for the node coloring are the ones we need.

In terms of normalization, the danger is that for graphs where all genes have low expression values, the user will see bright colors and assume that the genes are actually changing more than they are.  Some options for the user:
- normalize based on the absolute value of the largest magnitude fold change
- allow user to set normalization factor (like we do for edges)
[...]

PR #567, opened 2018-02-19
data structure redesign

eileenchoe:
#241:
- wrote a parser for expression, 2-column, and metadata sheets for a GRNmap input or output worksheet
- wrote corresponding unit tests
- integrated parser into `spreadsheet-controller.js`, simply appending the new data to the existing `network` object


eileenchoe:
Additionally, I refactored the `drawGraph()` function parameters to take the entire network data structure instead of passing in individual parts of the network object like `network.links` and `network.warnings` separately. This is in preparation for #566 where I'll need the additional data from the network object for node coloring.

yshin4:
Pull Request Checklist

yshin4:
PR Checklist complete, everything works as it should.

server/controllers/additional-sheet-parser.js
```diff
@@ -0,0 +1,89 @@
...
+var EXPRESSION_SHEET_SUFFIXES = ["_expression", "_optimized_expression", "_sigmas"];
+
+var isExpressionSheet = function (sheetName) {
+    for (var i = 0; i < EXPRESSION_SHEET_SUFFIXES.length; i++) {
+        var suffix = EXPRESSION_SHEET_SUFFIXES[i];
+        if (sheetName.includes(suffix)) {
+            return true;
+        }
+    }
+    return false;
```
dondi: How about
```javascript
    return ELEMENT_SHEET_SUFFIXES.some(function (suffix) {
        return sheetName.includes(suffix);
    });
```

server/controllers/additional-sheet-parser.js
```diff
@@ -0,0 +1,89 @@
...
+};
+
+// Helper function, to be replaced with the includes() method when GRNsight is ported to ES6
+var isInArray = function (value, array) {
+    return array.indexOf(value) > -1;
```
dondi: The `includes` function looks like it has wide enough support that we can just use it: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

server/controllers/additional-sheet-parser.js
```diff
@@ -0,0 +1,89 @@
...
+    return array;
+};
+
+module.exports = function (workbook) {
+
+    var output = {};
+    output["expression"] = {}; // expression data
+    output["meta"] = {};
+    // TODO: Need better name to generalize 2-column data
+    output["test"] = {}; // 2-column data
```
dondi: Why isn't this just:
```javascript
var output = {
    expression: {},
    meta: {},
    test: {}
};
```
.....I think this conveys the structure a little better.

server/controllers/additional-sheet-parser.js
```diff
@@ -0,0 +1,89 @@
...
+                if (geneName) {
+                    var rowData = sheet.data[j].slice(1);
+                    // Sometimes, missing data is at the end of the row. In this case, pad the
+                    // array with nulls
+                    if (rowData.length < numberOfDataPoints) {
+                        fillArray(null, rowData, numberOfDataPoints);
+                    }
+                    geneData[geneName] = rowData;
+                }
+            }
```
dondi: It feels like each of these cases should be decomposed into separate functions. It will make this loop easier to read, essentially "for each sheet, do one of three things," and each "thing" is a separate function.

Decomposition will also make this portion easier to unit-test.

server/controllers/spreadsheet-controller.js
```diff
@@ -432,6 +433,12 @@ var processGRNmap = function (path, res, app) {
     helpers.attachFileHeaders(res, path);
     network = parseSheet(sheet);
 
+    // Parse expression and 2-column data, then add to network object
+    var additionalData = parseAdditionalSheets(sheet);
+    for (var key in additionalData) {
+        network[key] = additionalData[key];
+    }
```
dondi: jQuery has an `extend` helper function which can replace this block with a one-liner, I think. Look it up here: https://api.jquery.com/jquery.extend/

test/additional-sheet-parser-tests.js
```diff
@@ -0,0 +1,91 @@
...
+        var workbook = xlsx.parse(__dirname + "/../test-files/spreadsheet-controller-test-files/" +
+        "2_column_data_format_test.xlsx");
+        var data = parseAdditionalSheets(workbook);
+        assert(data["test"]["degradation_rates"], {
+            "ACE2": 0.1118,
+            "ASH1": 0.2166,
+            "CIN5": 0.1005,
+            "GCR2": 0.0963
+        });
+    });
```
dondi: Cool, these tests capture the full black-box functionality of the additional sheet parser. With function decomposition, you can potentially narrow the granularity to just specific cases. Those tests would be truly more like _unit_ tests than these. There's value in both kinds, I think.

web-client/public/js/graph.js
```diff
@@ -370,18 +370,18 @@ var drawGraph = function (nodes, links, positiveWeights, negativeWeights, sheetT
     var weight = boundingBoxContainer.selectAll(".weight");
 
     simulation
-        .nodes(nodes)
+        .nodes(network.genes)
```
dondi: Go over this naming with @kdahlquist —I think we intentionally chose the name “nodes” over “genes” initially because there’s nothing specific about our adjacency matrices that says that the row/column headers represent genes at all. To preserve that generality, it may be better to stick with `network.nodes`. But then, perhaps we are now moving into territory where the generalization is getting forced...?

Either way, this is worth a little discussion I think.

server/controllers/spreadsheet-controller.js
eileenchoe: I don't think we can use jQuery on the server side, but I found `Object.assign()` which does the same thing!

server/controllers/additional-sheet-parser.js
eileenchoe: Sounds good

web-client/public/js/graph.js
eileenchoe: The network object that the spreadsheet controller creates has always contained the property `genes`. We've just been "renaming" it to nodes in upload.js before passing it in to the `drawGraph` function. @kdahlquist @dondi should I refactor it to `nodes` in the network object itself?

dondi: Ah OK I get it. True that an incoming GRNmap workbook _is_ gene-specific, so that is understandable. It is _GRNsight_ that is a little more generalized, and that explains the GRNsight code’s use of `nodes` once the import is finished.

This works for me then.

server/controllers/additional-sheet-parser.js
```diff
@@ -0,0 +1,91 @@
...
+module.exports = function (workbook) {
+
+    var output = {
+        expression: {}, // expression data
+        meta: {},
+        test: {} // 2-column data
+    };
+
+    for (var i = 0; i < workbook.length; i++) {
+        var sheet = workbook[i];
```
dondi: This is a classic pattern that, in more modern languages, is done as an iterator function rather than a `for` loop: index variable (`i` here) is only used in the beginning, and the value it represents is assigned right away (`sheet`), with that variable being the only one used for the rest of the loop.

When you spot that, it’s more compact (and some might say direct) to write:
```javascript
workbook.forEach(function (sheet) {
    // .....etc.....
})
```

server/controllers/spreadsheet-controller.js
```diff
@@ -432,6 +433,10 @@ var processGRNmap = function (path, res, app) {
     helpers.attachFileHeaders(res, path);
     network = parseSheet(sheet);
 
+    // Parse expression and 2-column data, then add to network object
+    var additionalData = parseAdditionalSheets(sheet);
+    Object.assign(network, additionalData);
```
dondi: Nice find!

server/controllers/spreadsheet-controller.js
```diff
@@ -432,6 +433,10 @@ var processGRNmap = function (path, res, app) {
...
+
     return (network.errors.length === 0) ?
     // If all looks well, return the network with an all clear
```
dondi: Didn’t see this previously: note lines 441-444 are part of the conditional expression that starts in line 440, so should actually be indented.

eileenchoe:
Thanks for the feedback @dondi. Since I've incorporated the changes, and CI looks good I'm going to go ahead and merge this PR now!

Issue #518, opened 2017-06-01
Change gray edge threshold value to percent scale (0 - 100%)


kdahlquist:
Leftover issue from #264.  Gray threshold value should be shown as percent, 0 - 100%, instead of fraction from 0 - 1.  @mihirsamdarshi noted that this wasn't a straightforward change and left it for later work.

mihirsamdarshi:
This is taking longer than I expected, but I hope to push something to fix this problem by this Wednesday's work session.

mihirsamdarshi:
I can't seem to figure this one out, because of the way that I designed the variable gray threshold. I think I have found a way to fix it, by appending a string to the value, however I am not sure how to incorporate it yet. I will try to fix this later this week.

kdahlquist:
@mihirsamdarshi is looking at the code for zoom on how to do this.

mihirsamdarshi:
I've figured out how to actually do this, now I simply have to implement my solution

dondi:
@mihirsamdarshi I saw the PR; I submitted a review. One small note and you should be good to go.

kdahlquist:
@yshin4 merged the PR.  However, she does not have rights/does not know how to make the changes live on the server.  Who has rights to do that these days?

mihirsamdarshi:
@yshin4 check issue #388 to generate an SSH key, and send that to Dondi to gain access to the server

yshin4:
@mihirsamdarshi oh, I do have the ssh key from last year. How do I make changes live on the server?

mihirsamdarshi:
https://github.com/dondi/GRNsight/wiki/Quick-Guide-to-Refreshing-GRNsight

yshin4:
@mihirsamdarshi nice, thanks! Followed the guide but I did not know why we would have to do multiple screens, so I just did it once. Tell me if I am wrong. But other than that I pulled on the server.

mihirsamdarshi:
@yshin4 you do have to screen multiple times. currently beta is down.

yshin4:
@mihirsamdarshi Yeah, sorry I didn't know how many times to do it. I reported to dondi an hour ago and he said he'll give it a look when he's in front of a computer. I tried to reset the pull and changed the head to a previous commit and pull again, but screens were already all closed. Let me know if you know a fix.

dondi:
[...]

PR #569, opened 2018-02-26
finished percent label #518

mihirsamdarshi:
this took way longer than it should have in hindsight

### DIFF
--- a/web-client/public/js/sliders.js
+++ b/web-client/public/js/sliders.js
@@ -183,7 +183,7 @@ var sliderGroupController = function (sliderArray) {
 };
 
 // Gray Threshold Slider Settings
-var updateGrayThresholdValue = function () {
+var outputUpdate = function () {
     var value = Math.round(($("#grayThresholdInput").val() * 100));
     $("#grayThresholdValue").html(value + "%");
 };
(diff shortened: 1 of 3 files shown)


mihirsamdarshi:
Pull Request Checklist

web-client/public/js/sliders.js
```diff
@@ -183,9 +183,7 @@ var sliderGroupController = function (sliderArray) {
...
 
 // Gray Threshold Slider Settings
-var graySlider = document.getElementById("grayThresholdInput");
-
-var outputUpdate = function (val) {
-    // val = Math.round(val * 100) + '%';
-    document.querySelector("#grayThresholdValue").value = val;
+var outputUpdate = function () {
+    var value = Math.round(($("#grayThresholdInput").val() * 100));
+    $("#grayThresholdValue").html(value + "%");
```
dondi: You can use `text` here instead of `html`, since the string you are setting doesn't have any HTML markup in it. As a general principle, you want to always use the "minimal" function that does the work that you need. In this case, since you don't need HTML to set the new value, you can opt for `text`:
```javascript
    $("#grayThresholdValue").text(value + "%");
```

mihirsamdarshi: fixed, thanks for the feedback

yshin4:
Since PR checklist is done, and dondi approves, will go ahead and merge.

PR #595, opened 2018-05-14
fixed #438, small express change, small upload.jade change

### LINKED ISSUE #438: Allow numbers as node labels
I tried to make a node label be a pure number, i.e., "111" and it got rejected by GRNsight.  I think we need to allow numbers to be node labels because someone may choose to use numbers instead of letters or letter/number combinations as node labels.

Someone else needs to verify this bug.  When it occurred, it was in the last cell on the right/bottom of my worksheet.

mihirsamdarshi:
Issue #438 was fixed by forcing all gene names to be strings. Additionally space was added before text in line 386 in upload.jade. Also, res.json had been updated, so res.json(status, obj) was updated to res.status(status).json(obj)

### DIFF
--- a/server/controllers/spreadsheet-controller.js
+++ b/server/controllers/spreadsheet-controller.js
@@ -301,7 +301,6 @@ var parseSheet = function (sheet) {
                                 network.genes.push(currentGene);
                             }
                         } catch (err) {
-                            console.log(err);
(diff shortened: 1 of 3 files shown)


mihirsamdarshi:
Pull Request Checklist

PR #614, opened 2018-05-16
starting the single state move

mihirsamdarshi:
This is an initial test, just to see if I'm on the right track. Everything I have copied still exists in the original sliders.js file.


web-client/public/js/grnsight.js
```diff
@@ -5,6 +5,9 @@ const container = require("./container").container;
 const displayStatistics = require("./graph-statistics").displayStatistics; // eslint-disable-line no-unused-vars
 const upload = require("./upload").upload;
 const nodeColoringController = require("./node-coloring").nodeColoringController;
+import setupHandlers from "setuphandlers";
+import grnState from "grnstate";
 
+setupHandlers(grnState);
```
dondi: These changes, as they stand, look good. I agree that this is how GRNsight should start up. You should also add `updateApp` here. The initialization sequence is: set up all event handlers, then make the GRNsight UI display the values of `grnState` for the first time. So:
```javascript
setupHandlers(grnState);
updateApp(grnState);
```

web-client/public/js/grnstate.js
```diff
@@ -0,0 +1,3 @@
+export var grnState = {
+    GRAVITY_LENGTH_WITHOUT_ZERO = 3
+}
```
dondi: The presence of the object itself is in the right direction, but I saw it more as a container for values that will change, and not constants. If you declare `GRAVITY_LENGTH_WITHOUT_ZERO` in this way, it is still possible to change it.

Better to really declare it as a constant (e.g., `const GRAVITY_LENGTH_WITHOUT_ZERO = 3` or as a property that is _read-only_). Instead, try modeling a value here that _will_ change while the app is running.

web-client/public/js/setuphandlers.js
```diff
@@ -0,0 +1,11 @@
+import grnState from "grnstate";
+import updateApp from "updateapp";
+export var setupHandlers = function (grnState) {
+    var updateSliderDisplayedValue = function (slider, element) {
+        var value = $("#" + $(element).attr("id")).val();
+        $(slider.valueId).html(value + ((slider.needsAppendedZeros &&
+          (value.length === grnState.GRAVITY_LENGTH_WITHOUT_ZERO)) ? "0" : ""));
+        slider.setCurrentVal(value);
+    };
```
dondi: This is a helper function that looks more like it belongs to the updateApp function rather than setupHandlers. Notice what it does: it updates what an element on the page is displaying. That's what updateApp is for. setupHandlers is for the _opposite_  direction: it sets up what to do when the user interacts with the page. So setupHandlers should be for stuff like `click`, `change`, and `input`. Anything that goes _toward_ a visible UI element belongs in updateApp.

web-client/public/js/updateapp.js
```diff
@@ -0,0 +1,4 @@
+import grnState from "grnstate";
+export var updateApp = function (grnState) {
+
+}
```
dondi: Note you don't have to import `grnState` here because it is already passed as a parameter.

mihirsamdarshi:
@dondi, I ran into a weird problem while working on this. It seems that the container doesn't load if I even import another file. This doesn't seem to have any correlation with anything linked to container because it occurs even when the imported files are empty

mihirsamdarshi:
There are errors that the modules can't be resolved, but that shouldn't affect the loading of the modules

mihirsamdarshi:
Specifically updateApp causes this problem, not setupHandlers

johnllopez616:
@mihirsamdarshi you have to make some changes to webpack.config.js. Note what I did to module.exports:

`module.exports = {
    entry: "./web-client/public/js/grnsight.js",
    output: {
        path: `${__dirname}/web-client/public/js`,
        filename: "grnsight.min.js"
    },
    module: {
        rules: [
            { test: path.join(__dirname, "es6"),
                loader: "babel-loader" }
        ]
    },
    resolve: {
        modules: ["js", "node_modules"],
    },
};
What this does is it allows webpack to resolve the files within the js folder automatically, if I'm not mistaken.

mihirsamdarshi:
Additionally @dondi, much of the `node-coloring.js` was split into a "controller" function already, is this something that we can just place into the `updateApp.js` file?

dondi:
Yes, I think it will go there eventually.

mihirsamdarshi:
@dondi could you look at my previous comment to see if that architecture was correct?

web-client/public/js/grnsight.js
```diff
@@ -1,3 +1,6 @@
+import setupHandlers from "setuphandlers";
+import updateApp from "updateapp";
+import grnState from "grnstate";
```
dondi: You import these but aren't calling. Make sure that the code all works with this in the code:
```javascript
setupHandlers();
updateApp(grnState);
```

webpack.config.js
```diff
@@ -10,5 +10,8 @@ module.exports = {
             { test: path.join(__dirname, "es6"),
                 loader: "babel-loader" }
         ]
+    },
+    resolve: {
+        modules: ["js", "node_modules"],
```
dondi: It's possible that you won't need this if you import with the "./" preceding the file name. Give it a try.

web-client/public/js/grnstate.js
```diff
@@ -0,0 +1,3 @@
+export var grnState = {
+    dashedLine : $("#dashedGrayLineButton").prop("checked"),
```
dondi: Remember that we said that `grnState` is _pure model_. This should just be `true` or `false`.

web-client/public/js/updateapp.js
```diff
@@ -0,0 +1,5 @@
+/* eslint no-unused-vars: [2, {"varsIgnorePattern": "setupHandlers|grnState"}] */
+var setupHandlers = require("./setupHandlers");
+export var updateApp = function (grnState) {
+    return grnState;
```
dondi: `updateApp` is supposed to set `$("#dashedGrayLineButton").prop("checked")` according to what `grnState.dashedLine` is.

PR #619, opened 2018-05-17
working on single state move, added new files, working out bugs

mihirsamdarshi:


web-client/public/js/graph.js
```diff
@@ -1217,35 +1236,44 @@ export var drawGraph = function (network, sliderController, nodeColoring) {
         return name1 > name2 ? 1 : -1;
     };
 
-    let layout = false;
+    gridLayout = false;
 
     var GRID_LAYOUT_BUTTON = "#gridLayoutButton";
     $(GRID_LAYOUT_BUTTON)[0].value = "Grid Layout";
-    $(GRID_LAYOUT_BUTTON).on("click", {handler: this}, function (event) { // eslint-disable-line no-unused-vars
+    $(GRID_LAYOUT_BUTTON).on("click", {handler: this},
```
mihirsamdarshi: @dondi I was working on moving this force graph layout controller to upload.js, just like the gray edge, dashed line, and physics option controllers are. However, I ran into a problem where the container decides to not load if I change 

`$(GRID_LAYOUT_BUTTON).on("click", {handler: this}, function (event) { ... }` 

to

`$(GRID_LAYOUT_BUTTON).on("click", {handler: this}, gridLayoutToggle)
 var  gridLayoutToggle = function (event) { ... }`

dondi: Timeout timeout—I thought the plan was to move _one_ thing, _very focused_, into the new cycle first? This is a huge set of changes that seems to be disproportionate to moving that single value. I think I need some clarification on how this task, which was supposed to be pretty small and limited, has become this gigantic diff.

mihirsamdarshi: @dondi so this isn't actually moving anything into the new set of files that I have created, this is organizing the existing codebase before tackling each bit step by step. I realized the need for this when I started moving bits and pieces and realized they were scattered all over graph.js

dondi: OK, so this is a distinct task, and not the original MVC refactor. That's fine, but I thought this reorganization would _follow_ the establishment of the clean setupHandlers -> grnState -> updateApp cycle. Don't we want to establish that as a solid foundation first, before then embarking on reorganizing the code?

Reorganizing the code now seems like potential duplicate work, because until you have the clean MVC loop, it won't really be clear where the new code will go.

web-client/public/js/container.js
```diff
@@ -1,5 +1,6 @@
 /* eslint no-unused-vars: [2, {"varsIgnorePattern": "grnTest|enableScroll|manualZoom"}] */
 export const container = function () {
+
```
dondi: Trivial, incidental change; get rid of it.

web-client/public/js/setuphandlers.js
```diff
@@ -0,0 +1,28 @@
...
+        grnState.dashedLine = false;
+    } else {
+        grnState.dashedLine = true;
+    }
+
+    if ($(GREY_EDGES_DASHED_SIDEBAR).prop("checked")) {
+        grnState.dashedLine = true;
+    } else {
+        grnState.dashedLine = false;
+    }
```
dondi: These `if` blocks shouldn't be here. Remember, `grnState` is the _single source of truth_. This code is going backwards: it's setting `grnState` based on the state of the UI. _This should not happen._ Instead, `grnState` _determines_ the state of the UI. That's why we call `updateApp` at the very beginning.

web-client/public/js/setuphandlers.js
```diff
@@ -0,0 +1,28 @@
...
+
+    $(GREY_EDGES_DASHED_MENU).click(function () {
+        updateApp.updateGreyEdgesAsDashedOptions(grnState.dashedLine);
+    });
+
+    $(GREY_EDGES_DASHED_SIDEBAR).on("change", function () {
+        updateApp.updateGreyEdgesAsDashedOptions(grnState.dashedLine);
+    });
```
dondi: This is more of the idea, but still needs some tweaks. _It should update `grnState`_. Remember: MVC, with C looping back to M. So, when the UI elements change, that change updates `grnState` _inside the event handler_.

And remember, `updateApp` is _the_ update function. `updateApp.updateGreyEdgesAsDashedOptions` is too limited. It should just be `updateApp(grnState)`. That way, the code can factor in the _entire_ `grnState` object. You will need that for properties that affect each other.

So, these functions should look more like this:
```javascript
    grnState.dashedLIne = $(GREY_EDGES_DASHED_MENU).prop("checked");
    updateApp(grnState);
```
See how that works? MVC. When the user clicks `$(GREY_EDGES_DASHED_MENU)` (C), the first line updates `grnState`'s corresponding value to match (M). Then `updateApp` cycles back to make sure that the entire app is in sync with that change in `grnState` (V).

If you want to think of this in React terms, these functions are like the `onChange` functions in React. Note that the `onChange` functions (C) then call `setState` (M). When `setState` is called, it triggers a `render` call (V), which returns the new markup for the component. It's the same cycle, conceptually.

If you're concerned that this will lead to a lot of `change` and `click` functions that look the same, then you're right. But that is not the issue at hand. We can consolidate duplicated code later. For now, we want to establish the right structure. Again, this is _no different_ from the React cycle of having a state (`grnState`), having `onXXXXX` handlers (`click`, `change`, and friends), and having those handlers call `setState` (updating `grnState`). If there is anything that React does somewhat invisibly, it is that `setState` automatically calls `render`. In our case, all of our handlers will call `updateApp` explicitly.

web-client/public/js/updateapp.js
```diff
@@ -0,0 +1,22 @@
...
+            $(GREY_EDGES_DASHED_MENU + " span").addClass("glyphicon-ok");
+            $(GREY_EDGES_DASHED_MENU).prop("checked", "checked");
+            $(GREY_EDGES_DASHED_SIDEBAR).prop("checked", "checked");
+        } else {
+            $(GREY_EDGES_DASHED_MENU + " span").removeClass("glyphicon-ok");
+            $(GREY_EDGES_DASHED_MENU).removeProp("checked");
+            $(GREY_EDGES_DASHED_SIDEBAR).removeProp("checked");
+        }
+        upload.drawGraph(currentNetwork, sliders, nodeColoring);
+    };
```
dondi: See the comment above. The function should be "opened" so that all of this update code is at the top level of `updateApp`.

Also, with regard to your technical question, yes, the `drawGraph` function will eventually grab all of its values from `grnState`. In fact, it will probably morph into `drawGraph(grnState)`. Depending on how quickly the transition to this cycle happens, the shift to passing `grnState` may happen sooner rather than later.

web-client/public/js/setuphandlers.js
mihirsamdarshi: I meant to make these 3 equal signs, would that be more correct?

mihirsamdarshi:
@dondi I'm having a little bit of trouble figuring out why in grnsight.js, neither `setupHandlers(grnState)` nor the `updateApp(grnState)` are being called.

mihirsamdarshi:
I made the decision to move all of graph.js to updateapp.js, in order to facilitate a smooth transition. however, @dondi, an error I am getting now has to do with web pack not being able to import the grid stuff, how would I solve this

mihirsamdarshi:
Never mind, I solved this, however displayNetwork can't seem to access drawGraph from upload.js, despite me changing the only reference to drawGraph. This seems to be because drawGraph is an export var, but it cannot be an export var because it is nested within updateApp. how would I make drawGraph to be able to be called from upload.js?

dondi:
@mihirsamdarshi I need to emphasize this: ***PLEASE STOP UNILATERALLY DECIDING TO LAUNCH INTO VERY LARGE DELTAS OF WORK***. We have stated multiple times now that ***YOU MUST LIMIT THE SIZE OF YOUR COMMITS***. Continuing to do this without consultation first will just use up more time than necessary. If you felt that this leap was necessary, please get into the habit of consulting the team first.

Notice the pattern you have fallen into: you try something, then hit an issue. Then you solve it but hit another issue. I won’t be shocked if resolving this latest issue will lead you into another one. By the time you finish (_if_ you finish), you will have made _huge_ changes to the code. That is _not_ the right way to work.

A company with whom I consult once had a newly-hired programmer do this, and that employee _was fired after two weeks on the job_. Seriously, it is a habit that must be avoided.

You were not on a holding pattern here; there were standing instructions for you to continue to read up on React or work on other tasks while waiting for me to be able to respond. It would not have been a waste of your time to pause here and do some forward learning in the meantime.

I will look into this issue and provide a response sometime tonight, and we can discuss this overall scenario at Tuesday’s meeting. That is what should have been done: Notify me that you got stuck with triggering a graph redraw, and _stop at that point_ so that I could see the code and think about the issue. You may have fixed the graph redraw but you then ran into a webpack problem. Then you might have fixed the webpack problem, but now you have run into a scoping issue, and this will lead to even more changes given the course that you are already on. It is an escalation pattern that almost always will lead to excess lost time.

The very prospect of a large number of changes must become an _immediate red flag_. It is a definite signal that you should ask for help first before going forward.

mihirsamdarshi:
@dondi I want to clarify, I meant to close this PR. This is all experimentation work that was meant to determine the scope of the effort to move our codebase to an MVC model. I believed it was appropriate for me to experiment with large moves on my own branch as it would not have affected our current beta. I wanted to see what kind of challenges I would be dealing with, which is why I asked earlier in the PR whether drawGraph would be moved to updateApp, and decided to move the function to updateApp to see what kind of rabbit hole I would be diving into. This was something that I wanted to explore during our in person meeting tomorrow

dondi:
@mihirsamdarshi OK, if this was exploratory work then that’s fine, but the practice is to not issue a pull request and to ask questions directly about a branch. Pull requests imply that something is meant to be merged eventually.

But that being said, it should not have even been an option to explore a change this big. The moment you realized it was a rabbit hole is the moment you drop it and consult. The PR was on the order of 1500 changed lines—already way too big for a unit of work.

Issue #621, opened 2018-05-18
Gene page API issues, documentation, and testing


kdahlquist:
Currently on beta v3.0.5, the gene page is opening, but data is not being loaded.

Copied over a post by @johnllopez616 from #573:

_I discovered that at least one of the data loading issues is related to the first part of ` getUniProtInfo()` within `api.js`.  What this function does is make a request to UniProt Retrieve/ID Mapping to retrieve the UniProt entry ID as part of the UniProt API request. I've noticed that the current function to make this request isn't reliable. I will investigate further later today._

Once the basic functionality is up for @kdahlquist to look at on beta, we also want to make sure that if a user has a gene that is not found in one or more databases, the appropriate "not found" messages come up.  There are two cases to consider:
- if the gene is just not found at all in any database
- if the gene is found in one or more, but not all databases

When the Fall 2017 BioDB wiki is back up, @kdahlquist will review the documentation there to generate a guide to the expected results.  If it's not up by Monday the 21st, she will re-generate this documentation so it is not a blocker to progress on this issue.

kdahlquist:
Problem: identical requests to UniProt will sometimes work and sometimes will not. A request for information about a particular gene will either return the proper information or will return an error.  Note that we don't know why this is happening.

Solution: have it make identical requests until one of the requests is successful and returns proper data. 

Next problem: None of the gene information will display on gene page unless all API calls are successful.

Proposed solution: These need to be independent of each other.  Not found or other such message should display on gene page for information not found from a particular database.

kdahlquist:
I just looked at beta v3.0.6 and there still isn't any information populated on the gene page.

johnllopez616:
It would appear that the api.js file returns a 404 error on beta when uploaded to the server.

dondi:
Yes, try using a fully-relative URL. (no leading slash)

kdahlquist:
@johnllopez616 and I discussed that an error modal will only come up if there gene information is not found at all in any database.

I'm noting here in the demo that @johnllopez616 showed me that the biodb class put "N/A" for information I requested that they were not able to implement during the semester.  @kdahlquist will need to review each of these items and determine whether we should give up on it or try to implement it.

johnllopez616:
[...]

Issue #620, opened 2018-05-18
Tweaks to look and feel of gene page


kdahlquist:
Moving relevant items from #573 to this issue which is all about tweaks to the look and feel of the gene page.

- [x] the gene page also needs the GRNsight footer (but not the page view counter); confirmed finished in beta v3.0.6 by @kdahlquist
- [x] title text for the new tab that opens should be "GRNsight - name of gene" instead of "Information about name of gene"  For example, instead of "Information About STB5" it should say "GRNsight - STB5"; confirmed finished in beta v3.0.6 by @kdahlquist
- [x] @kdahlquist will provide feedback on what colors to use for the buttons/section headers (see below); confirmed finished in beta v3.0.7 by @kdahlquist
- [x] Can the buttons be left-justified instead of centered above the accordian sections; confirmed finished in beta v3.0.7 by @kdahlquist
- [ ] Will you reduce the white space on the top and left?  See the News or About page for how much margin there should be.  This could still be reduced more.  The margin on the News and About page is something like 0.5 cm; left margin looks good in v3.0.10, but the top margin needs to be shortened still.  There appears to have been a regression in v3.0.13: the left margin is too big again.  The top margin could be shortened a little; see the News or About page for the appropriate height.  _moved this to new issue_
- [x] Font for the species name is too large, the species name is taking up too much room at the top of the page.  It makes sense to have the name of the gene be large, but not the species name.  Confirmed by @kdahlquist in beta v3.0.10 (looks nice).
- [x] There needs to be some white margin on the left in the different sections when they are populated with data.  Right now, the text appears flush with the edge.  Confirmed complete in beta 3.0.9 by @kdahlquist.
- [x] In the Sources section, please number the sources like this instead of using the superscripts and left-justify instead of center in the section: (@kdahlquist says almost complete in v3.0.13, but see note in bold below about NCBI *Gene* Database.)

1. Saccharomyces Genome Database
2. UniProt
3. NCBI Gene Database **(note the inclusion of the word "Gene" there)**
4. Jaspar Database
- [x] In the Sources section, "Saccharomyces" is spelled incorrectly. (@kdahlquist can't review yet because data not loaded due to Jaspar bug.)
- [x] Delete "Gene Map" button and section; there doesn't appear to be any data that goes there.  Confirmed complete in beta 3.0.9 by @kdahlquist.
[...]

PR #626, opened 2018-05-18
Gene Page Changes Made 5/18/18 #2

johnllopez616:
Addresses parts of issues #620 and #621:

By moving `api.js` into _web-client/public/gene/_ (and changing the directory of the script request within info.html), the js folders for the gene page are consolidated into, and when they are uploaded to a server, the browser _can actually locate them_. This should hopefully resolve our issues where not even the error messages would load in beta, because the browser simply couldn't access _api.js_ to begin with.

I addressed the stylistic changes in #620 by making sure each link in the menu and footer of the Gene Page takes the user to a separate tab. Additionally, I slightly moved the container with the gene page information to the left.


web-client/public/gene/info.html
```diff
@@ -312,7 +312,7 @@ <h5 class="mb-0">
       <script src="https://code.jquery.com/jquery-3.2.1.min.js" integrity="sha256-hwg4gsxgFZhOsEEamdOYGBf13FyQuiTwlAQgxVSNgt4=" crossorigin="anonymous"></script>
       <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.11.0/umd/popper.min.js" integrity="sha384-b/U6ypiBEHpOf/4+1nzFpr53nxSS+GLCkfwBdFNTxtclqqenISfwAzpKaMNFNmj4" crossorigin="anonymous"></script>
       <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/js/bootstrap.min.js" integrity="sha384-h0AbiXch4ZDo7tp9hKZ4TsHbi047NrKGLO3SEJAg45jXxnGIfYzk4Si90RDIqNm1" crossorigin="anonymous"></script>
-      <script src="/js/api.js"></script>
+      <script src="api.js"></script>
       <script src='info.js'></script>
```
dondi: I know this was pre-existing code, but in the spirit of steady improvement, let's stay consistent with using double-quotes for HTML attribute values.

web-client/public/gene/info.js
```diff
@@ -19,6 +19,7 @@
     // https://stackoverflow.com/questions/8648892/convert-url-parameters-to-a-javascript-object
 
     var api = window.api;
+    console.log(obj.symbol + typeof obj.symbol);
```
dondi: Good to lose the earlier `console.log`, but it seems to be compensated for by this one. Was this an oversight? In general we try to restrict console logs to when we debug, removing them before commits (with the one exception being the console.log where we display the resulting network after loading a file—but that's more like the exception that proves the rule).

PR #628, opened 2018-05-21
Gene Page Changes Made 5/21/2018

johnllopez616:
Addresses parts of #620 and #621 

After realizing that 4/5 API calls were successful on the` beta` branch, I noticed that JASPAR would return a 404 Error on `beta` while on `johnlopez` it would return the data. This was due to an outdated API call that could function on` localhost` but not on the GRNsight server. I updated the API call to the current documentation's standards, but I would receive an error stating that my` localhost` did not have access to the database. This "No 'Access-Control-Allow-Origin' header is present on the requested resource" error would occur until I installed the following Chrome extension: https://chrome.google.com/webstore/detail/allow-control-allow-origi/nlfbmbojpeacfghkpbjhddihlkkiljbi?hl=en-US.

I am unsure if this error will appear when the API call is made from `beta`, so there is a chance I will have to rework the JASPAR API link with another solution.

Additionally, I made the last 5 stylistic changes on the checklist of #620.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "grnsight",
-  "version": "3.0.3",
+  "version": "3.0.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


johnllopez616:
Will investigate why the travis build failed after lunch.

PR #630, opened 2018-05-22
Beta consolidation first-full-cycle example

dondi:
OK @mihirsamdarshi and @johnllopez616, try taking this for a spin. I based this on the `dashedLine` work that @mihirsamdarshi had already started. All tests pass, and the functionality seems to be identical to the current posted beta.

Hope this helps clarify the MVC cycle discussed last Tuesday, and also shows how state can be shared across modules (temporarily, we hope) while things are in transition. Presumably, those values eventually find their way either in `grnState`, `updateApp`, or `setupHandlers`.

Please take a look at this before the meeting so that we can discuss any questions that you might have about this code.

### DIFF
--- a/web-client/public/js/constants.js
+++ b/web-client/public/js/constants.js
@@ -1,0 +1,2 @@
+export const GREY_EDGES_DASHED_MENU = "#grey-edges-dashed-menu";
+export const GREY_EDGES_DASHED_SIDEBAR = "#dashedGrayLineButton";
--- a/web-client/public/js/grnsight.js
+++ b/web-client/public/js/grnsight.js
@@ -1,10 +1,17 @@
-const drawGraph = require("./graph").drawGraph;
-const sliderObject = require("./sliders").sliderObject;
-const sliderGroupController = require("./sliders").sliderGroupController;
-const container = require("./container").container;
-const displayStatistics = require("./graph-statistics").displayStatistics; // eslint-disable-line no-unused-vars
-const upload = require("./upload").upload;
-const nodeColoringController = require("./node-coloring").nodeColoringController;
+import { drawGraph } from "./graph";
+import { sliderObject } from "./sliders";
+import { sliderGroupController } from "./sliders";
+import { container } from "./container";
+import { displayStatistics } from "./graph-statistics"; // eslint-disable-line no-unused-vars
+import { upload } from "./upload";
+import { nodeColoringController } from "./node-coloring";
+
+import { grnState } from "./grnstate";
+import { updateApp } from "./update-app";
+import { setupHandlers } from "./setup-handlers";
+
+setupHandlers(grnState);
+updateApp(grnState);
 
 container();
 upload(sliderObject, sliderGroupController, drawGraph, nodeColoringController);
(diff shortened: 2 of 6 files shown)


web-client/public/js/grnsight.js
```diff
@@ -1,10 +1,17 @@
...
-const displayStatistics = require("./graph-statistics").displayStatistics; // eslint-disable-line no-unused-vars
-const upload = require("./upload").upload;
-const nodeColoringController = require("./node-coloring").nodeColoringController;
+import { drawGraph } from "./graph";
+import { sliderObject } from "./sliders";
+import { sliderGroupController } from "./sliders";
+import { container } from "./container";
+import { displayStatistics } from "./graph-statistics"; // eslint-disable-line no-unused-vars
+import { upload } from "./upload";
+import { nodeColoringController } from "./node-coloring";
```
dondi: The transition from `require` to `import` wasn't necessary but I was already modifying this file so I figured it wouldn't be a bad idea to show how to make this change. You'll notice that `import` is less verbose and generally easier to read.

web-client/public/js/constants.js
```diff
@@ -0,0 +1,2 @@
+export const GREY_EDGES_DASHED_MENU = "#grey-edges-dashed-menu";
+export const GREY_EDGES_DASHED_SIDEBAR = "#dashedGrayLineButton";
```
dondi: Consolidation into a `constants` module was not initially called for but the need for it was noticed right away based on the way these values were used. Another option instead of a separate `constants` file is to also fold these into `grnState`. Either way could work.

web-client/public/js/upload.js
```diff
@@ -1,3 +1,12 @@
+// TODO Likely a temporary location, while things are being moved to their "true" homes.
+//      But placed here for now so that the true MVC cycle of grnState, updateApp, and the
+//      controller code installed by setupHandlers can access them.
+export const uploadState = {
+    currentNetwork: null,
+    sliders: null,
+    nodeColoring: null
+};
```
dondi: This is the workaround for dealing with dependencies on variables that haven't been cleaned up yet. It does the job without requiring large changes to the code. Notice the additional lines below where we relay the values to this "holding object" as they are assigned or initialized.

dondi: As the transition proceeds, this object will probably grow a bit more. But presumably as things get more consolidated, it will then start contracting again as its code/values find their rightful place either in `grnState`, `updateApp`, or `setupHandlers`. Eventually it should disappear entirely.

dondi:
Not covered here is how to write a unit test for this cycle because it requires the setup of a mock document with just the UI elements involved (`GREY_EDGES_DASHED_MENU` and `GREY_EDGES_DASHED_SIDEBAR`), then simulating clicks on those to check whether `grnState` is updated correctly. Mocks for `drawGraph` would also be needed. So still a bit of work in that area before code of this type can be tested properly. But hope this PR in general helps convey the idea better.

mihirsamdarshi:
Pull Request Checklist
- [ ] Print works as expected

PR #631, opened 2018-05-22
single state move

mihirsamdarshi:


web-client/public/js/update-app.js
```diff
@@ -15,6 +17,26 @@ const refreshApp = () => {
...
+// Edge Weight Normalization
+    let valueValidator = (min, max, value) => {
+        return Math.min(max, Math.max(min, value));
+    };
+
+    let edgeWeightNormalizationInputValidation = (value) => {
+        return value === "" ? "" : valueValidator(MIN_EDGE_WEIGHT_NORMALIZATION, MAX_EDGE_WEIGHT_NORMALIZATION, value);
+    };
+
+    var validated = edgeWeightNormalizationInputValidation(value);
```
mihirsamdarshi: @dondi I'm having a little bit of trouble figuring out why the code is breaking at this point on line 29, it seems that validated isn't a valid variable or something of that sort

web-client/public/js/graph.js
```diff
@@ -1,4 +1,5 @@
 import Grid from "d3-v4-grid";
+import { grnState } from "./grnstate";
```
mihirsamdarshi: @dondi additionally, this doesn't seem to allow me to call grnState.normalizationMax from grnstate

mihirsamdarshi:
@dondi the normalization features have been moved, I will now move on to gray edge threshold. I will merge after you approve it

dondi:
@mihirsamdarshi OK thanks, just acknowledging that I got this update. Stand by on this until I give an explicit review. Thanks!

mihirsamdarshi:
ok, the I finished movement of gray edge and upload as well, and will stand by until further

web-client/public/js/graph.js
```diff
@@ -91,6 +92,8 @@ export var drawGraph = function (network, sliderController, nodeColoring) {
         document.getElementById("edge-weight-normalization-factor-menu").setAttribute("placeholder", "");
     } else {
         var maxWeight = d3.max(allWeights);
+        grnState.resetNormalizationMax = maxWeight;
+        grnState.normalizationMax = maxWeight;
```
dondi: Note that it was this assignment that triggered the shift of doing the entire `network` and loading/importing sequence first. In order to follow our new structure properly, _there should be no changes to the model outside of controller code_. When I first saw this on Tuesday, recall that we tracked down the origin of `maxWeight` to a calculation based purely on the network's weights (line 60/61; open up the diff to see it or look at your code directly).

This was the plan that we mapped out on Tuesday:
- [x] `grnState` now holds the current `network` (This has been done)
- [x] All loading and importing code now follows our new MVC flow: the triggers live in `setupHandlers`, and this code assigns the new `network` to `grnState`. Upon assigning `network`, `updateApp` refreshes the view (Also done)
- [x] The setting of the `network` is expected to trigger a cascade of related model calculations—_this is one of them_. To help you with implementation, the `network` property was implemented as a getter/setter, so that additional code can be executed when a new `network` is assigned (Also done)
- [ ] _Now this is what you should do:_ Expand the `network` setter code so that, when the network is assigned, _the calculation that leads to `maxWeight` is performed in `grnState`_. This will lead to the assignment of `resetNormalizationMax` and `normalizationMax`. `updateApp` then proceeds as it currently does, and if all goes well, the view is updated accordingly.

Remember, we are all about `grnState` -> `updateApp` -> controller code in `setupHandlers` -> `grnState`. _Do not digress from this flow unless there is a really good reason to do so_. (we have one so far—I marked it with a comment explaining the exception)

I hope this spells it out sufficiently for you. Let me know if you need further clarification.

web-client/public/js/grnsight.js
```diff
@@ -10,8 +10,8 @@ import { grnState } from "./grnstate";
...
 import { setupHandlers } from "./setup-handlers";
 
-setupHandlers(grnState);
-updateApp(grnState);
-
 container();
 upload(sliderObject, sliderGroupController, drawGraph, nodeColoringController);
+
+setupHandlers(grnState);
+updateApp(grnState);
```
dondi: I suspect that if you do the MVC refactor correctly, then you won't have to make this change in order.

web-client/public/js/grnstate.js
```diff
@@ -42,5 +42,8 @@ export const grnState = {
 
     dashedLine: false,
 
-    annotateLinks: () => annotateLinks(currentNetwork)
+    annotateLinks: () => annotateLinks(currentNetwork),
+
+    normalizationMax : null,
+    resetNormalizationMax: null
```
dondi: This works, although is there a more appropriate initial value than `null`?

web-client/public/js/setup-handlers.js
```diff
@@ -19,4 +19,19 @@ export const setupHandlers = grnState => {
...
+
+    $("#reset-normalization-factor-menu, #resetNormalizationButton").click(function () {
+        grnState.normalizationMax = grnState.resetNormalizationMax;
+        updateApp(grnState);
+    });
+
+    $("#edge-weight-normalization-factor-menu").on("change", function () {
+        grnState.normalizationMax = $("#edge-weight-normalization-factor-menu").val();
+        updateApp(grnState);
+    });
```
dondi: Good, this is the correct C -> M direction.

If a block of handlers looks related and appears to be lengthy, feel free to consolidate that into a single function (e.g., `setupNormalizationFactorHandlers`). It doesn't need to be in a different module the way the loaders/importers got moved. I leave the judgment call of when to do this up to you. Let's see if we agree on when to cross this threshold 🙂

web-client/public/js/update-app.js
```diff
@@ -48,6 +51,28 @@ export const updateApp = grnState => {
...
+        var edgeWeightNormalizationInputValidation = (value) => {
+            return value ===
+            "" ? "" : valueValidator(MIN_EDGE_WEIGHT_NORMALIZATION, MAX_EDGE_WEIGHT_NORMALIZATION, value);
+        };
+
+        var synchronizeNormalizationValues = (value) => {
+            var validated = edgeWeightNormalizationInputValidation(value);
+            $("#normalization-max").val(validated);
+            $("#edge-weight-normalization-factor-menu").val(validated);
+        };
```
dondi: Is there a good reason for embedding `valueValidator`, `edgeWeightNormalizationInputValidation`, and `synchronizeNormalizationValues` inside the `if` statement? I don't see any dependencies on the function's scope. At this location, these functions get defined over and over again whenever `updateApp` is called. If there is a scope dependency, then that is tolerable, but I don't see that here. Move these outside, the way `displayNetwork` is outside. That way they only get defined once.

web-client/public/js/update-app.js
```diff
@@ -48,6 +51,28 @@ export const updateApp = grnState => {
...
+
+        synchronizeNormalizationValues(grnState.normalizationMax);
+        refreshApp();
```
dondi: When those functions are moved out, then the part inside `updateApp` gets reduced to these two lines. Much more readable, yes?

web-client/public/js/update-app.js
```diff
@@ -48,6 +51,28 @@ export const updateApp = grnState => {
...
+    }
+
+// Dashed Line Synchronization
```
dondi: Line up the comment with the `if`.

web-client/public/js/upload.js
```diff
@@ -135,43 +135,17 @@ export const upload = function (sliderObject, sliderGroupController, drawGraph,
...
-    });
-
-    $("#reset-normalization-factor-menu, #resetNormalizationButton").click(function () {
-        synchronizeNormalizationValues("");
-    });
-
-    $("#edge-weight-normalization-factor-menu").on("change", function () {
-        synchronizeNormalizationValues($("#edge-weight-normalization-factor-menu").val());
-    });
-
```
dondi: Yes, more code that departs _upload.js_!

web-client/public/js/upload.js
```diff
@@ -135,43 +135,17 @@ export const upload = function (sliderObject, sliderGroupController, drawGraph,
...
     var GREY_EDGE_THRESHOLD_MENU = "#gray-edge-threshold-menu";
     var GREY_EDGE_THRESHOLD_SLIDER_SIDEBAR = "#grayThresholdInput";
     var GREY_EDGE_THRESHOLD_TEXT_SIDEBAR = "#grayThresholdValue";
 
     // Gray Edge Controller
 
+    let valueValidator = (min, max, value) => {
+        return Math.min(max, Math.max(min, value));
+    };
+
```
dondi: No, there shouldn't be any additions to _upload.js_ unless they are transitional. I think you forgot to take this out.

web-client/public/js/grnstate.js
mihirsamdarshi: as of right now, there is none because a preset normalization value does not exist until the network is actually loaded

web-client/public/js/upload.js
mihirsamdarshi: This was needed for gray edge threshold, which is still included in upload.js, when I took it out gray edge broke, so I had to add it

mihirsamdarshi:
All changes proposed by @dondi have been done, I do not know if I should merge or wait

web-client/public/js/grnstate.js
```diff
@@ -31,16 +32,20 @@ export const grnState = {
 
     newNetwork: false,
 
+    normalizationMax : null,
```
dondi: The linter didn't catch this? Typically we don't have a space before colons in property-value lists.

web-client/public/js/upload.js
dondi: OK. So it can go away when gray edge threshold gets moved. Mark this with a TODO comment then, so we don't forget about it later.

web-client/public/js/grnstate.js
```diff
@@ -31,16 +32,20 @@ export const grnState = {
...
+    resetNormalizationMax: null,
+
     get network () {
         return currentNetwork;
     },
 
     set network (network) {
         currentNetwork = network;
+        this.resetNormalizationMax = max(network.positiveWeights.concat(network.negativeWeights));
```
dondi: Whoa, after all that code in _graph.js_, this is what it boils down to? Shouldn't https://github.com/dondi/GRNsight/blob/beta/web-client/public/js/graph.js#L62-L72 be done to `allWeights` first?

(and yes, if you're thinking "but that needs `colorOptimal`" —then you have your next candidate to move into `grnState` 🙂)

web-client/public/js/grnsight.js
dondi: Don't forget to undo this change if it isn't needed. Part of the whole "make your commits as small as possible" theme.

web-client/public/js/grnstate.js
dondi: It would be understandable if we pushed `colorOptimal` to another phase, but in that case, leave a TODO comment here. You'd be amazed at how many times a developer might say "I'll get to that later" but they forget because other stuff comes up. Putting a comment here will help guide your future self.

test/grnstate-tests.js
```diff
@@ -0,0 +1,14 @@
...
+// import { setupHandlers } from "./setup-handlers";
+
+/* TODO: Framework for front-end test
+
+var testResetNormalization = (input, propertyExpected) => {
+    var sheet = xlsx.parse(input);
+    var network = spreadsheetController.parseSheet(sheet);
+
+    assert.equal(propertyExpected, grnState.propertyValue);
+}
```
dondi: Yes, this skeleton is in the right direction. Some notes:
- I don't think you'll need `updateApp` nor `setupHandlers`. This is strictly a unit test suite for `grnState`.
- You don't need to parse any sheets. Go straight to building a network _in memory_. We aren't testing the reading of sheets here (we already have lots of tests for that)—we are testing to make sure that `grnState` updates itself correctly.

Thus, your test code will look something like this:
```javascript
describe("The grnState model object", () => {
    it("updates the reset normalization maximum correctly", () => {
        grnState.network = {
            // Build a network here. Because reset normalization max only relies on positiveWeights and
            // negativeWeights, your test network may contain just those properties.
        };

        grnState.resetNormalizationMax.should.equal(/* whatever it should be */);
    });

    // Provide additional tests as needed.
});
```

web-client/public/js/graph.js
```diff
@@ -57,9 +52,12 @@ export var drawGraph = function (network, sliderController, nodeColoring) {
 
     var minimumScale = MIN_SCALE;
 
+    /* If grnState.colorOptimal is false, then weighting is ignored,
+    and the lines are all drawn as if it was an unweighted sheet */
+    // TODO: incorporate into grnState
```
dondi: OK, this comment works.

web-client/public/js/graph.js
```diff
@@ -57,9 +52,12 @@ export var drawGraph = function (network, sliderController, nodeColoring) {
...
     var allWeights = network.positiveWeights.concat(network.negativeWeights);
 
-    if (!colorOptimal) {
+    if (!grnState.colorOptimal) {
```
dondi: _This_ change I was envisioning for the _next_ iteration, not this one. Remember, _we are limiting the size of our units of work_. Looks like we have some further calibration to do in terms of determining the right boundaries for that. FWIW I viewed the `TODO` comment as something for _this_ PR so that we don't forget; but the actual work of _moving_ `colorOptimal` would be a future round.

dondi: Remember, we don't want to "pollute" our old code with new code; we want, as much as possible, to just be moving old code _out_ of there and into the new framework. So, rather than introducing `grnState` here, it's cleaner to move `colorOptimal` _out_ completely in a future round of work. (future as in, within the next few PRs)

web-client/public/js/setup-handlers.js
```diff
@@ -34,4 +34,9 @@ export const setupHandlers = grnState => {
         grnState.normalizationMax = $("#edge-weight-normalization-factor-menu").val();
         updateApp(grnState);
     });
+
+    if (!$("#colorEdges").hasClass("active")) {
+        grnState.colorOptimal = false;
+        updateApp(grnState);
+    }
```
dondi: This doesn't look like a handler. `$("#colorEdges").hasClass("active")` is _not_ an event like `click` or `change`. In that light, this is a backwards move: it changes `grnState` based on the state of the app. Remember, we want the opposite: `grnState` _dictates_ the state of the app.

`setupHandlers` changes `grnState` _when the user performs an action_. There is no action here.

web-client/public/js/grnsight.js
```diff
@@ -12,6 +12,5 @@ import { setupHandlers } from "./setup-handlers";
 
 setupHandlers(grnState);
 updateApp(grnState);
-
```
dondi: _Almost_ got this restored to the original 🙂 You missed a deleted blank line that shouldn't have been deleted. Remember, when undoing changes, the undo isn't complete until _the file itself disappears from the PR's list of changed files_.

test/grnstate-tests.js
mihirsamdarshi: I see. So @dondi, does this mean I don't have to go around creating tests such as the noErrors() function in tests.js (which was my original goal with resetNormalization function), and rather should should just stick to the describe, it, etc. framework?

mihirsamdarshi:
@dondi all changes have been made, its ready to merge

test/grnstate-tests.js
dondi: Correct, `noErrors` is the exception, not the rule. This becomes apparent when you consider the answer to the question: what exactly is `noErrors` testing? The answer that question shows you that the `noErrors` design was meant to address a very specific testing need, and that need does not apply here.

web-client/public/js/graph.js
```diff
@@ -43,7 +43,8 @@ export var drawGraph = function (network, sliderController, nodeColoring) {
         return node.name.length * 12 + 5;
     };
 
-  // If colorOptimal is false, then weighting is ignored, and the lines are all drawn as if it was an unweighted sheet
+    /* If colorOptimal is false, then weighting is ignored,
+    and the lines are all drawn as if it was an unweighted sheet */
     if (!$("#colorEdges").hasClass("active")) {
         colorOptimal = false;
     }
```
dondi: Great; now that this is restored, looking forward to the next round you can think about what exactly is the right way to handle this. Because of the directionality of the MVC cycle, this code will not translate directly. Before I just tell you what it should translate to, try your hand at it and I can verify. It is definitely not a copy-paste, but it involves similar parts (plus possibly one more that isn't in the above 3 lines).

web-client/public/js/grnstate.js
```diff
@@ -31,16 +32,21 @@ export const grnState = {
 
     newNetwork: false,
 
+    normalizationMax: null,
```
dondi: I'm observing that `normalizationMax` does not appear to be touched in this round, but we can wait-and-see for this one.

PR #632, opened 2018-05-22
Gene Page Changes Made 5/22/18

johnllopez616:
Addresses part of #620 and #621.

Regarding #620, I reduced the margin between the GRNsight logo header and the content of the gene page.

Regarding #621, I created a series of default values that read "no info". I eliminated the `filterData()` function and simplified the `getGeneInformationFunction()` to return a chain of promises that catch any errors when an AJAX call cannot be made from the `get[Database]Info()` functions, while still allowing the functions to modify the data of the default values. At the very end, the changed default values are passed into `info.js`.

Unfortunately, the issue regarding the JASPAR CORS error still persists without the use of a browser plugin. Additionally, with the plugin, the JASPAR function will not return due to the following error regarding loading the matrix:

`Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:5001' is therefore not allowed access.`

This might require us having to contact the developer, according to @mihirsamdarshi's suggestion.

### DIFF
--- a/web-client/public/gene/api.js
+++ b/web-client/public/gene/api.js
@@ -9,8 +9,6 @@
             var XMLParser = function (data) {
                 return serializer.serializeToString(data).replace(/\<.*?\>\s?/g, "");
             };
-          //  var apiErrors = [];
-          //  var apiCount = 5;
 
             var getUniProtInfo = function (geneSymbol) {
                 return $.get({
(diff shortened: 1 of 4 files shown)


johnllopez616:
After talking to @dondi at the meeting today, I changed the link so that JASPAR relies upon the pre-established relay so it can communicate with the server.

johnllopez616:
Pull Request Checklist
- [ ] Print works as expected

PR #633, opened 2018-05-23
Beta consolidation Round 2

dondi:
OK, here's what I was able to throw together. I'll add some commentary here and there so please scan the entire change log for any notes. This PR will be helpful for both @mihirsamdarshi and @johnllopez616, I'd say.

The TL;DR of this is that all of the loading/importing code has been moved to the new MVC cycle. The current network is now completely housed in `grnState`, loading/importing triggers are in `setupHandlers`, and there is more network redrawing code in `updateApp` now.

If you guys are good with these changes, you can go ahead and merge and _now_ try the gray threshold stuff. @mihirsamdarshi you probably can't really merge your git stash anymore, but you can potentially `git stash apply` it temporarily (merge conflicts and all) just as a frame of reference. Once you sort of get back what you need, then you can just discard those changes. (`git stash apply` instead of `pop` keeps the stash in your list so that you don't lose the code permanently)

Next note is that this PR is relatively large, but it does consist of multiple commits. I suggest that you also go over what changed one commit at a time, so that you can see the evolution from where we were Tuesday afternoon up to this point. Observing the phases I chose may be helpful to both of you in terms of your own choices in staging your work, so that the pieces stay small and manageable but the sum total is substantive but progressive.


web-client/public/js/constants.js
```diff
@@ -1,2 +1,23 @@
...
+const SCHADE_OUTPUT_ID     = "#schadeOutput";
+const SCHADE_OUTPUT_PATH   = "demo/schadeOutput";
+const SCHADE_OUTPUT_NAME   = "Demo #4: Weighted GRN (21 genes, 31 edges, Schade et al. 2004 data)";
+
+export const DEMO_INFORMATION = [
+    [ WEIGHTED_DEMO_ID,   WEIGHTED_DEMO_PATH,   WEIGHTED_DEMO_NAME   ],
+    [ UNWEIGHTED_DEMO_ID, UNWEIGHTED_DEMO_PATH, UNWEIGHTED_DEMO_NAME ],
+    [ SCHADE_INPUT_ID,    SCHADE_INPUT_PATH,    SCHADE_INPUT_NAME    ],
+    [ SCHADE_OUTPUT_ID,   SCHADE_OUTPUT_PATH,   SCHADE_OUTPUT_NAME   ]
+];
```
dondi: Imagine, all of these constants were in _upload.js_ but they served a single purpose. Nice to have them set aside cleanly as  a bunch of constants now. Note also that the only actual export needed was `DEMO_INFORMATION`—the rest are actually "private" constants. So again this helps clean the code up and clarify what is shared vs. not.

web-client/public/js/grnstate.js
```diff
@@ -1,3 +1,46 @@
...
+            link.stroke = "rgb(195, 61, 61)";     // Node coloring-consistent red edge color
+            network.positiveWeights.push(link.value);
+        } else {
+            link.type = "repressor";
+            // link.stroke = "DarkTurquoise";     // GRNsight v1 cyan edge color
+            link.stroke = "rgb(51, 124, 183)";    // Node coloring-consistent blue edge color
+            network.negativeWeights.push(link.value);
+        }
+    });
+};
```
dondi: The _import_ routines included this as part of the loading sequence. I chose to place this function in _grnstate.js_ because upon studying it, I noted that everything being done here is on the network—i.e., this is a 100% model-centric computation. So it felt right to make it part of the model module.

web-client/public/js/grnstate.js
```diff
@@ -1,3 +1,46 @@
...
+    newNetwork: false,
+
+    get network () {
+        return currentNetwork;
+    },
+
+    set network (network) {
+        currentNetwork = network;
+        this.newNetwork = true;
+    },
```
dondi: Here is the start of a getter and setter for the `network` property. I expect that the setter will grow more as additional code is refactored.

I was momentarily tempted to just integrate `annotateLinks` here. But if you look at the original code, `annotateLinks` is only called upon _import_ and not _open_. To play it safe I maintained this separation. However, you might discover that the operation is equally valid both ways, and if you do, then calling `annotateLinks` from with the setter can potentially streamline things.

web-client/public/js/setup-handlers.js
```diff
@@ -5,7 +5,11 @@ import {
     GREY_EDGES_DASHED_SIDEBAR
 } from "./constants";
 
+import { setupLoadAndImportHandlers } from "./setup-load-and-import-handlers";
```
dondi: We started by putting all of the load and import code right in _setup-handlers_ (see the individual commits), but when the main work was done, it made sense to separate them into their own module. Thus, the details are hidden away unless someone wants to look into that file, and here you only need the single high-level setup function.

web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -0,0 +1,152 @@
+import { updateApp } from "./update-app";
```
dondi: So this is a new file, but not new code—these are all transplants from _upload.js_, now cleanly separated to signal more clearly that these functions work in isolation of each other, solely for the purpose of loading and importing files.

dondi: Now that we have gained these very clean `import`/`export` constructs, we can now engage in the periodic exercise of separating out modules so that private activities remain private, with public/shared parts being very clearly stated via the `export` statements. Anything that isn't `export`ed is effectively private, and this gives you confidence that you can tinker with these internals without affecting other code unexpectedly.

web-client/public/js/grnstate.js
```diff
@@ -1,3 +1,46 @@
+let currentNetwork = null;
+
+const annotateLinks = network => {
+    // TODO This duplicates logic that is done on the server side for an .xlsx spreadsheet.
+    //      Think of a way to consolidate it. Having discovered this, it seems like this should
+    //      be done on the client side because it rearranges data redundantly, for ease of display.
```
dondi: This `TODO` comment was pre-existing and still looks like it is worth pursuing eventually.

web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -0,0 +1,152 @@
...
+            // displayStatistics(network);
+        }).error(networkErrorDisplayer);
+    };
+
+    /*
+     * Thanks to http://stackoverflow.com/questions/6974684/how-to-send-formdata-objects-with-ajax-requests-in-jquery
+     * for helping to resolve this.
+     */
+
+    // TODO Some opportunity for unification with loadGrn?
```
dondi: Another worthy `TODO`. There is definitely some similarity between `loadGrn` and `importGrn`. Consolidating those into one function will give us a nice little reduction in code.

web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -0,0 +1,152 @@
...
+        }).done(function (network) {
+            grnState.name = filename;
+            grnState.network = network;
+            grnState.annotateLinks();
+
+            reloader = () => importGrn(uploadRoute, filename, formData);
+
+            updateApp(grnState);
+        }).error(networkErrorDisplayer);
+    };
```
dondi: Note that these aren't _exactly_ the same `loadGrn` and `importGrn` functions from the original _upload.js_. They have been modified to fit the MVC cycle. Spot the differences for yourselves so that you can see how things changed.

web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -0,0 +1,152 @@
...
+
+    $("#reload").click(function () {
+        // Deleted `event` parameter but need `function` because of `this`.
+        if (!$(this).parent().hasClass("disabled")) {
+            if ($.isFunction(reloader)) {
+                reloader();
+            }
+        }
+    });
+};
```
dondi: The rest of this file is almost entirely copied from _upload.js_, the only difference being some code modernization. (`const`s, arrow functions, etc.)

web-client/public/js/update-app.js
```diff
@@ -14,7 +15,39 @@ const refreshApp = () => {
...
+    }
+
+    $("#fileName").text(name); // Set the name of the file to display in the top bar
+    $("input[type='range']").off("input"); // I have no idea why I do this. Investigate later.
+
+    // If more things need to be turned off, we'll add them to this array
+    [ "#resetSliders", "#resetSlidersMenu", "#undoReset", "#undoResetMenu" ].forEach(
+        selector => $(selector).off("click")
+    );
+};
```
dondi: @mihirsamdarshi This part may be of particular interest to you since you weren't sure how to handle `displayNetwork`. So note how it is now a helper function within _update-app_. I think that eventually, this might be integrated into `updateApp` itself. But for now we're keeping it separate since the original code only called this when a new network got loaded.

web-client/public/js/update-app.js
```diff
@@ -14,7 +15,39 @@ const refreshApp = () => {
...
+
 export const updateApp = grnState => {
+    if (grnState.newNetwork) {
+        displayNetwork(grnState.network, grnState.name);
+        refreshApp();
+
+        // Rare exception to the MVC cycle: right now we have no way of knowing whether the network has changed
+        // (which is what necessitates displayNetwork), so we mark the model here.
+        grnState.newNetwork = false;
+    }
```
dondi: Speaking of a "new network," here is some special-case logic that isn't fully MVC. Read the comment above and see if it makes sense.

I suspect, actually, that most of `displayNetwork` _is_ safe to call even when a network isn't freshly loaded or imported. So this special handling might become unnecessary eventually. But one step at a time.

web-client/public/js/upload.js
```diff
@@ -18,20 +18,6 @@ export const upload = function (sliderObject, sliderGroupController, drawGraph,
     var TOOLTIP_SHOW_DELAY    = 700;
     var TOOLTIP_HIDE_DELAY    = 100;
 
-  // Demo Stuff
```
dondi: OK, so this is a large diff, but note that it's _all deletions_. That's because we've now fully begun the process of breaking this huge monolith of code apart, separating it into the appropriate M-V-C modules. Eventually, _upload.js_ will cease to exist (I think).

web-client/public/js/upload.js
```diff
@@ -367,7 +151,7 @@ export const upload = function (sliderObject, sliderGroupController, drawGraph,
         var validated = edgeWeightNormalizationInputValidation(value);
         $("#normalization-max").val(validated);
         $("#edge-weight-normalization-factor-menu").val(validated);
-        drawGraph(currentNetwork, sliders, nodeColoring);
+        drawGraph(uploadState.currentNetwork, sliders, nodeColoring);
```
dondi: OK, _mostly_ deletions. The big transition here is that `currentNetwork` is no more; it has been fully transferred to `grnState`. `uploadState` is a temporary holding place for `currentNetwork`, to eventually be replaced fully by `grnState` as the transition progresses.

web-client/public/js/warnings.js
```diff
@@ -0,0 +1,98 @@
+export var displayWarnings = function (warnings) {
```
dondi: Another example of when to break out into a module: after studying this, I concluded that it was _completely_ self-contained. So it was a clear candidate for separation.

However, it doesn't have to be the only occupant in this file. You might find additional code relating to warnings that are cleanly separable. Now, you have a separate, clearly-named place to put that code.

web-client/public/js/warnings.js
```diff
@@ -0,0 +1,98 @@
...
+    var setPanel = (screenHeight - BORDER) + "px";
+    var minPanel = (MIN_SCREEN_HEIGHT - BORDER) + "px";
+    if (screenHeight > MIN_SCREEN_HEIGHT) {
+        $("#list-frame").css({height: setPanel});
+    } else {
+        $("#list-frame").css({height: minPanel});
+    }
+
+    $("#warningsModal").modal("show");
+};
```
dondi: OK, that's everything. Hope you find this useful. Again, try to look at the commits individually also; looking at them will give you an idea for how the code _evolved_ to this point. Looking at the PR alone misses the individual (smaller) steps that took things here. It's still potentially helpful to also look at these individual steps.

dondi:
Looks like we got a red X here due to the coverage drop, though it was quite small. Might be good to start adding unit tests against the new framework.

PR #636, opened 2018-05-23
Gene Page Changes Made 5/23/2018

johnllopez616:
Addresses the following issues in #620 

- [x] Let's make the hyperlinks on the page to be the main green GRNsight color, instead of multi-colored.  The buttons and section headers look nice with the different colors, but I think to be clear, the rest of the hyperlinks should just be the main GRNsight green color.  This affects both the General and Sources sections.
- [x] Let's use the Wikipedia convention of putting the source citations in square brackets like this <sup>[1]</sup>.
- [x] Let's make the source citations in the square brackets link to the sources section (like a Wikipedia page does).
- [x] In the Sources section, please move the list of citations over to the left-hand side, with the same margin as the other sections above it.
- [x] Let's regularize the font size and bolding for all the sections.  For example, in the regulation section, the field names are not bolded, in the protein section the font is really large.  Whatever the size they are in the general section is good (and bolded).
- [x] The protein sequence is not displaying correctly.  It has a leading space that should be removed and there are line breaks in the sequence that are not being respected when the window is maximized.
- [x] In the general section, it is "NCBI Gene ID" not just "NCBI ID".  NCBI has a lot of different databases and the one we are calling is the "Gene" Database.
- [x] The Jaspar sequence logo image and matrix need to be moved a little to the right.  The matrix is actually overlapping the edge of the section on the left-hand side.
- [x] Can the Jaspar sequence logo image be made a little smaller?

### DIFF
--- a/web-client/public/gene/api.js
+++ b/web-client/public/gene/api.js
@@ -250,6 +250,7 @@
                 } : {};
             };
 
+
             return $.when(
              getNCBIInfo(symbol)
            ).then(function (info1) {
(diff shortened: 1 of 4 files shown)


johnllopez616:
Pull Request Checklist
- [ ] Print works as expected

mihirsamdarshi:
@johnllopez616 is something like this supposed to happen? Also can we consider using the same sans serif font that the rest of GRNsight is using?
screen shot 2018-05-23 at 3 30 48 pm

kdahlquist:
To respond to @mihirsamdarshi's comment, I thought we were using the same sans serif font that the rest of GRNsight is using, or am I missing something.  That got changed earlier, right?

johnllopez616:
@kdahlquist @mihirsamdarshi The actual content on the gene page does not. I will make this change.

Issue #638, opened 2018-05-25
General Gene Page Bugs


johnllopez616:
I wanted to create a single issue thread regarding gene page issues relating to overall functionality or visual mishaps that are prominent enough to consider a bug rather than a stylistic change. 

API Functionality:

- [x] Whenever a user clicks on a node on beta, it does not link them to the gene page. It returns a 404 error instead.
- [x] If an API call is successfully made but returns a blank string instead of data, the "Not found" string does not load.
- [x] Whenever JASPAR makes an API call to a gene that doesn't exist, it returns undefined data, causing the above error to display, but also disallowing the overall error modal to display as well.

Visual Changes:

- [x] Source links attached at the end of gene page data would protrude into the accordion tabs.
- [ ] Add default "loading" text line to parts of the page

kdahlquist:
The "Page is loading" is an enhancement request and should get it's own issue, if the intent behind this issue is to report bugs.  I think this issue is closable, since I'm getting JASPAR data on gene pages.  If these bugs are fixed, then close this issue and open a new one with the enhancement request.

PR #637, opened 2018-05-25
Gene Page Changes Made 5/24/2018

johnllopez616:
Well. This took longer than anticipated.

Regarding #620, this update changes the font of the page and removes the superscripted links protruding into the headers of the accordion cards.

Regarding #621, the `jaspar-relay-controller` code is now happily in the `server` file with one caveat: The gene page had to be rewritten using Jade to support the dynamically changing JASPAR link.

Changes were made to `graph.js` to support this change, alongside `/server/app.js`, and `web-client/controllers/main.js.` @mihirsamdarshi Please check to see if any of these changes conflict with your branch, the next time you desire to pull from beta.

I understand that this PR should have consisted of multiple commits, but I wasn't aware of the scope of the changes required.

Additionally, I deleted `web-client/public/js/api-test.html`, a file leftover from the BioDB project not needed.

Issue #629, opened 2018-05-21
Investigate print css to get graph centered in what is printed


kdahlquist:
I just closed #365 because that was fixed, but as @NAnguiano noted, there is actually a problem with what is printed; the graph is cut off at the side.

johnllopez616:
![image](https://user-images.githubusercontent.com/21348600/40568868-3742a1d6-6032-11e8-8432-519bf2961feb.png)

It fits the page now, but it makes the nodes and the menu significantly smaller. Would you like those features changed as well?

Additionally, I don't even really see the functional need for the menu there...that can be removed from the print menu through CSS, however. What do you think?

kdahlquist:
@johnllopez616, agreed, we don't want the sidebar menu included in the print.  We also don't want the D-pad/zoom either if we can get rid of it.

I just tried printing off of both Firefox and Chrome.  In Chrome, the graph was completely on the printed page, but it wasn't for Firefox.

Let's see if getting rid of the sidebar and D-pad improves things.

johnllopez616:
![image](https://user-images.githubusercontent.com/21348600/40737103-9f27cdba-63f4-11e8-8392-88ece1668515.png)

Here's what I got from my Firefox. We will compare it to your computer's Firefox in the next PR.

kdahlquist:
I'm running Firefox v60.0.1 64-bit.  If that's what you have, hopefully it will look the same.

kdahlquist:
I'm going to mark this one good enough for now.  The D-pad and sidebar menu are gone.  I still see a phenomenon where the graph is cut off on the right.  I discovered that I can go into the Firefox page setup settings and select "shrink to fit page width" and then the graph will print smaller and nothing will be cut off.  However there is a blank second page when I do this.  Also weirdly, after printing the same graph 7 or 8 times at different zoom scale, the page header, sidebar, etc. came back.  I reloaded the page in the browser and that fixed that.

Anyway, I declare this good enough for now.  I'm going to stick a backlog label on this in case we want to revisit it in the future.  But we don't have time to continue messing with this now.

Issue #627, opened 2018-05-18
Rewording preference for "Format edges based on optimized weight parameters" and make it function like enable/disable node coloring


kdahlquist:
Upon reviewing the Client Side Testing Document #575, it occurred to me that there might be a simpler way to say "Format edges based on optimized weight parameters" vs. "Default to black edges with regular arrowheads".

For node coloring, we have a toggle that says either "Enable Node Coloring" vs. "Disable Node Coloring" that turns node coloring on and off in the presence of expression data.

In the case of edges, it could be "Enable Formatting of Edges Based on Weight Values" vs. "Disable Formatting of Edges Based on Weight Values".

This is a low priority item since it's fine now, but I wanted to record the thought in an issue.

kdahlquist:
I see the wording change on beta v3.0.13.  @johnllopez616, you are right, it is still not symmetrical with the way we enable/disable node coloring.

The functionality for enable/disable node coloring works like a toggle switch and the change is made immediately upon clicking.

The colored vs. black edge coloring feature was originally envisioned as a "preference" and requires a reload to take effect.

I think that, if it is possible, it would make more sense to turn the edge coloring on/off just like we do for node coloring and make the menu item work the same way.  To be clear, toggling the switch on/off would turn the edge coloring on/off immediately.

There is another bug associated with this feature, #591.

mihirsamdarshi:
I made a change so that upon clicking _Default to black edges with regular arrowheads_ the graph is reloaded

kdahlquist:
@mihirsamdarshi made it so that a reload is triggered when choosing the option.  However, we need to convert this to an enable/disable toggle switch like node coloring before closing this issue.

kdahlquist:
Actually, the enable/disable toggle doesn't work right for node coloring either in terms of the text on the menu item (the functionality works, though).

kdahlquist:
This is done.

PR #639, opened 2018-05-25
Gene Page Changes 5/25/2018 and Overall Bug Fixes

jlopez616:
First, all of the bugs listed and features requested in  #616,  #625, #627, #629 have been addressed. They will need review from @kdahlquist when the PR goes live.

Following PR #637, I realize the link to the new gene page on the right click is incorrect. I changed it here, which should be the same change I tried to make on the live beta. Additionally, I removed info.html.

Regarding #638, which I made today, every single line of the gene pages will have the "not found" sign appear should data be either missing or blank, including data pulled from JASPAR.

### DIFF
--- a/web-client/public/gene/api.js
+++ b/web-client/public/gene/api.js
@@ -186,8 +186,7 @@
                 };
 
                 for (var prop in uniprotTemplate) {
-                    if ((uniprotTemplate[prop] === undefined) ||
-                    (uniprotTemplate[prop] === null)) {
+                    if ((uniprotTemplate[prop] === undefined) || (uniprotTemplate[prop] === null)) {
                         uniprotTemplate[prop] = "Not found";
                     }
                 }
@@ -211,8 +210,7 @@
                 };
 
                 for (var prop in ncbiTemplate) {
-                    if ((ncbiTemplate[prop] === undefined) ||
-                    (ncbiTemplate[prop] === null)) {
+                    if ((ncbiTemplate[prop] === undefined) || (ncbiTemplate[prop] === null)) {
                         ncbiTemplate[prop] = "Not found";
                     }
                 }
@@ -252,8 +250,7 @@
 
                 for (var prop in yeastmineTemplate) {
 
-                    if ((yeastmineTemplate[prop] === undefined) ||
-                    (yeastmineTemplate[prop] === null)) {
+                    if ((yeastmineTemplate[prop] === undefined) || (yeastmineTemplate[prop] === null)) {
                         yeastmineTemplate[prop] = "Not found";
                     }
                 }
@@ -273,8 +270,7 @@
                 };
 
                 for (var prop in ensemblTemplate) {
-                    if ((ensemblTemplate[prop] === undefined) ||
-                    (ensemblTemplate[prop] === null)) {
+                    if ((ensemblTemplate[prop] === undefined) || (ensemblTemplate[prop] === null)) {
                         ensemblTemplate[prop] = "Not found";
                     }
                 }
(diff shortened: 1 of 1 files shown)


johnllopez616:
@dondi Ready for review

web-client/public/gene/api.js
```diff
@@ -178,18 +178,28 @@
...
+                var uniprotTemplate = {
                     uniprotID: XMLParser(data.getElementsByTagName("name")[0]),
                     proteinSequence: XMLParser(data.getElementsByTagName("sequence")[0]),
                     proteinType: XMLParser(data.getElementsByTagName("protein")[0].childNodes[1].childNodes[1]),
                     species: XMLParser(data.getElementsByTagName("organism")[0].childNodes[1]),
                 };
+
+                for (var prop in uniprotTemplate) {
+                    if ((uniprotTemplate[prop] === undefined) ||
+                    (uniprotTemplate[prop] === null)) {
```
dondi: Typically when a line break occurs mid-expression, the follow-up line is indented to indicate that the thought isn't "complete" yet. Alternatively, this particular expression doesn't actually exceed our 120-character max length yet, so can just be formatted as a single line.

web-client/public/gene/api.js
```diff
@@ -199,56 +209,97 @@
                     + XMLParser(data.getElementsByTagName("ChrStart")[0])
                      + ".." + XMLParser(data.getElementsByTagName("ChrStop")[0]) + ")",
                 };
+
+                for (var prop in ncbiTemplate) {
+                    if ((ncbiTemplate[prop] === undefined) ||
+                    (ncbiTemplate[prop] === null)) {
```
dondi: Ditto indentation.

web-client/public/gene/api.js
```diff
@@ -199,56 +209,97 @@
...
-                    cellularComponent: "N/A", // Information unavailable via regular API
+                    molecularFunction: "Not found", // Information unavailable via regular API
+                    biologicalProcess: "Not found", // Information unavailable via regular API
+                    cellularComponent: "Not found", // Information unavailable via regular API
                 };
+
+                for (var prop in yeastmineTemplate) {
+
+                    if ((yeastmineTemplate[prop] === undefined) ||
+                    (yeastmineTemplate[prop] === null)) {
```
dondi: Same.

web-client/public/gene/api.js
```diff
@@ -199,56 +209,97 @@
...
                     description: data.description,
-                    dnaSequence: "N/A", // Information unavailable via regular API
-                    geneLocation: "N/A", // Information unavailable via regular API
+                    dnaSequence: "Not found", // Information unavailable via regular API
+                    geneLocation: "Not found", // Information unavailable via regular API
                 };
+
+                for (var prop in ensemblTemplate) {
+                    if ((ensemblTemplate[prop] === undefined) ||
+                    (ensemblTemplate[prop] === null)) {
```
dondi: Yep.

web-client/public/gene/api.js
```diff
@@ -199,56 +209,97 @@
...
+                };
+
+                for (var prop in jasparTemplate) {
+                    if ((jasparTemplate[prop] === undefined) ||
+                    (jasparTemplate[prop] === null)) {
+                        jasparTemplate[prop] = "Not found";
+                    }
+                }
+
+                return jasparTemplate;
```
dondi: We're definitely going to need t capture this logic in a unit test. It will need an API-mocking tool like Sinon.js so that we can predictably return very specific responses in order to verify specific behaviors. If you can, see if you can do some advance reading on Sinon.js to get an idea for how this is done.

PR #640, opened 2018-05-29
Testing pull request

mihirsamdarshi:


dondi:
@mihirsamdarshi In answer to your questions:

> it seems that it can't parse the decimal values that I have inputted into the grnState network object, why would that be?

Do you mean this error?

                positiveWeights: [0 0.7715106466403678, 5.212343052095555, 0.826365326790911, 1.5180198066137216, 0.27964603867183396, 2.9870960868914778, 0.7743557796125339, 0.6701684007311992, 0.22069064019031245, 0.2841810246722045, 0.08852330457981578, 0.5563132195423768],
                                    ^^^^^^^^^^^^^^^^^^

    SyntaxError: Unexpected number

If so, sorry, you should be able to figure that out.

> Additionally, how would I go about adding more tests for reset normalization under the framework I have laid out?

I'm unclear on this one. You add more tests by declaring more `it` behaviors. Is that what you were asking? Or did you mean something else?

mihirsamdarshi:
Sorry, I realized what that error was seconds after posting that comment. Additionally, what I meant was that I was wondering whether I should create a test function in test.js in order to create multiple tests?

mihirsamdarshi:
Additionally @dondi , I am getting a syntax error from importing grnState, I suspect that I have put the file extension in wrong, however I am not sure why that is happening

```grnsight@3.0.9 test /Users/mihir/Documents/School/GRNsight_mihirsamdarshi
mocha --reporter list

/Users/mihir/Documents/School/GRNsight_mihirsamdarshi/test/grnstate-tests.js:3
import { grnState } from "/web-client/public/js/grnstate";
^^^^^^

SyntaxError: Unexpected token import
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at /Users/mihir/Documents/School/GRNsight_mihirsamdarshi/node_modules/mocha/lib/mocha.js:220:27
    at Array.forEach (native)
    at Mocha.loadFiles (/Users/mihir/Documents/School/GRNsight_mihirsamdarshi/node_modules/mocha/lib/mocha.js:217:14)
    at Mocha.run (/Users/mihir/Documents/School/GRNsight_mihirsamdarshi/node_modules/mocha/lib/mocha.js:469:10)
    at Object.<anonymous> (/Users/mihir/Documents/School/GRNsight_mihirsamdarshi/node_modules/mocha/bin/_mocha:404:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:383:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:496:3```

**Comment by dondi** [comment]
@mihirsamdarshi Sorry I’m still unclear about the tests question. In particular, I don’t really know what is intended when you said:
> I was wondering whether I should create a test function in test.js in order to create multiple tests?

I suspect that you are seeing something in our code base that is being interpreted as a requirement, but actually isn’t. To create multiple tests in Mocha, you just add more `describe` and `it` blocks, period. I think I will need to see a concrete example of what you mean, or conversely a concrete counterexample of something that can’t be achieved solely by adding more `describe` and `it` blocks, in order for me to fully understand what you’re after.

The second issue is the one I’d tackle first: additional configuration is necessary in order to make node “understand” the new import/export syntax. I will do some Google searching and post some leads.

**Comment by dondi** [comment]
Here is a possibility: https://github.com/mochajs/mocha/issues/2054

**Comment by mihirsamdarshi** [comment]
@dondi currently we have functions like `noErrors` and `duplicateGeneError` that we call in our mocha blocks of code, in files like _errors-graph-tests.js_. My question is should I consolidate the code that I have right now into a function similar to noErrors?

**Comment by mihirsamdarshi** [comment]
@dondi I do not believe we currently have a .babelrc file in our project

**Comment by mihirsamdarshi** [comment]
https://github.com/mochajs/mocha/issues/2655
this also may be a potential fix

**Comment by dondi** [comment]
@mihirsamdarshi Oh you meant those functions? OK, so that fits my initial theory:
> I suspect that you are seeing something in our code base that is being interpreted as a requirement, but actually isn’t.

So no, those functions were written to serve a very specific purpose for a very specific test pattern. Those are not conventions to follow. Just write your tests first. You would only concern yourself with helper functions like those if you see certain repeated patterns in the new test code. Until then, just write more `it` blocks.

**Comment by dondi** [comment]
@mihirsamdarshi OK so it looks like our priority is to get node to recognize import/export first. If you have already read up a bit on the issue, you know that the core need is to precompile the code into a form that doesn’t use import and export. It looks like mocha will need to be run with a setting that sends the code through Babel first.

See if you can get this to work by consulting the web for possible solutions and determining one that works for our specific setup. If you find that this is above your pay grade, let me know.

**Comment by mihirsamdarshi** [comment]
@dondi ok, makes senses. I did some research and it seems that we have to configure our compiler because node.js doesn't officially support es6 (meaning mocha does not either), should I go ahead and try to create a .babelrc file?

**Comment by mihirsamdarshi** [comment]
@dondi I found this tutorial that may be suitable for our purposes
http://jamesknelson.com/testing-in-es6-with-mocha-and-babel-6/

**Comment by dondi** [comment]
Sure, whatever works. You are exploring right now so should be in a position to try different things.

**Comment by mihirsamdarshi** [comment]
@dondi I think I fixed it, however can you double check the file path for grnState? I am not sure if there is supposed to be a period there or not

**Comment by mihirsamdarshi** [comment]
@dondi testing commits for resetNormalizationMax have been committed, as well the addition of babel, a .babelrc file, and the addition of `--compilers js:babel-core/register` to the testing script

File: test/grnstate-tests.js [hunk 1]
```diff
@@ -1,22 +1,23 @@
...
-            Build a network here. Because reset normalization max only relies on positiveWeights and negativeWeights,
-            your test network may contain just those properties.
+            positiveWeights: [0.7715106466403678, 5.212343052095555, 0.826365326790911, 1.5180198066137216, 0.27964603867183396, 2.9870960868914778, 0.7743557796125339, 0.6701684007311992, 0.22069064019031245, 0.2841810246722045, 0.08852330457981578, 0.5563132195423768],
+            negativeWeights: [-1.1672013354497313, -2.4438462486461163, -1.485274408559026, -0.7371671668856121, -2.9527940988546217, -3.3875125167057103, -0.0735729607899455, -0.12373154839822605, -0.5487650216350863, -0.442904423529279, -0.8662459811465724, -2.850006860838391, -1.5733259134754158, -0.02279063304065471, -1.9832439149172112, -0.25954746744906587],
         };
+        grnState.resetNormalizationMax.should.equal(5.212343052095555);
+    });
 
-        grnState.resetNormalizationMax.should.equal(whatever it should be);
+    it("should update the reset normalization maximum correctly", () => {
```
dondi: You have identical "should"s here. If that is the case, combine them under a single `it`.

You separate out `it`s when you are describing different behaviors.

test/grnstate-tests.js
```diff
@@ -1,22 +1,23 @@
-// var assert = require("chai").assert;
-// import { grnState } from "./grnstate";
-
-/* TODO: Framework for front-end test
-
-var testResetNormalization = (input, propertyExpected) => {
-    assert.equal(propertyExpected, grnState.propertyValue);
-}
+/* eslint-disable max-len */
+/* eslint no-unused-vars: [2, {"varsIgnorePattern": "should"}] */
```
dondi: Try to minimize these disables. Let me guess to see what you were trying to skip...

test/grnstate-tests.js
```diff
@@ -1,22 +1,23 @@
...
+import { grnState } from "../web-client/public/js/grnstate";
 
 describe("The grnState model object", () => {
-    it("updates the reset normalization maximum correctly", () => {
+    it("should update the reset normalization maximum correctly", () => {
         grnState.network = {
-            Build a network here. Because reset normalization max only relies on positiveWeights and negativeWeights,
-            your test network may contain just those properties.
+            positiveWeights: [0.7715106466403678, 5.212343052095555, 0.826365326790911, 1.5180198066137216, 0.27964603867183396, 2.9870960868914778, 0.7743557796125339, 0.6701684007311992, 0.22069064019031245, 0.2841810246722045, 0.08852330457981578, 0.5563132195423768],
+            negativeWeights: [-1.1672013354497313, -2.4438462486461163, -1.485274408559026, -0.7371671668856121, -2.9527940988546217, -3.3875125167057103, -0.0735729607899455, -0.12373154839822605, -0.5487650216350863, -0.442904423529279, -0.8662459811465724, -2.850006860838391, -1.5733259134754158, -0.02279063304065471, -1.9832439149172112, -0.25954746744906587],
```
dondi: Was this the reason for `disable max-len`? If so, then break lines! No one wants to read a very long array expression in a single line anyway.

test/grnstate-tests.js
```diff
@@ -1,22 +1,23 @@
...
-// import { grnState } from "./grnstate";
-
-/* TODO: Framework for front-end test
-
-var testResetNormalization = (input, propertyExpected) => {
-    assert.equal(propertyExpected, grnState.propertyValue);
-}
+/* eslint-disable max-len */
+/* eslint no-unused-vars: [2, {"varsIgnorePattern": "should"}] */
+var should = require("chai").should();
```
dondi: Can this be changed into an `import`, now that you have that working? No guarantees but it might help your second `eslint` exception.

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["es2015"]
+}
```
dondi: Confirmed, you don't need this as a separate file: https://babeljs.io/docs/usage/babelrc/ (see below for alternative)

package.json
```diff
@@ -30,6 +30,8 @@
     "xmlbuilder": "8.2.2"
   },
   "devDependencies": {
+    "babel-core": "^6.26.3",
+    "babel-preset-es2015": "^6.24.1",
```
dondi: Also confirmed, we don't need to add these lines. Our imports already have them (probably via `webpack`).

package.json
```diff
@@ -42,7 +44,7 @@
   },
```
dondi: As stated in https://babeljs.io/docs/usage/babelrc/, instead of a `.babelrc` file, you can put your settings in _package.json_. Insert this block above `"scripts"`:
```json
  "babel": {
    "presets": ["es2015"]
  },
```
I think I prefer this approach because it is more visible and keeps things in one place.

mihirsamdarshi:
@dondi, can you see why a TypeError comes up for .equal? I imported should as the documentation said, however I reverted to the old method of `var should = require...` because this error would pop up if I used import initially

test/grnstate-tests.js
```diff
@@ -1,22 +1,23 @@
-// var assert = require("chai").assert;
-// import { grnState } from "./grnstate";
-
-/* TODO: Framework for front-end test
-
-var testResetNormalization = (input, propertyExpected) => {
-    assert.equal(propertyExpected, grnState.propertyValue);
-}
+import { should } from "chai"
```
dondi: Let me know if you get stuck on this one.

dondi:
@mihirsamdarshi Read this: http://www.chaijs.com/guide/styles/#should

Once you have a full understanding for how `should` gets installed, the reason for `TypeError: Cannot read property 'equal' of undefined` should become evident.

And if it doesn't let me know. I want to build up your skill set at understanding root causes of problems, because that understanding will lead you to solutions more quickly.

mihirsamdarshi:
@dondi I guess it was because I wasn't calling should, however the docs also state that I could do `import "chai/register-should"` and I should have gotten the same result

dondi:
@mihirsamdarshi Yes, the docs state that `import "chai/register-should"` is needed. What happened when you tried that?

mihirsamdarshi:
The TypeError recurred when using `import "chai/register-should"`

mihirsamdarshi:
however, the latest commit uses 
`import chai from "chai";
chai.should();`

dondi:
@mihirsamdarshi Yes, according to the docs, both forms will work, but the second one is a bit of a hack, which is why they came up with `import "chai/register-should"`.

If you look up the history of `import "chai/register-should"`, you'll discover that this was introduced with Chai version 4.0.0, and we currently have version 1.10.0.

So you have a choice:
* Upgrade Chai to the latest 4.x.x version so you can use the cleaner `import "chai/register-shoud"`
* _or_ keep Chai at its current 1.10.0 level but be forced to use the `chai.should()` workaround

Which way would you rather go?

mihirsamdarshi:
I believe upgrading is the right way to go

dondi:
OK @mihirsamdarshi give it a shot

dondi:
(upgrading Chai, that is)

package.json
```diff
@@ -7,6 +7,7 @@
...
   "dependencies": {
     "body-parser": "^1.18.2",
+    "chai": "^4.1.2",
```
dondi: I missed that the new line for `chai` was in the `dependencies` block. It is a _developer_ dependency (i.e., you don't need Chai to run GRNsight).

package.json
```diff
@@ -40,9 +41,14 @@
     "mocha-lcov-reporter": "^1.3.0",
```
dondi: What should really have happened is that line 34 above was replaced with the new version designation.

PR #648, opened 2018-05-30
move gray edge controller to single state controller

mihirsamdarshi:


web-client/public/js/setup-handlers.js
```diff
@@ -20,18 +22,28 @@ export const setupHandlers = grnState => {
...
     });
 
-    $("#edge-weight-normalization-factor-menu").on("change", function () {
+    $("#edge-weight-normalization-factor-menu").change(() => {
         grnState.normalizationMax = $("#edge-weight-normalization-factor-menu").val();
         updateApp(grnState);
     });
+
+    $(GREY_EDGE_THRESHOLD_MENU).change(() => {
+        grnState.grayEdgeThreshold = Math.round(($(GREY_EDGE_THRESHOLD_MENU).val()));
```
dondi: This looks like there's one too many pairs of parentheses inside the `Math.round`?

web-client/public/js/setup-handlers.js
```diff
@@ -20,18 +22,28 @@ export const setupHandlers = grnState => {
...
+        updateApp(grnState);
+    });
+
+    $(GREY_EDGE_THRESHOLD_SLIDER_SIDEBAR).change(() => {
+        grnState.grayEdgeThreshold = Math.round(($(GREY_EDGE_THRESHOLD_SLIDER_SIDEBAR).val() * 100));
```
dondi: Same here, I don't think you need the one immediately inside the `round()` argument.

web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -117,16 +117,20 @@ export const setupLoadAndImportHandlers = grnState => {
...
-        // deleted event parameter
+    $("#upload").click(() =>
         $("#launchFileOpen").off("click").on("click", () => $("#upload").click())
     );
+    $("#upload-sif").click(() =>
+        $("#launchFileOpen").off("click").on("click", () => $("#upload-sif").click())
+    );
+    $("#upload-graphml").click(() =>
+        $("#launchFileOpen").off("click").on("click", () => $("#upload-graphml").click())
+    );
```
dondi: OK, cool you saw this; I noticed that it was left over from the earlier round.

PR #651, opened 2018-05-30
changed grnstate object property to a single string

mihirsamdarshi:


web-client/public/js/constants.js
```diff
@@ -37,3 +37,7 @@ export const WEIGHTS_HIDE_SIDE             = "#weightsNeverSide";
 export const WEIGHTS_SHOW_MOUSE_OVER_CLASS = ".weightsMouseOver";
 export const WEIGHTS_SHOW_ALWAYS_CLASS     = ".weightsAlways";
 export const WEIGHTS_HIDE_CLASS            = ".weightsNever";
+
+export const showWeightsMouseover = "showWeightsMouseover";
+export const showAllWeights = "showAllWeights";
+export const hideAllWeights = "hideAllWeights";
```
dondi: Note our convention with constants: all caps, underscore to separate words. It isn't a necessity but it makes constants easier to spot when looking at code.

web-client/public/js/grnstate.js
```diff
@@ -46,9 +49,7 @@ export const grnState = {
         this.newNetwork = true;
     },
 
-    showWeightsMouseover: true,
-    showAllWeights: false,
-    hideAllWeights: false,
+    edgeWeightDisplayOption: showWeightsMouseover,
```
dondi: For example here, just reading this line you're not sure if `showWeightsMouseover` is a variable or a constant. Making it all caps like `SHOW_WEIGHTS_MOUSEOVER` makes this clearer without needing a color-coding editor.

Make sure to expand your awareness of the code toward any conventions that exist. If something is being done in the code for no clear reason, it is worthwhile to ask.

When I work with other programmers, a positive sign (for me) is when they notice the _non_-existence of coding conventions. It means that they are actively looking at the source code and recognizing patterns in style and organization. And they can then tell when such patterns are insufficient or non-existent.

web-client/public/js/setup-handlers.js
```diff
@@ -8,6 +8,9 @@ import {
     WEIGHTS_SHOW_MOUSE_OVER_CLASS,
     WEIGHTS_SHOW_ALWAYS_CLASS,
     WEIGHTS_HIDE_CLASS,
+    showWeightsMouseover,
+    showAllWeights,
+    hideAllWeights,
```
dondi: So, notice how these new constants kind of stick out like a sore thumb when compared to the constants that came before them 🤔

Issue #591, opened 2018-04-26
When default to black edges is selected the various edge thicknesses still vary


mihirsamdarshi:
When default to black edges is selected the various edge thicknesses still vary

mihirsamdarshi:
This issue is fixed

kdahlquist:
Confirmed fixed in beta v3.0.16/3.1.0

PR #659, opened 2018-05-31
move colorOptimal in order to correct issue #591

mihirsamdarshi:


web-client/public/js/graph.js
```diff
@@ -43,11 +43,8 @@ export var drawGraph = function (network, sliderController, nodeColoring) {
         return node.name.length * 12 + 5;
     };
 
-    /* If colorOptimal is false, then weighting is ignored,
+    /* If .colorOptimal is false, then weighting is ignored,
     and the lines are all drawn as if it was an unweighted sheet */
```
dondi: This comment seems attached to the lines that you deleted; is there still reason for it to be here?

web-client/public/js/graph.js
```diff
@@ -1,4 +1,5 @@
 import Grid from "d3-v4-grid";
+import { grnState } from "./grnstate";
```
dondi: I see why it was necessary for you to do this, but let's make sure this is temporary. You treat `grnState` as read-only here, so that's good, in-keeping with the MVC flow. Eventually, that code should somehow make it to `updateApp` instead. But I can see from where it is used that this case is not as simple as the others, so this intermediate stage is OK.

web-client/public/js/setup-handlers.js
```diff
@@ -69,4 +69,13 @@ export const setupHandlers = grnState => {
...
 
+    $("#colorEdges").click(() => {
+        grnState.colorOptimal = true;
+        updateApp(grnState);
+    });
+
+    $("#blackEdges").click(() => {
+        grnState.colorOptimal = false;
+        updateApp(grnState);
+    });
```
dondi: So yes, these are the handlers which update `grnState` based on user actions on `#colorEdges` and `#blackEdges`, but shouldn't M->V code also exist in `updateApp`? i.e., Set `#colorEdges` and `#blackEdges` to the appropriate state based on the value of `grnState.colorOptimal`?

web-client/public/js/update-app.js
```diff
@@ -124,6 +126,11 @@ const synchronizeHideAllWeights = () => {
     $(WEIGHTS_HIDE_CLASS).addClass("selected");
 };
 
+const synchronizeColorOptimalValues = function () {
+    $(COLOR_PREFERENCES_CLASS).toggleClass(ACTIVE_COLOR_OPTION);
+    $(COLOR_PREFERENCES_CLASS + ">span").toggleClass("glyphicon-ok invisible");
```
dondi: I don't get this—`grnState` is not influencing this code at all. Instead, `toggleClass` is being used, which alternates between two states. That doesn't sound right. If `grnState.colorOptimization` changes in a way that is out-of-sync with toggling, then the states of the UI elements will be wrong.

Base your view code _directly_ on the status of `grnState`. `toggleClass` is reliant on the status quo—which isn't under the control of this code.

web-client/public/js/upload.js
```diff
@@ -44,20 +42,6 @@ export const upload = function (sliderObject, sliderGroupController, drawGraph,
...
     sliders.configureSliderControllers();
 
-    var settingsController = function () {
-        this.color = true;
-
-        this.setupSettingsHandlers = function () {
-            $(COLOR_PREFERENCES_CLASS).on("click", function () {
-                $(COLOR_PREFERENCES_CLASS).toggleClass(ACTIVE_COLOR_OPTION);
-                $(COLOR_PREFERENCES_CLASS + ">span").toggleClass("glyphicon-ok invisible");
-            });
```
dondi: OK, so I see what happened here—you just moved this code to `updateApp`. That is generally the right idea, but don’t do so blindly either. Always keep the M->V->C cycle in mind.

mihirsamdarshi:
@dondi that was a temporary fix, the latest one is up

mihirsamdarshi:
all changes requested by @dondi are fixed, will wait for travis build to finish then merge

PR #685, opened 2018-10-10
adding new, non-functional changes

mihirsamdarshi:


johnllopez616:
Mihir, when the checks pass, and when both of our pull requests are made, let me know on slack how you would like to proceed with the release.

dondi:
@mihirsamdarshi The function you refer to is clearly intended to be a constructor. In other words, it is meant to create objects. I don’t think that function needs to be available to the outside world. Instead, I would define that function in private, then create slider objects from that function for everything that needs a slider. Then those objects would be what go on grnstate.

More radically, it is also possible that with this new architecture, we might not need slider objects as objects per se at all. Have you tried the possibility of just representing the sliders’ values as a simple scalars in grnstate? Then update based on that scalar, and change that scalar when an event occurs in the slider. I wouldn’t be shocked if that approach does work. Unless you’ve already tried it and found something that prevents it from working.

mihirsamdarshi:
Removing the objects altogether was what I was thinking about doing, however I believed that it would take a longer time, since much of the `sliderGroupController` code was based on the object. This was why I wanted to see if I could put the constructor within the grnState object. I will push my additional changes tonight, and should be ready to do a merge into beta tonight with @johnllopez616

dondi:
Just noting here that the continuous integration tools are still rejecting this branch, and as far as I can tell, the issue is very addressable (undefined variables). I'd rather not approve this until the red X goes away. Thanks!

mihirsamdarshi:
Yes, I forgot to mention that I plan to re evaluate the constructor move and do some additional work before doing a release to beta.

dondi:
Understood @mihirsamdarshi but that is not a reason to leave your code broken, particularly given that the errors are easily addressed. In an industry environment where more stuff is happening in parallel, your co-workers will trust your work less if the build system rejects it.

dondi:
@mihirsamdarshi Sorry wasn't able to get to this last night. I haven't taken a super close look at it, but based on the newest commits things are looking like they are headed in the right direction. So far so good.

Issue #691, opened 2018-10-11
When sequence logo is not found on gene page, don't display the "image not found" box


kdahlquist:
When a JASPAR sequence logo image is not found because there is no JASPAR data to display, the browser shows an image box for the not found image.  We want to get rid of this box and just have it say "Not found" in text.

johnllopez616:
I believe I have corrected this issue. Will show at the next pull request

kdahlquist:
Fixed in beta 3.1.7

Issue #690, opened 2018-10-11
Gene page date needs to be lined up in 2 column format as before


kdahlquist:
A consequence of the refactor from #658 is that the two column format for displaying data on the gene page got messed up. It needs to look like what is currently on master (v3.1.3)

johnllopez616:
I believe to have fixed it. Is this appropriate?

![image](https://user-images.githubusercontent.com/21348600/47472996-97cb1c80-d7c5-11e8-9d64-e51edd62e523.png)

kdahlquist:
From what I saw, it looks good.
1. The sequence logo from Jaspar needs to be made smaller again.
2. The sources are missing their hyperlinks.

Issue #679, opened 2018-10-03
Incorrect JASPAR data lingering in gene page


kdahlquist:
If I load a dummy network in to beta with gene names that are not real genes (A, B, C), and open the gene page, there is lingering data from JASPAR that should not be there.

Under General Information it says:
JASPAR ID: MA0265.1

A sequence logo and frequency matrix (presumably from that ID shown above) is also displayed.

I'm attaching the file I used to trigger this.
[A-B-C_network.xlsx](https://github.com/dondi/GRNsight/files/2444096/A-B-C_network.xlsx)

Instead, "Not found" should be displayed.

kdahlquist:
Even though we have a kind of solution to this, I'm going to leave it open so that @johnllopez616 can go back and fix the queries as discussed in the previous comment.

dondi:
@dondi's commit comment can be found here: https://github.com/dondi/GRNsight/commit/c36323dbf9539b6f0fe5cdf8f9ccece80aaa6f86

Also @johnllopez616 let's shadow you the next time you refresh beta, in order to see what the gap in the documentation might be which is preventing successful refreshes.

kdahlquist:
@dondi commented on the commit with specifics, but what we want you to do is search for the `name` in the API request itself, not in the results that come back from a general search.

Please let us know if you have questions.

After we get the JASPAR query cleaned up, we are going to need to review the API requests for the other databases to make sure that they are not falling into this trap as well.  Furthermore, we see that the species (taxon id for _Saccharomyces cerevisiae_) is hard coded into the JASPAR query.  We are going to need to write a functionality so that the species is dynamically filled by whatever is found by the UniProt query.  Separate issues will be opened for each of these issues.

johnllopez616:
Changes reflected in PR #702

kdahlquist:
PR #702 will be updated with final changes requested by @dondi.  We will be able to close this issue when the changes are up on beta and @kdahlquist can review.

dondi:
This is now running on beta 3.1.8 where the following issues are observed:
![image](https://user-images.githubusercontent.com/382242/48585090-6800c780-e8e0-11e8-8756-82dfc3bec348.png)
That is an apparent 400 on Ensembl and 503 on Jaspar. Let's focus on tracking down these errors before closing this issue.

johnllopez616:
I will investigate this during my next work session using `curl` requests.

johnllopez616:
[...]

PR #693, opened 2018-11-01
Gene Page Changes 10/31/2018

johnllopez616:
Addresses the following issues:

#691 - Displays a "not found" text line when an image is unable to be retrieved.
#690 - Displays the information in the columns as previously desired, makes the sequence logo smaller, and gives the sources their hyperlinks
#679 - The new functionality is as follows: it looks through the search results, returned as an array, and looks for the gene with the same name. This is how the API is able to know whether or not the data is correct.

### DIFF
--- a/web-client/public/gene/api.js
+++ b/web-client/public/gene/api.js
@@ -175,7 +175,19 @@ let getJasparInfo = function (geneSymbol) {
             xhr.setRequestHeader("content-type", "application/json");
         },
     }).then(function (data) {
-        return (data.count !== 1 ? {} :
+        let value = undefined;
+        const nameCheck = (list) => {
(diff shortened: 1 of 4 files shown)


web-client/public/gene/api.js
```diff
@@ -175,7 +175,19 @@ let getJasparInfo = function (geneSymbol) {
             xhr.setRequestHeader("content-type", "application/json");
         },
     }).then(function (data) {
-        return (data.count !== 1 ? {} :
+        let value = undefined;
+        const nameCheck = (list) => {
+            if (list !== []) {
```
dondi: Note that `[] !== []` is `true`. If you are checking for an empty list, do something different.

Issue #701, opened 2018-11-08
iFrame Menu at Bottom of Gene Page is still not aligned to where it needs to be


johnllopez616:
Contrary to the closure of issue #689, the iFrame at the bottom of the gene page is cut off from the left. I have displayed it in the image below.

![image](https://user-images.githubusercontent.com/21348600/48171620-aa5b5080-e2b1-11e8-908b-83061befea88.png)

johnllopez616:
![image](https://user-images.githubusercontent.com/21348600/48173464-b0a0fb00-e2b8-11e8-8d51-58c94d68422c.png)

As of now, I can achieve the following desired bottom display, which differs from the display on the main GRNsight website.

Is it absolutely imperative that it must align precisely with what is on the GRNsight page? I understand that it is best practice to have things synchronized with each other. However, in your opinion, would this be acceptable?

![image](https://user-images.githubusercontent.com/21348600/48173489-d0d0ba00-e2b8-11e8-9137-5d5178b46c23.png)

johnllopez616:
Changes reflected in PR #702

johnllopez616:
In order to achieve the desired effect, several changes were made in gh-pages via the following commit: https://github.com/dondi/GRNsight/commit/e2531fa30cdb5b9fc5c78a41217d60e3d0332a29

kdahlquist:
looks good on beta 3.1.8

Issue #689, opened 2018-10-11
Menu is not centered at bottom of gene pages


kdahlquist:
When @johnllopez616 was fixing the items in issue #658, a new issue came up where the links at the bottom of the gene pages are not being centered as they are on other GRNsight pages.  He will now look into this.

johnllopez616:
I'm going to ask Dondi for help with this one. I cannot figure out the behavior of this iFrame for the life of me.

PR #702, opened 2018-11-08
JASPAR API call changed and iFrame Menu changed for Gene Page

johnllopez616:
Addresses issues #679 and #701 

This PR adds the desired functionality as discussed in the comments of #679 by modifying the JASPAR request call.

This PR also provides a solution to #689 and #701, to be viewed by @kdahlquist.


johnllopez616:
Pull Request Checklist
- [ ] Print works as expected

web-client/public/gene/api.js
```diff
@@ -169,34 +169,20 @@ let getEnsemblInfo = function (geneSymbol) {
...
-                    if (gene.name === geneSymbol) {
-                        value = gene;
-                    }
-                });
-            }
-            return value;
-        };
-
-        return ((nameCheck(data.results) === undefined) ? {} :
+        return (data.results.length === 0) ? {} :
```
dondi: Does the API guarantee that the response will have a `results` property, and that it will be an array? Make sure that those conditions are certain. Otherwise, you will need additional guard code to check whether `data.results` is undefined/null before inquiring about its length.

web-client/public/gene/info.css
```diff
@@ -40,16 +40,17 @@ sup {
...
   position: absolute;
   min-height: 450px;
   min-width: 160px;
+  z-index: -1;
 }
 
 #iframeFooter {
   width: 100%;
 	display: block;
+  z-index: 0;
```
dondi: I don't see how this fixes the issue. `z-index` is not used for layout changes. If this appears to fix things (and I don't think so, because I tried to apply the changes manually in the browser without a visible effect), then I think that appearance is coincidental.

The true way to fix this is to make the gene page layout structure match the GRNsight page layout structure. Upon inspection of the latter, the footer is actually part of the main section (`#mainContent`) on the GRNsight pages and so it retains a margin that offsets it from the sidebar. In the gene page, the footer is _both_ the sidebar and main content.

Based on this, it looks like the right first step is to put the footer iframe within `#mainPage`, which appears to be the closest equivalent, but in structure and layout, to `#mainContent` on the GRNsight pages. Once it is placed there, adjustments can then be made to ensure a better visual match.

@johnllopez616 remember that web page editing and layout is a big-picture activity. It can be counterproductive to make small pinpoint changes here and there when in reality, stepping back, there are broader structural issues that keep things from lining up (like what appears to be happening above). Remember also to not form some type of tunnel vision, where you might be so focused on a particular visual detail without realizing that changes are then happening to other parts of a page. The ultimate goal of web page design activities is to provide something that is simultaneously useful, informative, functional, and aesthetically pleasing. Maintaining all of these factors requires that you keep a birds-eye view of the page even while working on details, in order to avoid regressions and to catch smaller glitches that might not have been mentioned explicitly.

web-client/public/gene/api.js
kdahlquist: @johnllopez616 will fix this and update the pull request and send it to @dondi for one more review.  @johnllopez616 and @mihirsamdarshi will get together to do the pull request checklist and walk through updating the server.

web-client/public/gene/info.css
kdahlquist: @johnllopez616 made changes in gh-pages that interact with the z-index. He says he will document that.

johnllopez616:
Need to meet with @mihirsamdarshi to discuss the process for refreshing the server in order to avoid a total disruption of beta

PR #706, opened 2018-11-29
finishing up sliders move & starting node-coloring

mihirsamdarshi:


web-client/public/js/update-app.js
```diff
@@ -134,24 +137,45 @@ const synchronizeHideAllWeights = () => {
...
-const disableColorOptimal = function () {
+const disableColorOptimal = () => {
     $(COLOR_EDGES).removeClass(ACTIVE_COLOR_OPTION);
     $(COLOR_EDGES + ">span").removeClass("glyphicon-ok invisible");
     $(BLACK_EDGES).addClass(ACTIVE_COLOR_OPTION);
     $(BLACK_EDGES + ">span").addClass("glyphicon-ok");
 };
 
-export const updateSliderDisplayedValue = (slider, element) => {
+const updateSliderDisplayedValue = (slider, element) => {
```
mihirsamdarshi: my problem here is that most of grnState's properties are T/F or a set of defined values, with certain defined actions for update-app to take. However, I don't understand how to trigger an action based on a change to the grnState value for each slider. The method that I could think of would be to create a true/false property in grnstate, however this would likely circumvent the MVC model. I would set the property as equal to true if there was a change in setup-handlers, however, I would have to set it back to false after I ran `updateSliderDisplayedValue` which would be contained in setup-handlers

web-client/public/js/sliders.js
```diff
@@ -83,121 +40,13 @@ export var sliderGroupController = function (sliderArray) {
...
-        slider.setCurrentVal(value);
-    };
-    */
-
-    this.setCurrentVal = function (newVal) {
-        this.currentVal = newVal;
-    };
-
+// Object error when moving it
     this.initializeDefaultForces = function () {
```
mihirsamdarshi: this function is triggered in graph.js, however, if I just replace `sliderController.initializeDefaultForces();` with `modifyChargeParameter(-50);` & `modifyLinkDistanceParameter(500)` then I get the following error, which I have never seen before.  A google search associates it with TypeScript (?).

screen shot 2018-11-29 at 2 03 53 pm

web-client/public/js/graph.js
```diff
@@ -1467,10 +1467,10 @@ export var drawGraph = function (network, sliderController, nodeColoring) {
     }
 
     // Configures sliderController
-    sliderController.addForce(simulation);
-    sliderController.configureForceHandlers();
+    grnState.simulation = simulation;
     sliderController.initializeDefaultForces();
 
+    // figure out this code???
```
mihirsamdarshi: Additionally, I need a little bit of help figuring out what lines 1475 & 1476 are actually doing, in order to move that last function in sliders.js

dondi: `value` is not being compared to anything; it is the target of an assignment statement. The value assigned to `value` is either `linkDistValidator($(item).val())` if the value of `slider` is `"link"`. Otherwise, `value` gets `chargeValidator($(item).val())`.

Let me know if that clears that up.

web-client/public/js/setup-handlers.js
```diff
@@ -1,4 +1,6 @@
 import { updateApp } from "./update-app";
+import { modifyChargeParameter } from "./update-app";
+import { modifyLinkDistanceParameter } from "./update-app";
```
dondi: You can group multiple imports from the same file into a single list (actually, just like below).

web-client/public/js/sliders.js
dondi: The error means that something that the code is trying to treat like a function is actually an Object and not the function that the code thinks it is.

It can be easily tracked down by ascertaining the exact value that it is the expression before `(-50)` in the original code.

However, I need to see the version of the source code that triggers this error in order to explain more fully. What you're seeing here is already processed by `webpack` and is harder to debug in this form. Easier to debug in the pre-`webpack` form of the code.

web-client/public/js/update-app.js
dondi: At face value, `updateSliderDisplayedValue` takes the value stored in `element` and transfers that, possibly with an appended `0`, to the slider. So, the model in question is the said value. What you are seeing here is a view-to-view copy of data (`element` going to `slider`).

To MVC this, the _value itself_ is what goes in `grnState`. And then, when that value changes, that `value` should go to _both_ `element` and `slider`.

So the flow is: `grnState` holds the master copy of the value (whatever it is). When it updates the app, this value is placed in both `element` and `slider` (whatever those are). If `element` changes this value (which is what seems to be implied because the value is being copied from `element`), that updates `grnState` to the new value. `updateApp` then sets that new value to (again) both `element` and `slider`.

That might be easier drawn with a picture but see if that description clears that up.

web-client/public/js/graph.js
```diff
@@ -1484,39 +1484,42 @@ export var drawGraph = function (network, nodeColoring) {
...
-        var value = slider === "link" ? linkDistValidator($(item).val()) :
-            chargeValidator($(item).val());
+    const changeSliderValue = function (slider) {
         if (slider === "link") {
-            grnState.linkDistanceSlider.currentVal = value;
-            modifyLinkDistanceParameter(value);
-            $(LINK_DIST_VALUE).text(value);
-            $(LINK_DIST_SLIDER_ID).val(value);
-            $(LINK_DIST_MENU).val(value);
+            modifyLinkDistanceParameter(grnState.linkDistanceSlider.currentVal);
```
mihirsamdarshi: @dondi I have implemented somewhat of an MVC within graph.js, however I am hesitant to move it, because it would require me to call changeSliderValue (ideally contained within update-app.js) directly from within setup-handlers.js. A workaround would be for a T/F value to be set in the handler function, however that would have to be reset from within update-app.

mihirsamdarshi: Wait, do I not actually need to force changeSliderValue to be called?

web-client/public/js/setup-handlers.js
```diff
@@ -14,6 +14,12 @@ import {
     LOCK_SLIDERS_CLASS,
     UNDO_SLIDER_RESET_CLASS,
     RESET_SLIDERS_CLASS,
+    RESET_SLIDERS_MENU_OPTION,
+    UNDO_SLIDER_RESET_MENU,
+    /* LINK_DIST_MENU,
+    LINK_DIST_SLIDER_ID,
+    CHARGE_MENU,
+    CHARGE_SLIDER_ID, */
```
dondi: These are commented-out so can be removed.

web-client/public/js/update-app.js
```diff
@@ -33,13 +34,26 @@ import {
...
   UNDO_SLIDER_RESET_BUTTON,
   UNDO_SLIDER_RESET_MENU,
+  LINK_DIST_SLIDER_ID,
+  LINK_DIST_MENU,
+  LINK_DIST_DEFAULT_VALUE,
+//  LINK_DIST_VALUE,
+  CHARGE_SLIDER_ID,
+  CHARGE_MENU,
+  CHARGE_DEFAULT_VALUE,
+//  CHARGE_VALUE,
```
dondi: A couple more commented-out lines that can go.

web-client/public/js/update-app.js
```diff
@@ -212,52 +232,65 @@ export const updateApp = grnState => {
...
-        grnState.chargeSlider.backup = grnState.chargeSlider.currentVal;
-        grnState.linkDistanceSlider.backup = grnState.linkDistanceSlider.currentVal;
-        grnState.chargeSlider.currentVal = grnState.chargeSlider.defaultVal;
-        grnState.linkDistanceSlider.currentVal = grnState.linkDistanceSlider.defaultVal;
-        $("#charge-menu").val(grnState.chargeSlider.defaultVal);
-        $("#link-distance-menu").val(grnState.linkDistanceSlider.defaultVal);
+        CHARGE_BACKUP_VALUE = CHARGE_CURRENT_VALUE;
+        LINK_DIST_BACKUP_VALUE = LINK_DIST_CURRENT_VALUE;
+        CHARGE_CURRENT_VALUE = CHARGE_DEFAULT_VALUE;
+        LINK_DIST_CURRENT_VALUE = LINK_DIST_DEFAULT_VALUE;
```
dondi: So to verify, you are storing these values in a separate location, for later use? If that is so, use camel case rather than all caps.

web-client/public/js/update-app.js
```diff
@@ -212,52 +232,65 @@ export const updateApp = grnState => {
...
+        $(LINK_DIST_MENU).val(LINK_DIST_DEFAULT_VALUE);
     };
 
     const undoReset = () => {
-        grnState.chargeSlider.currentVal = grnState.chargeSlider.backup;
-        grnState.linkDistanceSlider.currentVal = grnState.linkDistanceSlider.backup;
-        $("#charge-menu").val(grnState.chargeSlider.backup);
-        $("#link-distance-menu").val(grnState.linkDistanceSlider.backup);
+        CHARGE_CURRENT_VALUE = CHARGE_BACKUP_VALUE;
+        LINK_DIST_CURRENT_VALUE = LINK_DIST_BACKUP_VALUE ;
```
dondi: Ditto with naming convention.

web-client/public/js/update-app.js
```diff
@@ -212,52 +232,65 @@ export const updateApp = grnState => {
...
 
     const updateLinkDistanceSliderValues = () => {
-        $(grnState.linkDistanceSlider.sliderId).val(grnState.linkDistanceSlider.currentVal);
-        $(grnState.linkDistanceSlider.valueId).html(grnState.linkDistanceSlider.currentVal +
+        $(LINK_DIST_SLIDER_ID).val(LINK_DIST_CURRENT_VALUE);
+        $(LINK_DIST_SLIDER_ID).html(LINK_DIST_CURRENT_VALUE +
           ((grnState.linkDistanceSlider.needsAppendedZeros
-            && grnState.linkDistanceSlider.currentVal.toString().length ===
-              GRAVITY_LENGTH_WITHOUT_ZERO) ? "0" : ""));
+            && LINK_DIST_CURRENT_VALUE.toString().length === GRAVITY_LENGTH_WITHOUT_ZERO) ? "0" : ""));
```
dondi: These lines (275-277) follow the same pattern as lines 267-270, so that looks like a good candidate for "pulling out" as a reusable function.

web-client/public/js/update-app.js
```diff
@@ -212,52 +232,65 @@ export const updateApp = grnState => {
...
         resetValues();
+        resetForce();
         $(UNDO_SLIDER_RESET_BUTTON).prop("disabled", false);
         $(UNDO_SLIDER_RESET_MENU).parent().removeClass("disabled");
         updateChargeSliderValues();
         updateLinkDistanceSliderValues();
     }
 
-    if (grnState.undoResetTriggered === false) {
+    if (grnState.undoResetTriggered === false && grnState.simulation !== undefined) {
```
dondi: Why can't `grnState.undoResetTriggered === false` be just `!grnState.undoResetTriggered`?

mihirsamdarshi:
@dondi, The original bug that was pointed out was that after a loaded graph went into Grid Layout, it was unable to revert to a Force Graph Layout. After doing some console logs, I figured out that somehow the code that controls the button for Force Graph or Grid Layout gets called multiple times; during this time, somehow the action seems to get changed back from the force graph setting to the grid layout setting. However my problem is that I don't know how to get to the root of where the button's code is getting called multiple time.

dondi:
@mihirsamdarshi What I would do here is to console log it meticulously. Even with breakpoints you can get lost. I would use strategically-chosen console logs to gradually build a full picture of the sequence of execution. Once you know the full sequence of what triggers what (and so on), then you can unravel and simplify things. Unfortunately there is no shortcut beyond running the sequence with those console log breadcrumbs (sometimes repeatedly) in order to truly understand what is going on.

mihirsamdarshi:
After 2 months, I am ready to merge my Pull Request

mihirsamdarshi:
Pull Request Checklist
- [ ] Print works as expected

johnllopez616:
Is it okay to merge this? Or should I wait for @dondi to review?

Issue #683, opened 2018-10-03
Begin to write xlsx export function for GRNsight


kdahlquist:
Once @jtorre39 wraps up his work with GRNmap, it will be time to switch over to writing the new functionality for GRNsight.  The first task will be to take the existing data stored by GRNsight's json format and export it into .xlsx format (for GRNmap).  I believe when @eileenchoe wrote the new data structure to capture expression data, she actually captured all information from the GRNmap input and output workbooks.  So, @jtorre39 can look to see if that is true.  If so, then he should be able to input a GRNmap input or output workbook into GRNsight and then export it back to xlsx and get an identical file.

kdahlquist:
@jtorre39 has finished his work with GRNmap as the release of v1.10 is immanent.  He is now starting to investigate this issue.

jtorre39:
@dondi question came up, is there an expected behavior for producing the `optimization_parameters` sheet for graphml or sif imported files? 

From what I've seen, if it's a spreadsheet we're importing we do store the it in a meta property within the network object. But the sif and graphml importers don't seem to handle any logic for populating something I can use to export to the `optimization_parameters` sheet.

jtorre39:
After examining the network object when importing GraphML and SIF, the meta property which normally populates the `optimization_parameters` sheet is empty. Unless I'm missing something here, we might have to come up with a dialogue box to populate the sheet or come up with default settings to put into the sheet.

kdahlquist:
You are right, sif and graphml imports are only going to have the network, not any of the other sheets. Later, if when we have populated our expression data database, we will get fancy and have a dialog to have the user populate those sheets. In the meantime, we can give a message saying that they are only going to get the network or network optimized weights sheet.

Also, unrelated, you should meet with @mihirsamdarshi to do a code walkthrough.

Kam D. Dahlquist, Ph.D.
Professor of Biology
Loyola Marymount University

-------- Original message --------

kdahlquist:
@jtorre39 was working on GRNmap last week, but has also figured out the data structure for GRNsight.  There are some more bugs with GRNmap that need to be tracked down, so this will be on hold a little until those are addressed.

jtorre39:
Spent most of the work session debugging the GRNmap issues that came up. Work on those are mostly done. Will spend a little under an hour sometime before the upcoming meeting working with @jiajialingling so we can begin writing the tests for export.

jtorre39:
[...]

PR #707, opened 2018-12-02
Xlsx Export

jtorre39:
#683

This PR includes implementation and tests for being able to export an XLSX sheet. This assumes the user is going from importing an xlsx to exporting an xlsx as going from SIF or other file formats requires a lot more integrated work to fill in missing data (such as optimization_parameters). 

This change is purely back-end.


.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
```
dondi: Strictly speaking, .json files aren't allowed to have comments. That's why GitHub is marking these in red.

https://stackoverflow.com/questions/244777/can-comments-be-used-in-json

May not be a big deal because this looks like a VS Code file, but just wanted to note the standard.

server/controllers/exporters/xlsx.js
```diff
@@ -0,0 +1,107 @@
+const xlsx = require("node-xlsx");
+
+const parseGeneNameArray = function (genes) {
```
dondi: Just a naming note; the `parse` functions here aren't really parsing—they're actually the opposite of parsing. These are more like "converting" or "building" parts of the sheet.

It's all just naming, but admittedly I had to double-take a bit when I saw that there was "parsing" going on until I realized that it actually pertained to converting/building/creating parts of an xlsx.

dondi: Unless I'm misunderstanding/misinterpreting the intent of these functions?

jtorre39: Nono that's fair! Was misconstruing the idea of parsing. Will either use "convert" or "build" for the names of these functions.

.vscode/launch.json
jtorre39: Is the JSON file ok to keep? Or should I remove it from the PR?

server/controllers/exporters/xlsx.js
jtorre39: Updated!

.vscode/launch.json
dondi: Yeaaahhh you know, this looks more like a local configuration that was specific to your workflow. So technically not part of the code base, but an individual developer's specific settings, which might vary.

Yes, let's remove this, and add `.vscode` to _.gitignore_ so that people are free to configure VS Code the way they choose.

jtorre39: Done!

PR #739, opened 2019-02-26
Node Coloring MVC move

mihirsamdarshi:
Also contains fix for locking sliders not working

### DIFF
--- a/web-client/public/js/update-app.js
+++ b/web-client/public/js/update-app.js
@@ -348,7 +348,7 @@ export const updateApp = grnState => {
         updatetoGridLayout();
     }
 
-// Node Coloring
+/* Node Coloring
 
     if (grnState.nodeColoring.avgTopDataset) {
         $(AVG_REPLICATE_VALS_TOP_MENU + " span").addClass("glyphicon-ok");
@@ -363,4 +363,7 @@ export const updateApp = grnState => {
     }
 
     refreshApp();
+
+*/
+
 };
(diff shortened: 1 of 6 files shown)


mihirsamdarshi:
this PR contains code clean up, the fix for the lock sliders option, and beginning stages of a node coloring move which has been disabled/commented out

mihirsamdarshi:
ready for review @dondi

web-client/public/js/update-app.js
```diff
@@ -71,7 +77,7 @@ const displayNetwork = (network, name) => {
     $("input[type='range']").off("input"); // I have no idea why I do this. Investigate later.
 
     // If more things need to be turned off, we'll add them to this array
-    [ "#resetSliders", "#resetSlidersMenu", "#undoReset", "#undoResetMenu" ].forEach(
+    [ "#resetSliders", RESET_SLIDERS_MENU_OPTION, "#undoReset", UNDO_SLIDER_RESET_MENU ].forEach(
```
dondi: The conversion to constants can be useful but is half-done here. May as well convert `#resetSliders` and `#undoReset` here too.

web-client/public/js/update-app.js
```diff
@@ -345,5 +348,22 @@ export const updateApp = grnState => {
...
+        $(AVG_REPLICATE_VALS_TOP_MENU + " span").removeClass("glyphicon-ok");
+        $(AVG_REPLICATE_VALS_TOP_MENU).removeProp("checked");
+        $(AVG_REPLICATE_VALS_TOP_SIDEBAR).removeProp("checked");
+        updaters.renderNodeColoring();
+    }
+
     refreshApp();
+
+*/
+
```
dondi: Maybe plan to eliminate comments soon after merging? I know they have some use while in the middle of things, but beyond that, let GitHub do the change-tracking instead.

Issue #708, opened 2018-12-04
Normalization factor not chosen correctly by GRNsight


kdahlquist:
I've been making figures of DB1-DB6 using GRNsight and I have discovered that there is a bug with the way the edge weight normalization factor is chosen by GRNsight.  

The way it is supposed to work is that the highest magnitude weight is used as the normalization factor.  Since the highest magnitude could be either negative or positive, it should be looking at the absolute value of the weights to determine this.  For example, if the largest magnitude positive weight is "5" and the largest magnitude negative weight is "-6", the normalization factor should be "6".

However, right now, in that situation, GRNsight is choosing "5" instead.

This will need to be fixed as a high priority in the Spring semester.

kdahlquist:
For an example input workbook, try
[15-genes_28-edges_db5-BK-dHAP4-fam_Sigmoid_estimation_output.xlsx](https://github.com/dondi/GRNsight/files/2654491/15-genes_28-edges_db5-BK-dHAP4-fam_Sigmoid_estimation_output.xlsx)

The largest positive weight is 3.224297 and the largest magnitude negative weight is -4.04334.  GRNsight is choosing 3.22... as the normalization factor when it should be choosing -4.04...

mihirsamdarshi:
addressed by PR #749

kdahlquist:
This is not fixed in beta 3.1.17; it might actually be more broke than it was before.

Take a look at the attached file which has just two weights in it's network_optimized_weights sheet.   The positive weight is "1.5" and the negative weight is "-10"  There are two phenomena to fix:
1. The 1.5 weight should be a thin red arrow and the -10 weight should be a thick blue blunt-end edge.  However, the two edges have the same thickness.
2. If you look at the value displayed in the edge weight normalization factor field, it says 1.5.  If I delete the 1.5, then 10 shows up underneath (gray).

[test-normalization-value.xlsx](https://github.com/dondi/GRNsight/files/3230053/test-normalization-value.xlsx)

mihirsamdarshi:
I have diagnosed the problem and it will be fixed in the next PR

kdahlquist:
This one is fixed in beta 3.1.21. Closing.

Issue #561, opened 2018-01-18
Setting the normalization factor changes the zoom factor to 100%


eileenchoe:
Found a bug while testing beta for release #557 (the bug also exists in master). When the normalization factor is applied via the "Set Factor" button, it forces the zoom to go to 100%. @yshin4 and I think the graph should remain at it's current zoom level when the normalization factor is changed.

kdahlquist:
Bug is still there in 3.1.12

mihirsamdarshi:
addressed by PR #749

kdahlquist:
Verified fixed in Beta v3.1.17.

Issue #437, opened 2017-04-12
Grey edges should have only one thickness


kdahlquist:
When investigating #392, I discovered that grey edges can actually have two possible thicknesses, thinnest and one a little thicker than that.  We probably never noticed this before, but it you can see it here:

![check-self-referential-widths-output-v1 18 1](https://cloud.githubusercontent.com/assets/6463867/24980523/801d4596-1f8d-11e7-8229-8267279f390e.jpg)
generated with this file:
[check-self-referential-widths.xlsx](https://github.com/dondi/GRNsight/files/917907/check-self-referential-widths.xlsx)

I think that the grey edges should have only one thickness and that we should review the assignment of thicknesses to edges just in case there are other issues lurking there.

kdahlquist:
This is probably related to #513 and #515.

kdahlquist:
I investigated this as potentially fixed because #513 and #515 are fixed.  However, this is not the case.  I still see the same behavior as reported in a previous comment with the same test file noted above.

Resurrecting part of a comment from #264:

Since there are 12 bins for edge thicknessess, here is what I think the bins should range from for normalized weights (to 4 decimal places, which is what we are rounding to in the display since 12 doesn't go into 1 very nicely) from thinnest to thickest (this is just considering magnitude):
* 0 - 0.0833
* 0.0834 - 0.1666
* 0.1667 - 0.2499
* 0.2500 - 0.3332
* 0.3333 - 0.4165
* 0.4166 - 0.4998
* 0.4999 - 0.5831
* 0.5832 - 0.6664
* 0.6665 - 0.7497
* 0.7498 - 0.8330
* 0.8331 - 0.9163
* 0.9164 - 1.0000

So, in this scheme, all normalized weights between 0 and 0.0500 should be grey and the thinnest bin.  Normalized weights between 0.0501 and 0.0833 are their respective color (magenta or cyan) and the thinnest bin.

However, since we are seeing grey edges with two different thicknesses, something is different about the way this is implemented than I am assuming.  I think we need to:
1. Figure out why this is happening.
2. Make a design decision as to _what_ we actually want to happen.

We now have a slider for setting the gray threshold.  So if the user moves the slider, one could end up with grey edges with multiple thicknesses.  However, I'm just confused why this is happening when the slider is set to the default 0.05.

kdahlquist:
There are 13 thicknesses (2 to 14 pixels) and they should be calculated in this manner:
[...]

PR #749, opened 2019-05-22
Fixing zoom & undo reset

### LINKED ISSUE #245: grey out undo reset when sliders are locked
Quote from issue #234 :
New site behaviors

There's been some new site behaviors added in that I wanted to note, just in case they were undesired.
1. When the user locks the sliders, the reset button also becomes greyed out. Before it was completely usable, but now the user cannot reset the sliders if they are locked. 

> We agree that this is a desired behavior.

However, I retained the ability for the user to undo a reset while the sliders are locked. Should the undo reset button be greyed out as well? It wouldn't be hard to implement.

> We think that the reset button should be greyed out since we are saying "locked" is "locked".
1. I greyed out the "undo reset" settings option when the undo reset button is disabled. I figured that it wasn't greyed out originally due to a bug, but I figured I'd note it anyways.

> Do you mean in the format menu?  We are confused.

mihirsamdarshi:
This PR fixes #245 & #561
Edit:
Added #708 to the list
Added #437 to the list
Added #753 to the list


mihirsamdarshi:
@dondi ready for review

mihirsamdarshi:
Pull Request Checklist
- [ ] Print works as expected

mihirsamdarshi:
Because all comments have been resolved, and all testing has been done, as soon as Travis CI passes, I will merge this PR and refresh beta

web-client/public/js/grnstate.js
```diff
@@ -71,6 +71,10 @@ export const grnState = {
     dashedLine: false,
 
     annotateLinks: () => annotateLinks(currentNetwork),
+
+// Zoom Parameter
+    zoomValue: 4,
```
dondi: @mihirsamdarshi Actually the protocol is you _do_ wait for an approving review before merging a PR even after you have responded to feedback. Because, for example, you might miss something like this. The earlier `4` was replaced with `DEFAULT_ZOOM_VALUE`, but what about this one?

The code reviewer is supposed to explicitly re-approve a PR before it gets merged, precisely because there might be loose ends like the one above.

Issue #687, opened 2018-10-11
Having node coloring and formatting of edge weights based on values disabled, and selecting grid layout causes graph to malfunction.


johnllopez616:
As identified by @mihirsamdarshi and I, if you have the following selections, in this order, it causes the graph to cease existing for weighted graphs:
(1) Node Coloring is disabled
(2) Grid Layout is selected
(3) Disable Formatting of Edge Weights Based on Values

kdahlquist:
Yikes!  Hopefully this can be tackled with the refactor.

kdahlquist:
Checked this on beta 3.1.12, and what happened to demo file 2 is that HMO1 dislodged from the grid and floated around, while the rest of the nodes stayed fixed in place.

mihirsamdarshi:
@dondi I am having a little trouble attacking this one because I am not sure where to start, especially because only one node in particular is being affected

kdahlquist:
Have you tried a non-demo file to see what happens?  Like a "toy" file, for instance?

dondi:
Keep me posted, I can look whenever you need to pass it along.

dondi:
Based on my initial triage, it looks like "Disable Formatting of Edges Based on Weight Values" is sufficient to trigger anomalies. The problem varies depending on the file loaded, but so far, something goes wrong any time that this option is chosen.

Based on this, it looks like a deep dive into just the code for this option is called for next. I’ll report back as I find things out.

dondi:
OK, I think I found the issue. A proposed fix can be found in https://github.com/dondi/GRNsight/pull/761

mihirsamdarshi:
Thanks @dondi, that fix was useful

kdahlquist:
The series of actions reported at the top of this comment thread no longer triggers any problems, so I'm closing this one.

PR #761, opened 2019-06-02
Issue 687 fixes

dondi:
Please see inline commentary in the _Files changed_ tab for some exposition on this proposed fix.

Note also that this branch was taken off of `mihirsamdarshi`. That way, if it looks good, you can merge it into `mihirsamdarshi` and your flow of merging into `beta` wouldn’t be disrupted.

### DIFF
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -485,11 +485,8 @@ export var drawGraph = function (network) {
         .attr("id", function (d) {
             return "path" + d.source.index + "_" + d.target.index;
         }).style("stroke-width", function (d) {
-            if (!grnState.colorOptimal) {
-                return d.strokeWidth = "2";
-            } else {
-                return d.strokeWidth = getEdgeThickness(d);
-            }
+            d.strokeWidth = grnState.colorOptimal ? getEdgeThickness(d) : 2;
+            return d.strokeWidth;
         }).style("stroke-dasharray", function (d) {
             if (unweighted || !grnState.colorOptimal) {
                 return "0";
--- a/web-client/public/js/update-app.js
+++ b/web-client/public/js/update-app.js
@@ -187,14 +187,14 @@ const synchronizeHideAllWeights = () => {
 // Toggle Weighted Functions
 const enableColorOptimal = () => {
     $(BLACK_EDGES).removeClass(ACTIVE_COLOR_OPTION);
-    $(BLACK_EDGES + ">span").removeClass("glyphicon-ok invisible");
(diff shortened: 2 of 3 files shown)


web-client/public/js/graph.js
```diff
@@ -485,11 +485,8 @@ export var drawGraph = function (network) {
         .attr("id", function (d) {
             return "path" + d.source.index + "_" + d.target.index;
         }).style("stroke-width", function (d) {
-            if (!grnState.colorOptimal) {
-                return d.strokeWidth = "2";
-            } else {
-                return d.strokeWidth = getEdgeThickness(d);
-            }
+            d.strokeWidth = grnState.colorOptimal ? getEdgeThickness(d) : 2;
+            return d.strokeWidth;
```
dondi: This was the root of the problem. The code relies on the side effect of assigning the determined `strokeWidth` to the corresponding link object because this `strokeWidth` is accessed later in the code to determine marker choice, edge thickness, distance from a node, etc.—it turns out to affect a lot of other characteristics in the graph.

Although technically, an assignment statement in JavaScript does have a value (that being the right-side value which was assigned to the left side of the `=`), it is generally bad practice to treat an assignment as such. In this case, the assignment statement, when used this way, appears to "pick up" too late—i.e., the rest of the simulation seems to use the prior value of `strokeWidth` in rendering the graph for certain nodes.

Recoding this correctly—i.e., assigning the property first and then returning it—appears to eliminate that race condition. My local version no longer shows the anomalies and errors seen previously when the "Disable Formatting of Edges Based on Weight Value" is chosen. Hope you see the same improvement.

web-client/public/js/update-app.js
```diff
@@ -187,14 +187,14 @@ const synchronizeHideAllWeights = () => {
 // Toggle Weighted Functions
 const enableColorOptimal = () => {
     $(BLACK_EDGES).removeClass(ACTIVE_COLOR_OPTION);
-    $(BLACK_EDGES + ">span").removeClass("glyphicon-ok invisible");
+    $(BLACK_EDGES + ">span").removeClass("glyphicon-ok");
```
dondi: This was a minor tweak, not directly related to #687 but I figured I would tweak it while I was here. Essentially, the `invisible` class is superfluous. If the `glyphicon` class is by itself, without the specific icon identifier, no icon appears anyway. So there’s no need to do anything with `invisible`.

web-client/public/js/graph.js
mihirsamdarshi: understood, I completely overlooked this nuance of Javascript

Issue #777, opened 2019-06-07
Node coloring data options not cleared when loading a new file without expression data


kdahlquist:
If I load demo 2 that has expression data, then I load demo 3 or 4 that don't have expression data, the sidebar menu is still displayed with the options for demo 2, i.e., it doesn't get cleared out when the new graph is loaded.  I can select the options (check/uncheck average box, etc.). Nothing happens to the graph, of course, but this stuff should go away when the next graph is loaded.

kdahlquist:
Actually, this is a problem no matter what workbook you add next. If you load a workbook with expression data second, it is still showing the options for the previous one.

dondi:
Sorry, I was unable to get to this one tonight. Maybe this can be one last thing for @mihirsamdarshi to address? This appears to be a matter of implementing more _update-app.js_ calls that update the state of the user interface based on the selected graph layout. The `updatetoForceGraph` and `updatetoGridLayout` functions appear to be the right places for making these adjustments.

kdahlquist:
@mihirsamdarshi is going to work on this for the next hour to see if he can fix it today before the release.  If it can't be fixed by early afternoon (1:00 or 2:00), then we will save it for later.

mihirsamdarshi:
This has been fixed in my branch

kdahlquist:
Did beta really get refreshed?  If so, you forgot to bump the version number on the website.  Assuming what is up is really 3.1.24, I'm still seeing the bug.

mihirsamdarshi:
yes, I forgot to push gh-pages

kdahlquist:
OK, I reloaded, but I'm still seeing the bug.

mihirsamdarshi:
Really? That is strange it is working for me on the website.

dondi:
I can confirm that I can see the fix; @kdahlquist try clearing your browser cache or try out in Incognito/Private Mode first. The "regular" browser should eventually update too.

However, the http/https `server/ga?path=_____` error still gets blocked, it seems.

kdahlquist:
OK, I completely closed and restarted the browser.  Let me be more specific as to what I see.

Going from demo 2 to demo 3 is fine.

However, going from demo 2 to demo 1 is not fine.  Demo 1 has _some_ expression data, but not all of what demo 2 has; demo 1 does not have the optimized_expression sheets.  However, the optimized expression choices are still in the drop down menus and when I select one of them, I get this:

[...]

PR #784, opened 2019-06-07
fix-ish for #777

mihirsamdarshi:


web-client/public/js/update-app.js
```diff
@@ -309,7 +309,14 @@ const hasExpressionData = (sheets) => {
 
     // renderNodeColoring: function () { }, // defined in graph.js
 
+const clearDropdownMenus = () => {
+    $(TOP_DATASET_SELECTION_SIDEBAR).html("");
+    $(BOTTOM_DATASET_SELECTION_SIDEBAR).html("");
```
dondi: The original version used `empty` instead of `html("")` —does that make a difference?

mihirsamdarshi: yes, it seems that empty was not replacing the HTML, and was fairly slow. This is much faster, and guarantees that the menu will clear.

Issue #59, opened 2014-06-12
Export vector image to pdf and/or svg in a format that can be read by Adobe Illustrator


kdahlquist:
We want the ability to export the graph image to a file in three formats:  jpg, pdf, html.  For the html version, it would be something similar to the way GenMAPP does it.

kdahlquist:
As per the reviewer comment on #271 we will 
* explore exporting the GRNsight graph to svg (which it is natively anyway)
* to cytoscape format (already have a converter that goes the other direction)
* print to pdf
* and if there are libraries for converting svg to other graphics formats

kdahlquist:
Reviewer 1 comment (#278):

"As it does not accept a standard input file type, the output of any other network analysis package requires conversion in to the matrix format required here. Similarly, the tool provides no export function (the option in the File menu remained stubbornly greyed out) and so I can’t take a network from GRNsight and utilise it elsewhere. I also can’t use GRNsight to convert the GRNmap format to something I might like to use elsewhere.

The authors refer to future features coming in version 2 (lines 323-329). I encourage them to consider implementing at least one standard filetype for displaying graph data within their tool. Be it sif, graphml or even gml, it would significantly increase the utility of the tool as it currently exists."

We need to look into import/export of adjacency matrix in sif, graphml, or gml format.

mihirsamdarshi:
#791 should address this

kdahlquist:
I checked this out on beta 4.0.2 and have the following observations:

* I saw the same thing as @dondi, that the file was saved twice.  I have my browser set to ask me where to save things so when I saved it the first time, that dialog opened again (or was underneath the first one).
* The suggested filename has the xlsx extension on it in front of the .png, so it reads "filename.xlsx.png".  Please strip the xlsx off of the suggested filename.
* When I saved the image, the font for the node labels was changed in the png from the sans-serif font we use on the website to a serif font (like Times, but I'm not sure what font it was) in the saved image.
* I actually like having a separate menu item for Export Image.  Eventually it will have different filetypes there, too.  This makes it clear that you are _just_ getting the image, not the data.  The only thing I can think of would be if someone wanted to export the image and data at the same time and create two files, but I think this representation is easier to understand to separate the two functions.
* However, it does make me think that maybe we need to move the "Import" functions into the "Open" menu item as we have talked about before.  
[...]

PR #791, opened 2019-07-31
Add save graph as PNG to GRNsight

mihirsamdarshi:
Added solution for #59 to GRNsight. @dondi wanted to see if it was ok to use this library and if the implementation was ok.


dondi:
I've gone ahead and merged this into `beta` so it can be tried out on the next release. However the comments from the review still stand and ideally they should be addressed before we release as final.

I can move the menu item to “Export Data” pretty easily, but the other changes will take deeper investigation (and confirmation first to see if it happens on other setups).

PR #796, opened 2019-09-04
updates to export image code

mihirsamdarshi:


mihirsamdarshi:
@dondi a couple questions. The first that I have is to resolve the fact that the save svg to png package returns two promises, both PNG files. Is there any way that I can create a function that waits until both have been returned, to determine which one to save? And the second is that my save svg to svg function doesn't seem to save a file. Can you help me figure out why that is happening? When I click on the link to the blob generated, it says "Cannot GET _link to file_"

dondi:
@mihirsamdarshi The first one is quick so I’ll answer that here: you want [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) for that.

For the second question I’ll need to look at the code…if I don’t get to it before the meeting, we can take a look at that time.

web-client/public/js/graph.js
```diff
@@ -654,27 +653,27 @@ export var drawGraph = function (network) {
...
+                return d.value.toPrecision(4);
+            })
+            .each(function (d) {
+                d.weightElement = d3.select(this);
+            });
 
     }
 
-  /* Big thanks to the following for the smart edges
+    /* Big thanks to the following for the smart edges
```
dondi: I see a lot of re-indenting in the latest changes to this PR and most of them look right—but this one doesn’t. Note where the asterisks are in the succeeding lines.
```suggestion
  /* Big thanks to the following for the smart edges
```

web-client/public/js/graph.js
```diff
@@ -1332,7 +1333,7 @@ export var drawGraph = function (network) {
                 return "translate(" + d.x + "," + d.y + ")";
             });
 
-      /* Allows for looping edges.
+            /* Allows for looping edges.
```
dondi: The reindenting on this comment looks off as well.
```suggestion
      /* Allows for looping edges.
```

web-client/public/js/setup-handlers.js
```diff
@@ -75,9 +91,29 @@ export const setupHandlers = grnState => {
         updateApp(grnState);
     };
 
-    $(GRID_LAYOUT_BUTTON).click(() => setGraphLayout(grnState.graphLayout === FORCE_GRAPH ? GRID_LAYOUT : FORCE_GRAPH));
-    $(FORCE_GRAPH_CLASS).click(() => setGraphLayout(FORCE_GRAPH));
-    $(GRID_LAYOUT_CLASS).click(() => setGraphLayout(GRID_LAYOUT));
+// Grid and Force Graph Layout
```
dondi: The indentation on _this_ comment looks inconsistent with how the other comments have been re-indented. Pick a convention and stay with it.
```suggestion
    // Grid and Force Graph Layout
```

mihirsamdarshi:
ready to be merged @dondi

dondi:
@mihirsamdarshi There’s a failing test on this branch. Please fix it before we merge. (see Travis report for details)

web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -51,6 +58,8 @@ const uploadHandler = (uploadRoute, uploader) => {
 };
 
 const networkErrorDisplayer = xhr => {
+    // re-enable upload button
+    disableUpload(false);
```
dondi: Small naming note here: see how `disableUpload` as a function name doesn't line up with its parameter of true or false. ("disable upload to true" doesn't make a lot of plain English sense)

Instead, you can do this:
- `enableUpload()`
- `disableUpload()`
- `setUploadState(state)`

Issue #740, opened 2019-02-26
Manipulating graph manually during grid layout should change force graph layout button back to grid layout button


kdahlquist:
Manipulating graph manually during grid layout should change force graph layout button back to grid layout button. Moved over here from issue #600.

Unless we do this, once someone moves a node around there is no way to get back to the grid unless reloading or going through force graph layout options.

kdahlquist:
Two buttons, "Force Graph" and "Grid".  Always active. 

Idempotent.

dondi:
When both buttons are active, the user can click either button anytime and just put the graph in that mode, unconditionally. This will accommodate the ability to "reset" any manually-dragged nodes. So clicking "Force Graph" will "unlock" all manually-dragged nodes so that they go back to the physics engine. Similarly, clicking "Grid" will "resnap" all manually-dragged nodes so that they go back to the grid.

mihirsamdarshi:
PR #802 has a fix in

kdahlquist:
Confirmed in beta 4.0.5

Issue #800, opened 2019-09-25
Font is changed in beta 4.0.4


kdahlquist:
The node font is now a serif font in beta 4.0.4.  Can we change it back to what it was before?

kdahlquist:
![image](https://user-images.githubusercontent.com/6463867/65647365-533a9900-dfb3-11e9-8800-d4dc7d616db6.png)

mihirsamdarshi:
Added fix for this, will be issued in next PR

kdahlquist:
confirmed fixed in beta 4.0.5

PR #802, opened 2019-09-26
add bug fixes

mihirsamdarshi:
this PR fixes issues #740, #800 and more


web-client/public/js/setup-handlers.js
```diff
@@ -92,31 +93,29 @@ export const setupHandlers = grnState => {
...
+        updaters.setNodesToGrid();
+        grnState.graphLayout = GRID_LAYOUT;
+        updateApp(grnState);
     });
 
     $(GRID_LAYOUT_CLASS).click(() =>  {
-        setGraphLayout(GRID_LAYOUT);
+        updaters.setNodesToGrid();
+        grnState.graphLayout = GRID_LAYOUT;
+        updateApp(grnState);
```
dondi: Looks like these click handlers can be consolidated—what’s here right now is duplication of two functions into four. Does putting both selectors in the same jQuery call work? For example:
```javascript
$(FORCE_GRAPH_BUTTON, FORCE_GRAPH_CLASS).click( // .....etc.
```
This way, the functions for setting nodes to force graph vs. grid aren’t duplicated.

(if a comma-separated list doesn’t work, I’m sure we can come up with a way to make both selectors work together)

mihirsamdarshi: Multiple selectors will work, but the format we have our constants in does not work. This selector requires ` $("#forceGraph, #forceGraphButton").click(() => { //...` where multiple selectors are contained inside the same string. This: `$(FORCE_GRAPH_BUTTON, FORCE_GRAPH_CLASS).click( // ...` will not work, because it is passing multiple separate strings. Is there a way to clean this up?

web-client/public/js/update-app.js
```diff
@@ -486,6 +485,14 @@ export const updateApp = grnState => {
         updaters.removeNodeColoring();
     }
 
+    if (grnState.network !== null &&  grnState.network.sheetType === "weighted") {
+        $(EXPORT_TO_WEIGHTED_SIF).parent().removeClass("startDisabled").removeClass("disabled");
+        $(EXPORT_TO_WEIGHTED_GML).parent().removeClass("startDisabled").removeClass("disabled");
+    } else if (grnState.network !== null &&  grnState.network.sheetType === "unweighted") {
+        $(EXPORT_TO_WEIGHTED_SIF).parent().removeClass("startDisabled").addClass("disabled");
+        $(EXPORT_TO_WEIGHTED_GML).parent().removeClass("startDisabled").addClass("disabled");
+    }
```
dondi: This block doesn't do anything when `grnState.network === null`.....seems like we should accommodate that case?

web-client/views/upload.jade
```diff
@@ -67,20 +64,15 @@ html
...
                       a(href='#' id='exportAsUnweightedSif') To Unweighted SIF
-                    li(class='startDisabled disabled weighted export')
+                    li(class='startDisabled disabled')
                       a(href='#' id='exportAsWeightedSif') To Weighted SIF
                     li(class='divider')
-                    li(class='startDisabled disabled unweighted export')
+                    li(class='startDisabled disabled')
                       a(href='#' id='exportAsUnweightedGraphMl') To Unweighted GraphML
-                    li(class='startDisabled disabled weighted export')
+                    li(class='startDisabled disabled')
```
dondi: These classes can actually be useful in the future (maybe even now) for easily picking out the commands that have to do with export or with weightedness or both.

PR #844, opened 2020-04-12
All changes to design as proposed by Dr. Dahlquist

Kevboe:


dondi:
@Kevboe based on the last commit, it looks like you’re working to clean up Travis errors. Looks like you just have a few left—we're standardized on double quotes for the GRNsight code base and not single quotes. I anticipate that these will be easy to fix, if you haven't already fixed them but just haven't pushed yet.

You can use `npm run lint` locally to "preflight" the errors so that you don't have to commit and push repeatedly.

web-client/public/js/update-app.js
```diff
@@ -447,6 +482,10 @@ const updateBottomDataset = () => {
     updaters.renderNodeColoring();
 };
 
+if (grnState.genePageData.identified === false) {
```
dondi, line 485: We typically don’t directly compare to `true` or `false`—instead, best practice is to try to let the expression itself sound like a `true` or `false` value:
```suggestion
if (!grnState.genePageData.identified) {
```
Note how that sounds like “if not identified” (presumption of course is that the reader knows that `!` means “not”—which they should, if they know the language).

web-client/public/js/setup-handlers.js
```diff
@@ -423,6 +429,48 @@ export const setupHandlers = grnState => {
...
+        updateApp(grnState);
+    });
+
+    $(SPECIES_BUTTON_FLY).click(function () {
+        var selection = $(SPECIES_BUTTON_FLY).attr("id");
+        grnState.genePageData.identified = false;
+        identifySpeciesMenu(selection);
+        updateApp(grnState);
+    });
+
```
dondi, line 473: These functions all have the same general structure, varying only by certain values—ripe for a parameterized helper function. But this consolidation can happen later.

web-client/public/js/update-app.js
```diff
@@ -322,6 +328,34 @@ const hasExpressionData = (sheets) => {
...
+    if ($(SPECIES_DISPLAY).val() === "Homo_sapiens") {
+        $(SPECIES_BUTTON_HUMAN + " span").addClass("glyphicon-ok");
+    }
+    if ($(SPECIES_DISPLAY).val() === "Mus_musculus") {
+        $(SPECIES_BUTTON_MOUSE + " span").addClass("glyphicon-ok");
+    }
+    if ($(SPECIES_DISPLAY).val() === "Saccharomyces_cerevisiae") {
+        $(SPECIES_BUTTON_YEAST + " span").addClass("glyphicon-ok");
+    }
+};
```
dondi, line 357: Some consolidation potential here also, but not quite as pronounced as the other one. Again, it can wait.

Kevboe:
Thanks for the tips, definitely will look to make these changes in the future. I'm sure we can talk about consolidation of functions, some of which are clunky, but I couldn't think of a better way to do it.

PR #845, opened 2020-04-12
Node coloring with expression database works

afiller1:
I have the baseline for making it so that the user can color their GRN using data from our expression database. Currently, I have it working so that, when a user uploads a workbook with no expression data, the nodes will color based on the Barreto data. Currently, it doesn't take in user input as to which data set to use, that's hard coded for now. I was mostly focusing on getting everything connected and inserting the data from the database into the whole network correctly and in the correct JSON format. I mostly tried it out with a file I uploaded with only network data (something funky was happening with demo 3... I will discuss this in the meeting). But node coloring happens once the query is made to the database. I also adjusted the UI slightly, but that is also up for change suggestions. Overall, the baseline functionality of this feature now works to the degree that I've tried it out. I've merged with beta, so there should be no conflicts remaining.


kdahlquist:
I just want to chime in that demos 3 and 4 are using a legacy format for GRNmap, i.e., they are there because they were published in the 2015 Bulletin of Mathematical Biology paper on an alpha version of GRNmap.  After that was published we changed the format of the input file, but kept the demos there because they are tied to the published paper.  I don't know if that helps or not.

dondi:
@afiller1 That last linting error, AFAICT, is due to `async`/`await`—looks like our tech stack isn’t  up for processing that yet. Getting it to do that will take some package updates and broad testing, so the quicker change for now is to switch to `then`/`catch` style for the async code.

web-client/public/js/setup-handlers.js
```diff
@@ -216,11 +217,15 @@ export const setupHandlers = grnState => {
 // Node Coloring
     $(NODE_COLORING_TOGGLE_CLASS).click(() => {
         grnState.nodeColoring.nodeColoringEnabled = !grnState.nodeColoring.nodeColoringEnabled;
+        console.log("nodecoloringenabled: " + grnState.nodeColoring.nodeColoringEnabled);
+        // $(NODE_COLORING_TOGGLE_SIDEBAR).prop("checked", !$(NODE_COLORING_TOGGLE_SIDEBAR).prop("checked"));
+        console.log("node coloring checked:" + $(NODE_COLORING_TOGGLE_SIDEBAR).prop("checked"));
```
dondi, line 222: We'll want to take these logs out once the code has stabilized.

web-client/public/js/update-app.js
```diff
@@ -312,13 +312,14 @@ const shortenExpressionSheetName = (name) => {
 };
 
 const hasExpressionData = (sheets) => {
+    // grnState.nodeColoring.showMenu = true;
```
dondi, line 315: Same with commented-out code. That's why we have version control 😊

web-client/public/js/update-app.js
```diff
@@ -525,21 +538,69 @@ export const updateApp = grnState => {
         updatetoGridLayout();
     }
 
+
 // Node Coloring
     if (grnState.network !== null && grnState.nodeColoring.nodeColoringEnabled
-      && hasExpressionData(grnState.network.expression)) {
+    && hasExpressionData(grnState.network.expression)) {
```
dondi, line 544: Looks like an unintentional indentation-only change here?

web-client/views/upload.jade
```diff
@@ -382,6 +391,43 @@ html
...
+                    //- Refresh selection upon new upload
+                    //- Add "onclick" functionality for checkbox
+                    option(value="data-set-select-barreto") Barreto 2018 wt
+                    option(value="data-set-select-dahlquistdcin5") Dahlquist 2018 dcin5
+                    option(value="data-set-select-dahlquistdgln3") Dahlquist 2018 dgln3
+                    option(value="data-set-select-dahlquistdhap4") Dahlquist 2018 dhap4
+                    option(value="data-set-select-dahlquistdzap1") Dahlquist 2018 dzap1
+                    option(value="data-set-select-dahlquistwt") Dahlquist 2018 wt
+                    option(value="data-set-select-kitagawa") Kitagawa 2002 wt
+                    option(value="data-set-select-thorsen") Thorsen 2007 wt
```
dondi, line 414: Next step: This part should eventually become dynamic. Instead of being hardcoded in _upload.jade_, this should start out empty, and a web service call to the server should exist which returns this list.

web-client/views/upload.jade
```diff
@@ -382,6 +391,43 @@ html
...
+                select(multiple id="timepoints-select")
+                    option(value="timepoint-select-10") 10
+                    option(value="timepoint-select-15") 15
+                    option(value="timepoint-select-20") 20
+                    option(value="timepoint-select-30") 30
+                    option(value="timepoint-select-40") 40
+                    option(value="timepoint-select-60") 60
+                    option(value="timepoint-select-90") 90
+                    option(value="timepoint-select-120") 120
+                    option(value="timepoint-select-1080") 1080
```
dondi, line 425: By extension, this will change also. Again it will start empty, but when a data set is chosen, a second web service call will return the available time points for the chosen data set.

afiller1:
Node coloring now updates based on the user selection! Need to add a timepoints selection option still, and figure out the logistics of using the db to do node coloring on a workbook that already has expression data.

afiller1:
Only linter errors are "Identifier is not in camel case" errors, but the identifiers in the object where that's a problem NEED not to be in camel case. They are the names of the datasets, which are traditionally structured with underscores.

PR #847, opened 2020-04-20
Kev species display

Kevboe:
Small changes as asked by Dr. Dahlquist

### DIFF
--- a/web-client/public/gene/api.js
+++ b/web-client/public/gene/api.js
@@ -361,16 +361,15 @@ let parseJaspar = function (data) {
                ) {
                    const errorString1 = "No gene information was retrieved for " + symbol.symbol + ".";
 
-                  
-                   const errorString2 = "This could have happened because:" 
-                   const errorString3 = "You can check back later to see if gene information"
-                   + " can be retrieved or submit an issue to https://github.com/dondi/GRNsight.";
+                   const errorString2 = "This could have happened because:";
+                   const errorString3 = "You can check back later to see if gene information" +
+                   " can be retrieved or submit an issue to https://github.com/dondi/GRNsight.";
 
                    $("#error2").text(errorString2);
-                   var errorString4 = $('<ul/>').appendTo('#error2');
+                   var errorString4 = $("<ul/>").appendTo("#error2");
                    errorString4.append("<li>The wrong species is selected </li>");
                    errorString4.append("<li>GRNsight could not access the gene information"
-                   +" from one of the source databases</li>");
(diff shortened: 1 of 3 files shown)


web-client/public/gene/api.js
```diff
@@ -363,15 +361,18 @@ let parseJaspar = function (data) {
                ) {
                    const errorString1 = "No gene information was retrieved for " + symbol.symbol + ".";
 
-                   const errorString2 = "This could have happened because either"
-                    + " GRNsight could not access the gene information from one of the source databases"
-                    + " or because no information exists for the gene in the source databases.";
+                   const errorString2 = "This could have happened because:";
+                   const errorString3 = "You can check back later to see if gene information" +
+                   " can be retrieved or submit an issue to https://github.com/dondi/GRNsight.";
```
dondi, line 366: I would indent continued lines one level deeper to emphasize that they are continuations and not new statements (I wonder if I can make the linter flag those…), but that’s fine, no need to hold up the pull request over that.

web-client/public/gene/api.js
```diff
@@ -363,15 +361,18 @@ let parseJaspar = function (data) {
...
 
-                   const errorString3 = "You can check back later to see if gene information"
-                    + " can be retrieved or submit an issue to https://github.com/dondi/GRNsight.";
+                   $("#error2").text(errorString2);
+                   var errorString4 = $("<ul/>").appendTo("#error2");
+                   errorString4.append("<li>The wrong species is selected </li>");
+                   errorString4.append("<li>GRNsight could not access the gene information"
+                   + " from one of the source databases</li>");
```
dondi, line 372: This too.

web-client/public/gene/api.js
```diff
@@ -363,15 +361,18 @@ let parseJaspar = function (data) {
...
+                   errorString4.append("<li>No information exists for the gene in the source databases.</li>");
 
                    $("#error1").text(errorString1);
-                   $("#error2").text(errorString2);
                    $("#error3").text(errorString3);
```
dondi, line 376: Can we find a way to name these identifiers more semantically? Numbers in variable names are very difficult to maintain. I realize this may not have been your code originally but you’re in charge now and if you encounter “code smells” like this, you are empowered to address them.

So notice right here, this change deletes the line pertaining to `error2`. So now what, we have `error1` going to `error3`? Future programmers will look at that and go 🤷 This is an example of the maintenance issues one may fall into when using numbered identifiers.

Issue #832, opened 2020-01-27
Gray Edge Threshold set to 100% should make all edges gray


kdahlquist:
When I slide the Gray Edge Threshold all the way to the right and it says 100% using demo file 2, all the edges turn gray except for one that is still red.  I think they should all be gray.

igreen1:
Fixed in pull request from igreen1 branch #876 Awaiting review

(link since I'm not entirely sure the number is correct!)
https://github.com/dondi/GRNsight/pull/876#issue-486977387

igreen1:
Issue resolved in beta 4.0.19, pr #876 by merging igreen1 into beta branch :)

kdahlquist:
Confirmed fixed in Beta 4.0.19

PR #876, opened 2020-09-15
Fixed bug #832

igreen1:
Fixes issue #832 - Gray Edge Threshold set to 100% should make all edges gray. 

This bug occurred when the edge in question had a weight with precision higher than 3 decimal places due to the rounding of maxWeight in the code. Solved by rounding the comparison value before comparing to grayThreshold.

Changes noted in commit messages:
1. Changed line 484 to use 'normalize' function instead of performing operation inline. This format follows the convention elsewhere in code (not an error; just styling)
2. Changed all lines comparing normalize(d) to grayThreshold to round the result of normalize. 
    "normalize(d).toPrecision(4) <= grayThreshold"

### DIFF
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -465,7 +465,7 @@ export var drawGraph = function (network) {
         }).style("stroke", function (d) {
             if (unweighted || !grnState.colorOptimal) {
                 return "black";
-            } else if (normalize(d).toPrecision(4) <= grayThreshold) {
+            } else if (normalize(d) <= grayThreshold) {
                 return "gray";
             } else {
                 return d.stroke;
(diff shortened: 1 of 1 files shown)


web-client/public/js/graph.js
```diff
@@ -481,7 +481,7 @@ export var drawGraph = function (network) {
             var xOffsets;
             var color;
 
-            if (Math.abs(d.value / maxWeight) <= grayThreshold) {
+            if (normalize(d).toPrecision(4) <= grayThreshold) {
```
dondi, line 484: Does _every_ call to `normalize` now follow with `toPrecision(4)`? Because if so, then we can integrate the `toPrecision(4)` conversion itself into the `normalize` function, thus leaving all of these lines as just `normalize(d)` once more, with `toPrecision(4)` already built-in.

dondi (commented):
Generally in the right direction, with one consolidation opportunity if `toPrecision(4)` is now consistently applied to all `normalize` calls.

igreen1:
Agreed :) Edited to move .toPrecision(4) call to within the function definition

dondi (approved):
Nice—note how much smaller the delta is…to enact the same fix!

PR #880, opened 2020-09-29
Igreen1

### LINKED ISSUE #444: Weights should not be drawn outside the bounding box when the bounding box/viewport is fixed
When a weighted graph is loaded and the "always show weights" option is selected, some of the weights are drawn outside the viewport. While this isn't so much an issue with the ability to scroll around, I figured I'd point it out in case this is something we wanted to change. I assume someone choosing fixed may not want to need to scroll outside the viewport.

screen shot 2017-04-18 at 7 21 54 pm

You can see in this screenshot that is seems that GLN3 doesn't have a weight. The weight is actually drawn right outside the viewport, and can be seen if I scroll down.
screen shot 2017-04-18 at 7 22 43 pm

With non-looping edges, if the two nodes are at a great enough distance and the edge is up against the top or bottom of the bounding box, the curving of the edge will cause the weight to be drawn out of bounds. Self-referential edges will always draw out of bounds if right up against the bottom of the viewport.
Here you can see these overlapping weights drawn out of bounds on the top:
screen shot 2017-04-18 at 7 23 34 pm
And here on the bottom:
screen shot 2017-04-18 at 7 24 51 pm

I'm not sure if this is something we would want to fix or not, but I hadn't noticed it before and I'm not sure if this behavior is alright or not.

### LINKED ISSUE #700: Nodes go slightly over edge when "Restrict Graph to Viewport" and "Fit to Window" is selected
While testing for PR #692, I noticed that, as you can see in the image below, when the "Restrict Graph to Viewport" and "Fit to Window" options are selected, and a node is pushed to the right, it goes slightly past the viewport.

I tested this on beta, and the same results apply.

image

This was discovered on through the 21-genes_31-edges_Schade-data_input.graphml file.

igreen1:
Fixes #700 and fixes #444 and makes two minor corrections in the draw self-referential edge function

* #700 was fixed by changing the calculation of RightBoundary to include an offset, which is a variable declared to be 5.

* #444 was fixed with two changes. For self-referential edges, the textbox with the weight was bounded within the viewport. For all other edges, the edge path calculation was tweaked such that edges always curve inwards when near the edge (added an offset to the points on the edge line)

The two small corrections are
1. Adding a comment as it appears a calculation has changed in the background since xRotation doesn't affect the self-referential edge drawing. A warning for future editors
2. Fixing an indenting style issue for the return statement

### DIFF
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -1317,7 +1317,7 @@ export var drawGraph = function (network) {
 
                 var selfReferringEdgeWidth = (selfReferringEdge ? getSelfReferringRadius(selfReferringEdge) +
                     selfReferringEdge.strokeWidth + 2 : 0);
-                var rightBoundary = width - (d.textWidth + 5) - BOUNDARY_MARGIN - selfReferringEdgeWidth;
+                var rightBoundary = width - (d.textWidth + OFFSET_VALUE) - BOUNDARY_MARGIN - selfReferringEdgeWidth;
                 var currentXPos = Math.max(BOUNDARY_MARGIN, Math.min(rightBoundary, d.x));
                 if (adaptive && width < MAX_WIDTH &&
                     (currentXPos === BOUNDARY_MARGIN || currentXPos === rightBoundary)) {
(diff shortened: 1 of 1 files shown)


dondi (approved):
LGTM

PR #881, opened 2020-10-02
Grayed out Undo Reset Button

luluriffic:
finally!!


dondi (approved):
Looks good—I’m curious about the one-directionality of the logic (i.e., if `slidersLocked` is true, then we will disable Undo Reset, but if it’s false, we do nothing). I can see how that is possible—after all, locking the sliders isn't the only factor that determines whether Undo Reset should be doable—so I guess we'll have to see what testing does.

PR #885, opened 2020-10-09
New fix for #444 for non-self-referential edges to preserve the arcs.

igreen1:
A quick fix for #444 by forcing weights to stay in the bounds with an offset of 20.

I will keep this on my radar for a better arc calculation and work on a non-magic number edge offset (something based on the precision of the number), but this will definitely work for now. Keeps weights within the boundary


dondi (approved):
Looks good in principle; will have to see how it plays out on beta.

Issue #886, opened 2020-10-16
GraphML and SIF imports run afoul of recent species feature


dondi:
The species feature that was worked on last year does not carry over to our GraphML and SIF import code. The underlying principle of our files is that the only true requirement is a network—everything else should be “icing on the cake” and so should not produce errors if they are missing.

For the GraphML and SIF importers, the `meta` property which contains the `species` will be missing because there is flat-out no (standard) way to introduce that value in the files. Thus, the species code that was added last year should gracefully skip itself if the `meta` property is missing.

For the default behavior, there is existing code in the workbook importer and we should investigate how to refactor GRNsight so that the existing default logic applies across all file formats.

igreen1:
@igreen1  got the import function working by adding a manual optional (I don't believe optional chaining works in the Node we are running unfortunately). There is definitely a better way but it would require more refactoring. I will be working tonight on that but wanted to give everyone an update well before the meeting. I'd call this a working solution but not an elegant one.

However, now, there are issues with the expression dataset. Graphml doesn't have expression data, so accessing that throws errors. Most, but not all, of the attempts to access the expression property are shielded by 'if exists'. So, @igreen1  need to expand this to all accessing. This is another thing I will be working on tonight ! 

(I had a midterm Wednesday in my hardest class, so only did 2/4 of my hours. I skipped my Monday hours and will do them tonight)

dondi:
@igreen1 has implemented a fix but needs to update the test suite since this is new behavior.

igreen1:
When attempting the build, the test throws the following error

![image (1)](https://user-images.githubusercontent.com/40191185/98403735-ac825200-2026-11eb-8542-44c943e80a75.png)

This is because the best fix I found, following the notation of the .xslx import, requires an addition of .meta and .expression to the sif and graphml imports. However, the test is expecting a deep equality. The next goal, then, is to add this new format to the tests.

Discussed in the meeting today how to implement this fix; should be a simple paste into the expected results of two empty fields.

igreen1:
Officially working on igreen1 commit #6abe1dd 

Fixed the test issue by adding expression and meta objects.

Also converted expression and meta imports to objects instead of arrays to more closely reflect the type of data they store.

PR is waiting for some final bug fixes on the graphML validation.

dondi:
Clarification for @ahmad00m—SIF and GraphML used to report errors when they were loaded because they don't contain species information. This should no longer be an issue.

[...]

Issue #541, opened 2017-10-11
Implement GraphML XSD validation


dondi:
@ebachoura work on #363 has revealed that other errors may get through because they are actually schema validation errors. To avoid too much custom coding, we will explore implementing GraphML XSD validation because the XSD does exist anyway (http://graphml.graphdrawing.org/xmlns/1.1/graphml.xsd).

We can try this out to see whether this provides benefits. It remains an open question whether it is widely followed to begin with.

igreen1:
After digging around for a while trying to understand xsd, it appears it was already implemented.

In the import and export,
`xsi:schemelocation = http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd`

@igreen1 can change this to the 1.1? There doesn't appear to be significant changes, though, and changing the scheme might cause unforseen issues, so probably best to avoid.

dondi:
@igreen1 I think that line identifies _which_ XSD is supposed to be followed but I haven't seen evidence that there is an actual _check_ on whether the XML file follows that format.

Ideally there's an easy-to-use function that performs this check. Then we would write code that intercepts the result of this check and reports an error if needed.

igreen1:
#74c292d theoretically fixes this, but it is untested (see below). 

Most of the time @igreen1 spent this week was on the following:

When attempting to test the solution, a bigger bug was found. I will wait until our meeting to open a issue thread as its possible it is a user error. The files used are created by:
1. Open demo graph
2. Export demo graph as unweighted or weighted graphml (File > Export Data > To graphml)
3. Open the downloaded file (File > Open File > select the file)

On my local server, when these files are added, the server shows `POST /upload-graphml 400` which implies an issue with the upload function. The name does NOT update. Clearly, the upload has entirely failed and GRNSight is continuing like it didn't happen.This happens on BOTH igreen1 branch and beta branch, which indicates the issue is not with the most recent igreen1 commit. Testing on the main branch doesn't make sense since it doesn't have the prerequisite features. Other than failing to upload, no other issues arise and I can continue to use GRNSight without issue.

[...]

PR #896, opened 2020-11-13
Igreen1

igreen1:
Addresses issue #886 by finally fixing the tests that didn't allow the last PR to function properly. The fix adds an expression and a meta property to all import networks so later attempts to access these properties returns the expected undefined instead of an error.

Addresses issue #444 by increasing the margin of the self-referential edges

The fix for #541 was removed from my branch due to a sudden issue with the library. I suspect it is an issue with my local machine and hope to resolve it tonight (after my lab report)


dondi (approved):
In terms of addressing the issue, this looks good to me, although the number of times that `meta`/`expression` had to be manually added makes me wonder whether some consolidation is called for here (e.g., some kind of default “initialize empty network” function or some such.

Perhaps a refactor to do next.

Issue #883, opened 2020-10-02
Test import of GRNmap input workbooks for two-column-formatted sheets


dondi:
Similar treatment as #734 in terms of testing and warning-/error-catching for the following sheets:
- Production Rate
- Degradation Rate
- Optimized Production Rates
- Threshold B
- Optimized Threshold B

Onariaginosa:
Created and implemented tests for the Two Column sheets. These are:
- `idLabelError`: if ID is not in the upper left cell.
- `incorrectColumnHeaderError`: if they submit a sheet with the wrong B column header for their sheet
- `missingColumnHeaderError`: if they submit a sheet with no B column header
- `invalidGeneTypeError`: if they try and submit a Gene name that is not a string, or is undefined (may be split into missing GeneName)
- `invalidGeneLengthError`: if they go over the max Gene name length
- `specialCharacterError`: if they have special characters in their genes
- `invalidValueError`: If they give us a value that is not a number for gene value, or one that is undefined (may be split into missing ValueError)
Unimplemented Tests:
- A test that makes sure that the gene names listed are the same throughout all sheets (I didn't implement it yet because I'm not sure if this should be required)

 However, I made sure that the errors/warnings are currently not being sent currently because of the current structure of the of the data objects. I'm not sure if the two colomn sheet data is being used when rendering the graphs, so I did not send the errors and warnings along with the data yet. Additionally, should warnings and errors be generated per sheet, or should I add them to one congregated warnings/errors list?

Onariaginosa:
This is completed. Pull request pending. I added tests to the mass tester list, and changed the spreadsheet-controller code to include the error and warnings from each of the two column sheets. Decided against creating a test that made sure that gene names listed were consistent throughout the various two column sheets.

dondi:
@ahmad00m has found that this issue appears to be resolved. Some notes may be forthcoming but otherwise can be closed.

ahmad00m:
All the implemented tests have been tested and the errors provided by tests were accurate.

PR #899, opened 2020-12-04
Test import of GRNmap input workbooks for two-column-formatted sheets

Onariaginosa:
Added tests mentioned in issue #883


Onariaginosa:
Accidentally selected master instead of beta

PR #900, opened 2020-12-04
 Test import of GRNmap input workbooks for two-column-formatted sheets

Onariaginosa:
Added tests mentioned in issue #883


test/test.js
```diff
@@ -502,6 +504,117 @@ var betweennessCentrality = function (input, directed, node, centrality) {
 };
 */
 
+
+
+
+
+
```
dondi, line 511: You don’t really need to space things out this much but nbd 😂

test/test.js
```diff
@@ -502,6 +504,117 @@ var betweennessCentrality = function (input, directed, node, centrality) {
...
+    }
+    assert.equal(frequency, additionalSheetExtraneousDataWarningCount);
+};
+
+
+
+
+
+
+
```
dondi, line 616: Hmmm, if you really wanted to separate things this distinctly, maybe a new module file is called for? 🤔

dondi (approved):
Generally looks good, with some notes on blank lines in _test.js_.

Issue #897, opened 2020-11-13
Explore possible refactor to consolidate creation of new networks


dondi:
The fix to #886 revealed that adding empty `meta` properties had to be done in multiple places. It would be good to explore whether this can be consolidated into a single utility function (e.g., `createEmptyNetwork` or other appropriate name) so that future changes to the baseline network object would only have to be done in one place.

kdahlquist:
This is giving me flashbacks to GRNmap... :)

igreen1:
Done!
Added createEmptyNetwork and initNetwork
createEmptyNetwork... creates an empty network
initNetwork creates an empty network then assigns fields based on what's passed in (either fills all the fields, none, or something in between)

Commit #4ba0807 on igreen1 has the fix ready to pull :)

igreen1:
Would love comments on the name
createEmptyNetwork seems verbose to me
initNetwork feels,,, short. I think initialize is the technically correct term for assigning values to variables but it ignores the portion of the code that also creates an empty network. But 'createNetwork' or 'createNetworkwithArgs' felt worse. Will discuss at the meeting if not before

kdahlquist:
createEmptyNetwork sounds OK to me!  initializeNetwork is OK, too.  I don't think it's bad to have names this long if they are clear.  Someone else reading the code later is going to know exactly what you are talking about.

My only question is that is it creating an empty network only or is it creating an empty data structure (like all the data, like expression, etc.) in which case the "network" part of it might be misleading and it should be called createEmptyDataStructure or initializeDataStructure (not sure if the case is right here).

@dondi, thoughts?

dondi:
Agreed, clarity is more important than brevity when it comes to naming. At least, that is the preference these days now that programming languages don't have the same memory constraints that they used to have.

Good point about network vs. data structure…I agree with the latter though data structure is too generic.

I mean technically it is the network still…just a highly annotated one at this point. Does the "network network" (the part consisting solely of nodes and edges) have a reasonable alternate name?

Naming is hard. Might need to think a bit. If we keep "network" for the nodes and edges (which does also make sense because in the workbook these are the "network sheets") we would need a new name for the whole shebang that is more specific than "data structure." 🤔

kdahlquist:
Maybe we call it a "workbook" with a nod to Excel.  That at least implies that it's more than just the network data.

dondi:
“Workbook” did cross my mind. `createEmptyWorkbook` …that can work.

[...]

PR #901, opened 2020-12-06
Igreen1

igreen1:
Addresses the renaming refactor from issues #897 

This changes code references of network, the object containing all the GRN information, to workbook to reflect the expansive nature of the data structure, which contains information far beyond the mere network at this point.

Some references to network, though, were left when they are referencing the network sheet of the xlsx file or the network portion of the workbook (in some tests)

Moreover, the former use of workbook to reference .xlsx files was changed to workbookFile to denote that these merely hold the workbook.


dondi (approved):
LGTM. Great wrap-up to the semester!

Issue #674, opened 2018-09-05
Security vulnerabilities in packages used by GRNsight


mihirsamdarshi:
A new feature in NPM allows us to see packages that have known security vulnerabilities due to being outdated. This occurs during npm install. A few have issues listed as "high" or "critical." I believe that it would be in our best interest to examine these packages and see if we can do anything to solve this, such as updating certain packages.

kdahlquist:
Let's talk about this one at a meeting.

dondi:
I've been meaning to do this…the best time, I figured, would be when beta is solid and just before a master release. That way, if the vulnerability updates break anything, then we'll know that it really is due to the updates.

mihirsamdarshi:
We can always integrate [Dependabot](https://github.com/marketplace/dependabot-preview), which automatically issues PRs that keep our dependencies up to date. That way, we can choose which dependencies to update and which ones to keep, and we are actively alerted to when new features or anything we should be aware of comes out.

dondi:
A new branch, _security-updates_, has been created that updates an many security fixes as possible without overly disrupting the code. Vulnerabilities on this branch, as reported by _npm audit_, are now done to just 8. Unfortunately, these 8 involve packages that we used that have been deprecated: _jade_ is now _pug_, and _istanbul_ now appears to be _nyc_.

I held back on these wholesale library changes to avoid incurring potentially major code changes (especially for _pug_). I might explore these a little later if the current updates pan out.

_security-updates_ is in sync with the latest beta 4.0.29, so for review purposes, GRNsight devs are asked to try this branch out, do an _npm install_, and see if things remain functional @Onariaginosa @igreen1 @luluriffic

If these changes don’t appear to break anything, we can merge these into _beta_ in preparation for the next major release. Exploring a switch to _pug_ and _nyc_ to replace _jade_ and _istanbul_, respectively, might be explored.

dondi:
@dondi and @Onariaginosa tried a pass at these updates but there are some issues with required Node versions between what’s on the EC2 server—however the EC2 server is on the latest LTS and we prefer to stay on LTS. @dondi will gradually traverse library versions to find one that works with the version of node on EC2

dondi:
Although v5.1.x applied security updates that were applicable at a time, most likely this will be an on-going process that will need to be done periodically. If anything, we are not as behind as we used to be before, but over time we may need to do this again

I wonder if security updates should just become a part of the pre-release workflow? (or at least part of every few releases)

[...]

PR #912, opened 2021-02-10
Security updates first wave

dondi:


dondi:
Initial attempt at merging #674 into beta but it looks like there are some build issues. Will need to work on those.

dondi:
This was done as a sequential phase of the prior release, so the distinct PR is no longer needed

Issue #888, opened 2020-10-23
Reverse arrow button directions for graph


dondi:
The visible arrow buttons correspond better to moving the graph, therefore the left arrow should move the graph to the _left_, etc.

dondi:
Already merged from PR #887.

dondi:
Directionality was confirmed but additional logic is needed to keep the graph within the gray border (for now, while it’s present—see #889). The arrows should move the _graph_ inside the border and the border shouldn’t “tag along” with the movement. Feel free to contact @dondi is @luluriffic runs into any issues.

dondi:
@luluriffic and @dondi have been interacting during the week with how to implement this, with @luluriffic specifically reading through the code and seeking clarification when necessary.

luluriffic:
In the process of writing moveWithinViewport() function

dondi:
@luluriffic is continuing with this and has become more familiar with the code now; there is confidence that is will be done soon.

igreen1:
@dondi I'm a little confused on this issue. Now that the gray border is removed, what is the goal here?

Should the graph be pushed up 'against' the now invisible border ? From #804, 'the arrow keys and center button should still move the entire graph but solely within the visible boundaries' Does 'visible boundaries' mean the viewport (entire 'artboard') or the bounding box within the viewport? Or should behaviour be changed such that viewport size = bounding box size?

dondi:
Indeed, good point @igreen1—there are actually more nuances to this, which is why this set of issues turned out to be trickier than they seemed. I was meaning to ask to brief you after Wednesday but the week got away from me. Do you have some availability to chat about this, maybe Monday afternoon? Probably best to walk you through what we're after over a screenshare

igreen1:
Monday afternoon works perfectly! I'm free after 2pm

dondi:
OK—I have a meeting right at 2pm but 3pm will work; is that OK?

igreen1:
sounds great!

igreen1:
Fixed by moving the graph in the `move` function instead of the camera

igreen1:
`move` was reverted to old logic, with a flip of signs to fix directions. This was done as @kdahlquist didn't want to nodes to 'drift' when movement was done. Now the camera moves but in the opposite direction, mimicking the movement of the graph without requiring an actual movement so the nodes don't drift

ahmad00m:
@ahmad00m confirms that the behavior required is correctly implemented.

Issue #804, opened 2019-10-07
Moving entire graph with "grab hand" moves top and left boundary that individual genes can be moved to


kdahlquist:
I'm not sure how to describe this, but here goes:

Currently, when restrict graph to view port is not checked, we allow the genes to be moved manually beyond the viewport to the right and bottom, but we do not allow them to move beyond the top or left side of the viewport.

I noticed that when I moved the entire graph, by grabbing it with the hand, it changes the top and left boundary for a node to below the viewport.  I.e., I can't manually drag the node beyond an invisible barrier.  

I can demo this in the meeting if it's not clear, just want to record it before I forget.

dondi:
At the meeting we could not replicate this but @igreen1 can do a deeper dive to verify.

igreen1:
This issue yet lives!

I was wrong. I am able to move the graph out of bounds but this issue is for the nodes themselves. I am still unable to move a /node/ (gene) beyond the viewport, but the graph/network as a whole can be moved beyond the viewport.

luluriffic:
Issue isn't consistent! Sometimes I'm able to move a node past the left boundary, but not the top, or past the top, but not the left after moving the whole graph/network. Bizarre and I will investigate further this week.

dondi:
After discussing this further during the meeting, the intended behavior for "Restrict Graph to Viewport" was refined further as follows:
* When "Restrict Graph to Viewport" is turned _off_, the graph and pan/zoom controls are allowed to traverse a pseudo-infinite 2D coordinate space
* When "Restrict Graph to Viewport" is turned _on_, the restriction to a viewport currently affects the nodes but not the pan/zoom controls—but _it should_

1. Pan (arrows + center button) should no longer pertain to movement of the "camera" but should be movement of the entire graph within the coordinate space (a.k.a. artboard). Thus, when "Restrict Graph to Viewport" is turned on, the arrow keys and center button should still move _the entire graph_ but _solely within the visible boundaries_. When "Restrict" is off, the arrows will still work but allow the graph to move within the artboard without restrictions.
2. Zoom should still work when "Restrict" is turned on, but the zoom-out should be restricted to some fixed percentage past the artboard and the gray box representing the viewport should be invisible; zoom-in should still work but is limited to the physical size of the graph—i.e., when "Restrict" is turned on, the user should not be able to zoom to a point where the entire graph isn't visible.

dondi:
[...]

PR #922, opened 2021-04-07
Issues #905, #888, #804

### LINKED ISSUE #905: Log Fold Change Max value selection on sidebar has issues
On @igreen1 machine (Firefox 84.0.2, Ubuntu 20.04), the Log Fold Change Max value selection on the sidebar does not work. It does, however, work on the dropdown (Node ->  Log Fold Change Max Value). This issue exists even after every single extensions was removed from @igreen1's browser (including the seemingly unrelated ones)

image

This is possibly linked to the lack of a 'set' button. On @dondi's machine there is an additional 'set' button next to the input value. This issue requires additional help in replication to find where the bug lies.

No relevant web console output was logged on page load. @igreen1 will try to replicate this issue on Windows Firefox and on Windows Edge/Chrome.

igreen1:
Summary changelog:
grnsight.styl changed to fix #905 by reducing size of button so overflow is unnecessary
Changed button direction for #888 by moving camera in opposite direction (flipping signs)
Changed restrict graph to viewport behaviour in #804. Now when restrict-graph is chosen, movement controls are not available and zoom is locked to 100%.

Pull Request Checklist
- [x] Print works as expected


dondi (approved):
LGTM

dondi:
Ready for checklisting. Please notify @dondi when ready merge.

ahmad00m:
The demos worked as expected at first but after a while the nodes would start moving further apart from one another. Also, the higher number of nodes the more nodes would move apart instead of reaching equilibrium and stopping. Moreover, I noticed some of the options in Top/Bottom Dataset in Node column did not appear on the graph and it showed a message saying "Expression Database is loading".

igreen1:
good catch @ahmad00m that was an interesting bug. the `center` function calls a restart on the simulation. In fact, looking at beta, @igreen1  was able to produce a similar bug by toggling 'restrict graph to viewport'. It causes the nodes to jostle for a long period of time.

This bug can be fixed by removing the restart of simulation in the center command, which does not appear to affect the simulation at all. This will be issued in a commit shortly

igreen1:
Second commit should make it impossible to highlight for all computer types: commit d4e224950cf9dc27a4c863cfa684f0e5a1fd7ffa

@dondi should I add this 'no-highlight' to nodes as well? I'm editing the file anyways.


@igreen1 have been unable to find the reasoning behind the expression database error; however, it happens on the beta branch too so likely a result of needing access, which only the official website has. Will confirm in meetign

dondi:
@dondi will look into network/access restrictions to verify whether the external expression database requests are behaving as expected.

dondi (approved):
LGTM

dondi:
There are some loose-end checkboxes in this PR but I’ll go ahead and merge and we can seek to resolve those as standalone issues.

Issue #923, opened 2021-04-14
Inconsistency of viewport size in Dropdown menu 


ahmad00m:
The dropdown menu for viewport size does not correctly reflect the size of the graph. I have attached a picture for this issue. 

<img width="1222" alt="Screenshot 2021-04-13 at 6 45 37 pm" src="https://user-images.githubusercontent.com/75926755/114642396-b15bc980-9c88-11eb-85bd-de00fe417233.png">

igreen1:
In a last minute sprint, a refactor is ready that puts the viewport size options within `grnstate` and puts the control of viewport size in the `updateApp` logic.

Commit 06982c1fa30bcbcda52e7853ff9fb815426c48f6 holds the final fix for this bug.

dondi:
PR pending; will try to include in the release.

dondi:
Was released with v5.0.0

PR #926, opened 2021-04-21
Igreen1

igreen1:
Fix for #923 by moving logic from `container.js` to new `grnState` model for viewport size control

Pull Request Checklist
- [x] All four Demos look as expected when loaded
- [x] Print works as expected

### DIFF
--- a/web-client/public/js/container.js
+++ b/web-client/public/js/container.js
@@ -50,18 +50,7 @@ export const container = function () {
         });
     }
 
-    requestWindowDimensions();
-
-    if (pageWidth < MEDIUM_PAGE_WIDTH) {
-        $("#boundBoxS").prop("checked", true);
-        container.addClass("containerS");
-    } else if (pageWidth > MEDIUM_PAGE_WIDTH && pageWidth < LARGE_PAGE_WIDTH) {
-        $("#boundBoxM").prop("checked", true);
-        container.addClass("containerM");
-    } else {
-        $("#boundBoxL").prop("checked", true);
-        container.addClass("containerL");
-    }
+    
 
     var setSizeIndicator = function (selector) {
         $("#viewport-size-s span, #viewport-size-m span, #viewport-size-l span, #viewport-size-fit span")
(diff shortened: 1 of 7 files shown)


web-client/public/js/container.js
```diff
@@ -50,80 +51,78 @@ export const container = function () {
...
+
+    // $("#viewport-size-fit").on("click", function () {
+    //     $("#boundBoxFit").prop("checked", true).trigger("click");
+    //     fit();
+    // });
+
+    // $("#boundBoxFit").on("click", function () {
+    //     $("#viewport-size-fit").prop("checked", true);
+    //     fit();
+    // });
```
dondi, line 125: If the fix is verified, probably safe to remove completely in order to avoid confusion.

web-client/public/js/update-app.js
```diff
@@ -210,6 +221,121 @@ const synchronizeHideAllWeights = () => {
     $(WEIGHTS_HIDE_CLASS).addClass("selected");
 };
 
+// Viewport
+const synchronizeViewportSizeSmall = () => {
+    $(VIEWPORT_SIZE_S_DROPDOWN + " span").removeClass("glyphicon-ok");
+    $(VIEWPORT_SIZE_M_DROPDOWN + " span").removeClass("glyphicon-ok");
+    $(VIEWPORT_SIZE_L_DROPDOWN + " span").removeClass("glyphicon-ok");
+    $(VIEWPORT_SIZE_FIT_DROPDOWN + " span").removeClass("glyphicon-ok");
```
dondi, line 229: Fun tip, this is how we can eliminate the latter (repeated) half of these statements:
```javascript
$(
    [
        VIEWPORT_SIZE_S_DROPDOWN + " span",
        VIEWPORT_SIZE_M_DROPDOWN + " span",
        VIEWPORT_SIZE_L_DROPDOWN + " span",
        VIEWPORT_SIZE_FIT_DROPDOWN + " span"
    ].join(", ")
).removeClass("glyphicon-ok");
```

Same pattern can be applied to the other blocks below.

dondi (commented):
Pretty much there; we can eliminate a little redundancy but otherwise ready to go.

dondi (approved):
LGTM, comments all cleared too!

PR #929, opened 2021-05-16
Jade -> Pug Dependency Upgrade

mihirsamdarshi:
This upgrade has been long overdue, so I added it in. I also upgraded the Google API library, given that the previous one had been deprecated.

There was a listing error that I fixed in `container.js`

I also removed the onlysidebar.html file - it wasn't serving any purpose


dondi:
@mihirsamdarshi Sorry for the delay…pandemic summer 😅 But we are back and discovered that some server updates need to take place. Just wanted to say, this is appreciated and not forgotten, and we hope to merge when possible. Thanks!

Issue #935, opened 2021-09-16
Export to unweighted excel feature re-imports as weighted


Onariaginosa:
When you export to unweighted excel, it exports a weighted excel workbook.

dondi:
Expression sheets are OK to export with an unweighted graph because the weights pertain to just the adjacency matrix. The thing to verify is that an unweighted export should re-import without weights.

dondi:
Verified fixed in v5.1.0

Issue #938, opened 2021-09-30
Implement workbook export with user-chosen expression datasets


dondi:
Implement the user interface and functionality for choosing expression datasets to include in an exported workbook.

This task aligns with #935

For maximum flexibility, the user will be allowed to choose from any available dataset, regardless of whether they have used or selected it.

_But_ restrict the selections only to one source at a time, because expression data from multiple sources may not co-exist well, especially with GRNmap.

Onariaginosa:
New Database Schema

dondi:
@Onariaginosa and @dondi looked this and traced the root cause to a missing `id` header row on the problematic expression data sheet. However, the sheet also didn't have any data; this issue may be related. @Onariaginosa will investigate.

Onariaginosa:
I added the fall2021 database to the aws database. On the Onariaginosa_temporary branch, I changed the query to match our current internal database structure, but I was having trouble accessing the results of said query, or even checking to see what the query returns.

dondi:
@Onariaginosa will continue to learn some Sequelize to work out how to access query results and data; she will also provide sample of the expression metadata table so we can populate that table with appropriate values for the Dahlquist data.

Onariaginosa:
Needed for the expression_metadata table:
replicate_index | ncbi_geo_id | pubmed_id | control_yeast_strain | treatment_yeast_strain | control | treatment | concentration_value | concentration_unit | time_value | time_unit | number_of_replicates | expression_table

Onariaginosa:
1 | GSE9336     | 12269742  | S288C                | S288C                  | No Thiuram   | Thiuram   |                  75 | uM                 |         15 | m         |                    3 | Kitagawa_2002_log2_expression   | 
               2 | GSE9336     | 12269742  | S288C                | S288C                  | No Thiuram   | Thiuram   |                  75 | uM                 |         30 | m         |                    3 | Kitagawa_2002_log2_expression   | 
               3 | GSE9336     | 12269742  | S288C                | S288C                  | No Thiuram   | Thiuram   |                  75 | uM                 |        120 | m         |                    3 | Kitagawa_2002_log2_expression   | 
               1 | GSE6129     | 17327492  | W303-1A              | W303-1A                | No Arsenite  | Arsenite  |                   1 | mM                 |         60 | m         |                    6 | Thorsen_2007_log2_expression    | 
[...]

PR #940, opened 2021-10-23
Completed functionality of issue #938 and fixed bugs #935 and #941

Onariaginosa:
Completed the issues mentioned in #938 and shifted to pug (was necessary to run branch on beta, branch Onariaginosa is on jade still just in case I need to revert)
Also fixed bug in  #935 so that when the user selects export unweighted, they get an unweighted graph.
Fixed bug in #941 so that the graph displays properly when you change the bottom data set to Expression data within the database first, then change the top data set to Expression data that requires a query to the database.

Pull Request Checklist
- [x] All four Demos look as expected when loaded
- [x] Print works as expected


ahmad00m:
@ahmad00m went through the check list and determined everything is working except for the export function to SIF and GraphML. I believe they are correctly exported but they don't have the correct graph upon import (their node coloring is turned off). Also, I'm not sure if reload is doing what it is supposed to do which I will further discuss during the meeting. 
On the other hand, the excel export works just fine.

Issue #978, opened 2022-09-07
Clarifying Export-to-Excel interface


kdahlquist:
![Export-to-Excel-modal_v6 0 0](https://user-images.githubusercontent.com/6463867/188939459-9b2d34ba-6e66-4e4f-a373-e73ad1e514e9.jpg)

I'd like to clarify the language of the Export-to-Excel modal.

The top part should say:
Select the edge type
- Unweighted
- Weighted

Instead of not showing weighted when it's not possible, let's show it, but have it grayed out.

For the bottom part:
The current data should be the top selection, but for the rest, could they be in alphabetical order like they appear in the expression data drop-down menu?

In my mind, if I select expression data from the drop-down in the node menu, that is the option that should be selected when I chose export data to excel.  For example, if I load demo #1, change the data to Barreto and then click export, Barreto data should be the one selected, not demo 1.

kdahlquist:
![Export-to-Excel-modal2_v6 0 0](https://user-images.githubusercontent.com/6463867/188941486-fe52b935-fd4b-4ea8-a61c-731b18ecb3d3.jpg)

The second modal then needs to be dependent on the first.  If I chose Barreto data, then that is the only option for an expression data sheet that I should be given.  It looks like it is adding the Dahlquist data to this.

kdahlquist:
I would also generalize the second modal so that the use can select whether or not to export any of the worksheets:
- [ ] degradation_rates
- [ ] production_rates
- [ ] wt_log2_expression
- [ ] any other additional _log2_expression sheets
- [ ] network_weights
- [ ] optimization_parameters
- [ ] threshold_b

Note that for non-Dahlquist data, they are all wt for now, so for Barreto, for example, it would just be the one sheet "wt_log2_expression"

Note also, the "network_weights" sheet is _in addition_ to the "network" sheet.  It is a duplicate with the same name, but is necessary for GRNmap to work.

kdahlquist:
Also, regarding the placement of the buttons.

Should the Continue button go down and to the left of the Close button?

Also, maybe the Close button should be labeled Cancel

kdahlquist:
For the second modal, move the Export button to the left of the "Close" button, but rename "Close" to "Cancel".  Also, since there are two modal windows in a series, should there be a "back" button to take you back to the previous one?

Onariaginosa:
I started making headway on this refactor / clarification

- I moved the buttons and added the back button all into the footer. I renamed close to cancel.
[...]

Issue #977, opened 2022-09-07
Node menu inactivated after loading network from database


kdahlquist:
If the first thing I do upon landing on the GRNsight page is to load a network from a database, the node menu is disabled, so I am unable to color the nodes with expression data from the expression database.

I think that now that we have an expression database, the node menu should always be enabled because even if a user doesn't have their own data, they can use the database data.

dondi:
Although it is possible to work around this by loading a demo first, and then loading a network from a database, there is now a follow-up issue where node coloring actually doesn’t take place until a dataset is chosen

ahmad00m:
I believe if the node coloring should always be enabled this function need to be always enabled. This function is in `setup-handlers.js`, I tried to test it on localhost but I could not connect to it through the multiple attempts I made. So, first I hope to troubleshoot connecting to localhost and then check whether this is the right piece of code that needs to be changed. 

<img width="714" alt="Screen Shot 2022-09-14 at 4 56 44 AM" src="https://user-images.githubusercontent.com/75926755/190147371-8b0f0e75-452f-4826-949f-0a3bef6b726e.png">

ahmad00m:
UPDATE 
I tried to run the localhost on my laptop and it was able to connect to the server; however, I still got an error in my console and I was unable to load demos.

<img width="1440" alt="Screen Shot 2022-09-14 at 9 25 17 AM" src="https://user-images.githubusercontent.com/75926755/190210440-727f24ea-ba57-4f94-874b-6c6aab762396.png">

dondi:
@ahmad00m Regarding node coloring, I would agree that it is appropriate for it to be always enabled now, except where there is no network loaded

Go ahead and start a branch to work on it, or possible pair w/ @Sarronnn since she will also benefit from doing more of these types of fixes

dondi:
For the demo errors, both @Onariaginosa and @dondi confirmed that their local builds are loading demos successfully, so that looks like a localized issue

dondi:
Local build remains unresolved but bug fix could be done as a pair w/ @Sarronnn

dondi:
Local issue with @ahmad00m has been fixed. Turns out that it was a specific problem with Node 17. Switching to Node 16 or  18 resolves it; @ahmad00m chose to go to 18 and GRNsight ran successfully

ahmad00m:
Here is the linked that helped resolve this issue.
https://stackoverflow.com/a/69789658

dondi:
[...]

Issue #976, opened 2022-08-31
Demo indicators appear to out of sync


dondi:
In current v6.0.0, choosing a demo from the sidebar dropdown menu does not update the UI consistently.

Choosing from the menubar continues to work as expected.

Onariaginosa:
I fixed this issue on the Onariaginosa branch!

kdahlquist:
PR #985

dondi:
Already merged into beta and verified at an earlier meeting

Issue #972, opened 2022-08-12
creating network from database miscounts edges in display of name


kdahlquist:
I used the create a network from database tool to make a network with 12 genes and 17 edges.  The count of the edges is correct on the far right of the menu toolbar, but the Custom Network "filename" that displays to the right of the demo menu miscounts the edges and says "12 genes, 7 edges".  This carries through to the export to Excel menu.  See screenshots:

![image](https://user-images.githubusercontent.com/6463867/184408619-2d8f56a8-d214-4b5d-8619-b38ad62b1e80.png)

![image](https://user-images.githubusercontent.com/6463867/184408717-c6c80715-cc19-446e-801c-586a26322ca1.png)

Onariaginosa:
This was fixed in WIP PR #985

dondi:
Already merged

PR #985, opened 2022-09-14
WIP: Fixed Demo Indicators out of sync bug and updated export modal and functionality

Onariaginosa:
#976 Fixed this issue. The demo sidebar dropdown works as expected now.
#978 UI fixes and beginnings of the new excel export implementation.
#990 Fixed this issue. GRNsight properly handles when a database generated network does not have any regulatory connections.
#977 @ahmad00m also fixed the node coloring menu inactivated issue in this pr
#972 Fixed this issue. Edges are now counted correctly in the custom workbook title
#961 Documentation for loading database scripts completed


web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -167,9 +168,11 @@ export const setupLoadAndImportHandlers = grnState => {
         $(demoClass).on("click", () => {
             loadDemo(demoPath, demoClass, demoName);
         });
-
         $("#demoSourceDropdown").on("change", () => {
-            loadDemo(demoPath, demoClass, demoName);
+            let selected = `.${$("#demoSourceDropdown").val()}`;
```
dondi, line 172: When you know for sure that you won’t need to reassign a variable, default to `const` instead of `let`
```suggestion
            const selected = `.${$("#demoSourceDropdown").val()}`;
```

web-client/public/js/upload.js
```diff
@@ -386,33 +391,78 @@ export const upload = function () {
...
+    // var handleExportButtonBack = function (weight, source) {
+    //     return function () {
+    //         handleExportExcelModal();
+    //         $("input[name=network-weights]").removeAttr("checked");
+    //         $("input[name=network-weights][value=" + weight + "]").prop("checked", true);
+    //         $("input[name=expressionSource]").removeAttr("checked");
+    //         $("input[name=expressionSource][value=" + source + "]").prop("checked", true);
+    //     };
+    // };
+
```
dondi, line 436: Looks like this can go bye-bye
```suggestion
```

dondi (commented):
Just a couple of tweaks and I think this will work!

web-client/public/js/createNetwork.js
```diff
@@ -114,14 +135,21 @@ containing "-", "_", and alpha-numeric characters only`);
...
     // get sources from database
         queryNetworkDatabase({type:"NetworkSource", info:null}).then(function (response) {
             $("#createNetworkQuestions-container").append(createHTMLforForm(Object.keys(response.sources)));
             grnState.customWorkbook.sources = response.sources;
             grnState.customWorkbook.source = Object.keys(response.sources).length === 1 ?
                 Object.keys(response.sources)[0] : null;
+            for (let source in response.sources) {
+                let i = source.indexOf(":");
+                let date = new Date(source.substring(i + 1));
+                date = getFormattedDateTime(date);
```
dondi, line 150: Hmmm this is a little sketchy. Starting the `date` variable out as a `Date` and then replacing it with its string…JavaScript lets you do this but other languages won’t, and ideally we don’t get into the habit of doing this. If we create a new value, we should put that in a new variable

server/controllers/additional-sheet-parser.js
```diff
@@ -315,7 +315,7 @@ module.exports = function (workbookFile) {
             // above line creates an object from the optimization paramerters sheet
             // these are part of the "meta" property
         } else if (TWO_COL_SHEET_NAMES.includes(sheet.name)) {
-            output["test"][sheet.name] = parseTwoColumnSheet(sheet);
+            output["twoColumnSheets"][sheet.name] = parseTwoColumnSheet(sheet);
```
dondi, line 318: Any reason why this isn’t `output.twoColumnSheets`? (it’s written that way elsewhere)

dondi (commented):
Overall looks pretty good except for just a couple of code style fine points. Once those are changed I think this can go…

server/controllers/additional-sheet-parser.js
Onariaginosa: I changed it to to `output.twoColumnSheets`. It must of been residual from fixing linter errors.

web-client/public/js/createNetwork.js
Onariaginosa: I changed this as requested

web-client/public/js/createNetwork.js
```diff
@@ -144,11 +144,14 @@ containing "-", "_", and alpha-numeric characters only`);
...
+            let timestamp;
             for (let source in response.sources) {
-                let i = source.indexOf(":");
-                let date = new Date(source.substring(i + 1));
-                date = getFormattedDateTime(date);
-                grnState.customWorkbook.sources[source].timestamp = date;
+                i = source.indexOf(":");
+                dateTime = new Date(source.substring(i + 1));
+                timestamp = getFormattedDateTime(dateTime);
+                grnState.customWorkbook.sources[source].timestamp = timestamp;
```
dondi, line 154: I see what you did here but…why keep the `let`? These variables are not used outside of the block scope of the `for` loop, so there is no need to define them outside. Also, in this new version, variables aren’t re-assigned, so overall this is the cleanest option, I’d say:
```suggestion
                const i = source.indexOf(":");
                const dateTime = new Date(source.substring(i + 1));
                const timestamp = getFormattedDateTime(dateTime);
                grnState.customWorkbook.sources[source].timestamp = timestamp;
```

(removing all the preceding `let`s of course)

dondi (commented):
Almost there! Just one little tweak. Values reflected here based on the latest capabilities of JavaScript:
- Use `const` as much as possible
- Declare variables at the deepest scope in which they’re needed

dondi (approved):
LGTM 🤗

Issue #995, opened 2022-10-19
Investigate beta 6.0.2 deployment database regression


dondi:
When #985 was merged into beta, that was pulled into the server but doing so appears to have regressed the database functionality: gene searches no longer produce results. The same branch was verified to have worked locally for @ahmad00m so the issue looks like it is server-specific

dondi:
Based on @Onariaginosa’s findings, the inconsistencies we have noticed are due to time zone shifts. The general principle for handling this is to store all timestamp values in UTC and to defer all computations to local time until the very end, if at all. In this particular case, since timestamp is not involved in the user interface, our goal is to move everything to UTC

Our long-term fix is to be fully “time-zone-aware” across our stack. This involves:
1. We will consider original data to be in UTC; in the future, any code that adds a timestamp should include time zone with it
2. Our `source` table must change the `time_stamp` column to type `timestamp with time zone`—and further, we will store values in this table in UTC
3. All of our browser code must be “aware” of time zone as a factor and must convert all timestamp parameters to UTC before sending to our API

Based on looking at current behavior, short-term fix may be just #3. Recommendation is to use the `date-fns` to parse and format dates—this is the current go-to library for reliable date processing. @dondi and @Onariaginosa can work offline to do this, with @ahmad00m and @Sarronnn being welcome to join, time permitting 😁

dondi:
This is looking fixed; just need a couple of follow-ups to ensure that this isn’t a one-off:
1. As it turns out, Los Angeles drops Daylight Savings Time this weekend, so re-checking functionality after we change time zones will be worthwhile
2. @ahmad00m will verify the “start from scratch” sequence to make sure that these adjustments work well when initializing a database from the beginning
3. Less crucial but potentially useful: have another machine temporarily set to a different time zone connect to the “start from scratch” system to make sure that the time zone fix holds—check your network settings beforehand to determine the IP address of your local machine on the local network, and use that instead of `localhost` to connect from the different-time-zone computer

ahmad00m:
With the help of @Onariaginosa  I was able to make a new copy of the database with all the new changes in the schema. However, for some unknown (at the moment) reasons the queries are still pending and nothing gets returned. Attached are screenshots of what this exactly means. I also tried changing the time zone, however the new mac update is weird and doesn't change the time zone but I was able to change the time manually, so this might have some effect on the timestamps shown for the network source. 
[...]

Issue #989, opened 2022-09-21
yeast genes that have naming patterns not allowed by load network from database function


kdahlquist:
I've been playing around with the load network from database function to see what happens when I try to find yeast genes that have unconventional naming patterns.
- First, the good news.  The 3 genes that had a comma in their standard name and 1 gene that had an apostrophe got renamed!  So no problems there.
- However, I've found 2 genes that have parentheses in their standard names, `MF(ALPHA)1` and `MF(ALPHA)2`.  If I search for either of those with their parentheses, they trigger the error modal.  They are in the database, though, because if I search on them based on their systematic names, YPL187W and YGL089C, they come up properly.  I think we are going to need to allow parentheses in the input to that search field so that they will be properly retrieved.  
- I did a scan of some old microarray data I had to find these.  We should probably scan the data in the database to make sure that there aren't any other oddballs out there.

kdahlquist:
Just for posterity, here are the genes I tested.
![oddball-yeast-genes](https://user-images.githubusercontent.com/6463867/191616071-55016af8-1858-4f3b-a971-2335a441a1e7.jpg)

kdahlquist:
* DUR12 used to be DUR1,2
* IMP21 used to be IMP2'
* ARG56 used to be ARG5,6
* ADE57 used to be ADE5,7

Onariaginosa:
I scanned the database and found that the only genes in our database with unusual names are the following:
![Screenshot from 2022-11-09 11-10-41](https://user-images.githubusercontent.com/21343072/200919907-0468ee26-1edd-480e-a98a-a41894a05a7f.png)

Because of this I just created an exceptions object that maps the irregular name to a name that we accept within GRNsight:
![Screenshot from 2022-11-09 11-12-37](https://user-images.githubusercontent.com/21343072/200920283-8ec4a93e-481e-4900-bffd-eb86cd9bb14f.png)

This issues should be resolved in PR #996, as I was working on fixes for that simultaneously.

Issue #984, opened 2022-09-07
Bad interaction between node coloring, load network from database, and export to Excel


kdahlquist:
As noted in issue #977, the node coloring menu should be activated all the time now because we have a gene expression database.  One fallout is this interaction. Do the following:
- load a demo file
- load a network from the database
- the node coloring panel/menu is now active, with the enable node coloring checkbox checked, but no node coloring is showing on the graph
- Export to Excel and the modal opens, but you can't actually click on anything (radio buttons or other buttons) nothing works.

Onariaginosa:
I fixed this and it should make its way into the never ending pr #996. The `grnState.node.coloring.topDataset` and `grnState.nodeColoring.bottomDataset` settings weren't getting when a user generates a new network, so I set them = to undefined and it worked.

kdahlquist:
confirmed fixed in beta 6.0.7

PR #996, opened 2022-11-02
Fixed Database Time Zone, Finished Clarifying Export to Excel Functionality, Fixed Irregular Gene Names Bug

Onariaginosa:
#984 #978, #995, #989 were completed and fixed!

### DIFF
--- a/server/dals/expression-dal.js
+++ b/server/dals/expression-dal.js
@@ -96,17 +96,14 @@ const buildExpressionDataQuery = function (dataset, timepoints, genes) {
 
 
 const buildExpressionQuery = function (query) {
-    switch (query.type) {
-    case "DegradationRates":
-        return buildExpressionProductionDegradationRatesQuery("degradation_rate", query.genes);
-    case "ProductionRates":
-        return buildExpressionProductionDegradationRatesQuery("production_rate", query.genes);
-    case undefined:
-        // This is the original expression database call.
-        // Will restructure code so that it follows the standardized query type, query data structure that all
-        // other queries use
-        return buildExpressionDataQuery(query.dataset, query.timepoints, query.genes);
+    const expressionQueries = {
+        "DegradationRates": buildExpressionProductionDegradationRatesQuery("degradation_rate", query.genes),
+        "ProductionRates" : buildExpressionProductionDegradationRatesQuery("production_rate", query.genes),
+        "ExpressionData" : buildExpressionDataQuery(query.dataset, query.timepoints, query.genes)
     }
+    if (Object.keys(expressionQueries).includes(query.type)) {
+        return expressionQueries[query.type]
+    } 
 };
 
 const listExpressionGeneData = function (gene, totalOutput) {
@@ -144,19 +141,17 @@ module.exports = {
         return sequelize.query(buildExpressionQuery(req.query),
                 { type: sequelize.QueryTypes.SELECT })
                     .then(function (stdname) {
+                        convertToJSON = {
+                            "DegradationRates" : convertProductionDegradationRateToJSON(stdname, "degradation_rate"),
+                            "ProductionRates" : convertProductionDegradationRateToJSON(stdname, "production_rate"),
+                            "ExpressionData" : convertExpressionToJSON(
+                                stdname, req.query.dataset, expressionTimepointsByDataset[req.query.dataset], req.query.genes.split(","))
+                        }
                         const type = req.query.type;
-                        if (type === "DegradationRates") {
-                            return res.send(convertProductionDegradationRateToJSON(stdname, "degradation_rate"));
-
-                        } else if (type === "ProductionRates") {
-                            return res.send(convertProductionDegradationRateToJSON(stdname, "production_rate"));
-
-                        } else if (type === undefined) {
-                            let dataset = req.query.dataset;
-                            let geneList = req.query.genes.split(",");
-                            let response = convertExpressionToJSON(
-                                stdname, dataset, expressionTimepointsByDataset[dataset], geneList);
-                            return res.send(response);
+                        if (Object.keys(convertToJSON).includes(type)) {
+                            return res.send(convertToJSON[type])
+                        } else {
+                            return res.send(500, { errors: "Something went wrong."});  
                         }
                     });
     }
--- a/server/dals/expression-dal.js
+++ b/server/dals/expression-dal.js
@@ -100,10 +100,10 @@ const buildExpressionQuery = function (query) {
         "DegradationRates": buildExpressionProductionDegradationRatesQuery("degradation_rate", query.genes),
         "ProductionRates" : buildExpressionProductionDegradationRatesQuery("production_rate", query.genes),
         "ExpressionData" : buildExpressionDataQuery(query.dataset, query.timepoints, query.genes)
-    }
+    };
     if (Object.keys(expressionQueries).includes(query.type)) {
-        return expressionQueries[query.type]
-    } 
+        return expressionQueries[query.type];
+    }
 };
 
 const listExpressionGeneData = function (gene, totalOutput) {
@@ -141,17 +141,21 @@ module.exports = {
         return sequelize.query(buildExpressionQuery(req.query),
                 { type: sequelize.QueryTypes.SELECT })
                     .then(function (stdname) {
-                        convertToJSON = {
+                        const convertToJSON = {
                             "DegradationRates" : convertProductionDegradationRateToJSON(stdname, "degradation_rate"),
                             "ProductionRates" : convertProductionDegradationRateToJSON(stdname, "production_rate"),
                             "ExpressionData" : convertExpressionToJSON(
-                                stdname, req.query.dataset, expressionTimepointsByDataset[req.query.dataset], req.query.genes.split(","))
-                        }
+                                stdname,
+                                req.query.dataset,
+                                expressionTimepointsByDataset[req.query.dataset],
+                                req.query.genes.split(",")
+                            )
+                        };
                         const type = req.query.type;
                         if (Object.keys(convertToJSON).includes(type)) {
-                            return res.send(convertToJSON[type])
+                            return res.send(convertToJSON[type]);
                         } else {
-                            return res.send(500, { errors: "Something went wrong."});  
+                            return res.send(500, { errors: "Something went wrong."});
                         }
                     });
     }
--- a/server/dals/expression-dal.js
+++ b/server/dals/expression-dal.js
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
 const Sequelize = require("sequelize");
 require("dotenv").config();
 var env = process.env.NODE_ENV || "development";
@@ -97,12 +98,12 @@ const buildExpressionDataQuery = function (dataset, timepoints, genes) {
 
 const buildExpressionQuery = function (query) {
     const expressionQueries = {
-        "DegradationRates": buildExpressionProductionDegradationRatesQuery("degradation_rate", query.genes),
-        "ProductionRates" : buildExpressionProductionDegradationRatesQuery("production_rate", query.genes),
-        "ExpressionData" : buildExpressionDataQuery(query.dataset, query.timepoints, query.genes)
+        "DegradationRates": () => {return buildExpressionProductionDegradationRatesQuery("degradation_rate", query.genes);},
+        "ProductionRates" : () => {return buildExpressionProductionDegradationRatesQuery("production_rate", query.genes);},
+        "ExpressionData" : () => {return buildExpressionDataQuery(query.dataset, query.timepoints, query.genes);}
     };
     if (Object.keys(expressionQueries).includes(query.type)) {
-        return expressionQueries[query.type];
+        return expressionQueries[query.type]();
     }
 };
 
@@ -142,21 +143,20 @@ module.exports = {
                 { type: sequelize.QueryTypes.SELECT })
                     .then(function (stdname) {
                         const convertToJSON = {
-                            "DegradationRates" : convertProductionDegradationRateToJSON(stdname, "degradation_rate"),
-                            "ProductionRates" : convertProductionDegradationRateToJSON(stdname, "production_rate"),
-                            "ExpressionData" : convertExpressionToJSON(
-                                stdname,
-                                req.query.dataset,
-                                expressionTimepointsByDataset[req.query.dataset],
-                                req.query.genes.split(",")
-                            )
+                            "DegradationRates" : () => {return convertProductionDegradationRateToJSON(stdname, "degradation_rate");},
+                            "ProductionRates" : () => {return convertProductionDegradationRateToJSON(stdname, "production_rate");},
+                            "ExpressionData" : () => {
+                                return convertExpressionToJSON(
+                                    stdname,
+                                    req.query.dataset,
+                                    expressionTimepointsByDataset[req.query.dataset],
+                                    req.query.genes.split(",")
+                                );
+                            }
                         };
                         const type = req.query.type;
-                        if (Object.keys(convertToJSON).includes(type)) {
-                            return res.send(convertToJSON[type]);
-                        } else {
-                            return res.send(500, { errors: "Something went wrong."});
-                        }
+
+                        return (Object.keys(convertToJSON).includes(type)) ? res.send(convertToJSON[type]()) : res.send(500, { errors: "Something went wrong."});
                     });
     }
 };
--- a/server/dals/expression-dal.js
+++ b/server/dals/expression-dal.js
@@ -98,9 +98,9 @@ const buildExpressionDataQuery = function (dataset, timepoints, genes) {
 
 const buildExpressionQuery = function (query) {
     const expressionQueries = {
-        "DegradationRates": () => {return buildExpressionProductionDegradationRatesQuery("degradation_rate", query.genes);},
-        "ProductionRates" : () => {return buildExpressionProductionDegradationRatesQuery("production_rate", query.genes);},
-        "ExpressionData" : () => {return buildExpressionDataQuery(query.dataset, query.timepoints, query.genes);}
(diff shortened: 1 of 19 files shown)


dondi:
After merge, @Onariaginosa will onboard @ahmad00m and @Sarronnn on the process of updating beta

server/dals/expression-dal.js
```diff
@@ -107,16 +130,34 @@ let convertExpressionToJSON = function (totalOutput, dataset, timePoints, allGen
...
+                        } else if (type === "ProductionRates") {
+                            return res.send(convertProductionDegradationRateToJSON(stdname, "production_rate"));
+
+                        } else if (type === undefined) {
+                            let dataset = req.query.dataset;
+                            let geneList = req.query.genes.split(",");
+                            let response = convertExpressionToJSON(
+                                stdname, dataset, expressionTimepointsByDataset[dataset], geneList);
+                            return res.send(response);
+                        }
```
dondi, line 160: Hmmm second time I’m seeing this pattern…maybe we should consider refactoring to data-driven or else we might never get around to it 😅

web-client/public/js/api/grnsight-api.js
```diff
@@ -19,6 +19,36 @@ const buildExpressionURL = function (selection, genes) {
...
+    let result = "";
+    if (Array.isArray(genes)) {
+        for (let gene of genes) {
+            result += `${gene},`;
+        }
+    } else {
+        for (let gene in genes) {
+            result += `${gene},`;
+        }
+    }
```
dondi, line 33: Isn’t the first case doable with `.join`? Also, what data structure does the `else` case cover—an object? An iterable?

Looking to see how we can consolidate this

web-client/public/js/api/grnsight-api.js
```diff
@@ -19,6 +19,36 @@ const buildExpressionURL = function (selection, genes) {
...
+    let baseQuery = `expressiondb?type=${queryType}`;
+    if (queryInfo !== null) {
+        for (let header in queryInfo) {
+            if (header === "genes") {
+                baseQuery += `&${header}=${buildNetworkGenesQuery(queryInfo[header])}`;
+            } else {
+                baseQuery += `&${header}=${queryInfo[header]}`;
+            }
+        }
+    }
```
dondi, line 48: I think there is a simpler way to build this—see `URLSearchParams` https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

dondi (commented):
Overall pretty good, but there are some specific refactoring opportunities…I thought at first we might be able to wait on them, but there are enough that it’s probably better to settle them now. If the comments aren’t clear enough, let me know and I can go over what I have in mind

web-client/public/js/api/grnsight-api.js
Onariaginosa: Sometimes the genes are passed in as an object with keys, and other times the genes are passed in as an array, so if it's an object you can iterate through the keys with `let gene in genes`, and if it's an array you can iterate through the genes as is with `let gene of genes`.

web-client/public/js/api/grnsight-api.js
Onariaginosa: So, I tried implementing this change, but I keep on getting 404 not found errors. I printed the new url string and it looked fine, but the actual network call is returning errors.
Screenshot from 2022-11-09 10-56-03

server/dals/expression-dal.js
Onariaginosa: I changed the if else and the above switch statements for these functions, however there are still a lot of switch statements and if else in the code. Should I do a refactor for all instances of this?

server/dals/expression-dal.js
```diff
@@ -96,16 +96,13 @@ const buildExpressionDataQuery = function (dataset, timepoints, genes) {
...
-        return buildExpressionProductionDegradationRatesQuery("production_rate", query.genes);
-    case undefined:
-        // This is the original expression database call.
-        // Will restructure code so that it follows the standardized query type, query data structure that all
-        // other queries use
-        return buildExpressionDataQuery(query.dataset, query.timepoints, query.genes);
+    const expressionQueries = {
+        "DegradationRates": buildExpressionProductionDegradationRatesQuery("degradation_rate", query.genes),
+        "ProductionRates" : buildExpressionProductionDegradationRatesQuery("production_rate", query.genes),
+        "ExpressionData" : buildExpressionDataQuery(query.dataset, query.timepoints, query.genes)
```
dondi, line 102: Suggest changing to functions in order to avoid evaluating all three even if you end up using just one

(this is a general pattern when you are doing data-driven implementation—watch out for when expressions are evaluated and generally try to avoid doing that until you know for sure that’s what you want)

server/dals/expression-dal.js
```diff
@@ -144,19 +141,21 @@ module.exports = {
...
                     .then(function (stdname) {
+                        const convertToJSON = {
+                            "DegradationRates" : convertProductionDegradationRateToJSON(stdname, "degradation_rate"),
+                            "ProductionRates" : convertProductionDegradationRateToJSON(stdname, "production_rate"),
+                            "ExpressionData" : convertExpressionToJSON(
+                                stdname,
+                                req.query.dataset,
+                                expressionTimepointsByDataset[req.query.dataset],
+                                req.query.genes.split(",")
+                            )
```
dondi, line 152: Same about functions.

server/dals/expression-dal.js
```diff
@@ -144,19 +141,21 @@ module.exports = {
...
-                            let dataset = req.query.dataset;
-                            let geneList = req.query.genes.split(",");
-                            let response = convertExpressionToJSON(
-                                stdname, dataset, expressionTimepointsByDataset[dataset], geneList);
-                            return res.send(response);
+                        if (Object.keys(convertToJSON).includes(type)) {
+                            return res.send(convertToJSON[type]);
+                        } else {
+                            return res.send(500, { errors: "Something went wrong."});
                         }
```
dondi, line 159: This can be a conditional expression `? :`
```suggestion
                        const response = Object.keys(convertToJSON).includes(type) ?
                          res.send(convertToJSON[type]) :
                          res.send(500, { errors: "Something went wrong."});

                        return response;
```

…or something like that. If you can align the parameters, you might be able consolidate the `res.send`

server/dals/network-dal.js
```diff
@@ -51,8 +51,8 @@ const buildQueryByType = function (queryType, query) {
         return buildNetworkSourceQuery();
     case "NetworkGeneFromSource":
         return buildNetworkGeneFromSourceQuery(query.gene, query.source, query.timestamp);
-    case "CreateNetwork":
-        return buildCreateNetworkQuery(query.genes, query.source, query.timestamp);
+    case "GenerateNetwork":
+        return buildGenerateNetworkQuery(query.genes, query.source, query.timestamp);
     }
```
dondi, line 56: This is a `switch` with no `default`: is it intentional for this function to return `undefined` if none of the cases match?

dondi: Data-driven design is probably called for here too

web-client/public/js/generateNetwork.js
```diff
@@ -151,22 +179,21 @@ containing "-", "_", and alpha-numeric characters only`);
...
-            let headers = {
-                type:"CreateNetwork",
-                info: {
-                    genes: grnState.customWorkbook.genes,
-                    source:grnState.customWorkbook.sources[source].source,
-                    timestamp:grnState.customWorkbook.sources[source].timestamp
-                }
+            const genes = Object.keys(grnState.customWorkbook.genes).map(g => {
+                return grnState.customWorkbook.genes[g];
+            });
```
dondi, line 184: Interesting that lint gets mad here. This may be a setting

web-client/public/js/update-app.js
```diff
@@ -372,8 +372,11 @@ const enableNodeColoringUI = function () {
 const loadExpressionDatabase = function (isTopDataset) {
     startLoadingIcon();
     queryExpressionDatabase({
+        type:"ExpressionData",
         dataset: isTopDataset ? grnState.nodeColoring.topDataset : grnState.nodeColoring.bottomDataset,
-        genes : grnState.workbook.genes
+        genes : grnState.workbook.genes.map(x => {
+            return x.name;
+        }).join(",")
```
dondi, line 379: Might be worth a separate issue to see if we can configure the linter to accept the one-line version of this

dondi (changes requested):
Getting close—main refactor is to use functions in data-driven designs. Plus a few other minor ones

web-client/public/js/update-app.js
Onariaginosa: I fixed that! And made them all one liners again

web-client/public/js/generateNetwork.js
Onariaginosa: It was indeed a lint setting. I updated the `.eslintrc.yaml` file and changed the settings

server/dals/network-dal.js
Onariaginosa: I fixed that as well

server/dals/expression-dal.js
Onariaginosa: I fixed this too. It only calls the function it needs

server/dals/expression-dal.js
Onariaginosa: I fixed this

web-client/public/js/api/grnsight-api.js
```diff
@@ -8,17 +7,57 @@ const buildExpressionTimepointsString = function (selection) {
...
+        for (let header in queryInfo) {
+            if (header === "genes") {
+                searchParams.append(header, buildNetworkGenesQuery(queryInfo[header]));
+            } else {
+                searchParams.append(header, queryInfo[header]);
+            }
+        }
+    }
+    console.log(searchParams.toString())
+    return searchParams.toString()
```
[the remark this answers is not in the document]
Onariaginosa: This has been fixed as well

web-client/public/js/api/grnsight-api.js
```diff
@@ -8,17 +7,57 @@ const buildExpressionTimepointsString = function (selection) {
...
+            if (header === "genes") {
+                baseQuery += `&${header}=${buildNetworkGenesQuery(queryInfo[header])}`;
+            } else {
+                baseQuery += `&${header}=${queryInfo[header]}`;
+            }
+        }
+    }
+    return baseQuery;
+
+    const searchParams = new URLSearchParams('expressiondb?');
```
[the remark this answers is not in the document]
Onariaginosa: This is fixed

server/dals/expression-dal.js
```diff
@@ -96,17 +96,14 @@ const buildExpressionDataQuery = function (dataset, timepoints, genes) {
...
-        // other queries use
-        return buildExpressionDataQuery(query.dataset, query.timepoints, query.genes);
+    const expressionQueries = {
+        "DegradationRates": buildExpressionProductionDegradationRatesQuery("degradation_rate", query.genes),
+        "ProductionRates" : buildExpressionProductionDegradationRatesQuery("production_rate", query.genes),
+        "ExpressionData" : buildExpressionDataQuery(query.dataset, query.timepoints, query.genes)
     }
+    if (Object.keys(expressionQueries).includes(query.type)) {
+        return expressionQueries[query.type]
+    }
```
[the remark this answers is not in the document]
Onariaginosa: This is also fixed

web-client/public/js/update-app.js
```diff
@@ -372,8 +372,9 @@ const enableNodeColoringUI = function () {
 const loadExpressionDatabase = function (isTopDataset) {
     startLoadingIcon();
     queryExpressionDatabase({
+        type:"ExpressionData",
         dataset: isTopDataset ? grnState.nodeColoring.topDataset : grnState.nodeColoring.bottomDataset,
-        genes : grnState.workbook.genes
+        genes : grnState.workbook.genes.map(x => {return x.name;}).join(",")
```
dondi, line 377: That sneaky `x` again 😅—here I think the ideal name is clear, yes? It’s mapping from a `genes` array, so each member item is a…

web-client/public/js/upload.js
```diff
@@ -83,6 +91,121 @@ export const upload = function () {
...
+                                }
+                            });
+                            if (finalExportSheets["optimization_parameters"] === null) {
+                                finalExportSheets["optimization_parameters"] = updateOptimizationParameters(finalExportSheets);
+                            }
+                            grnState.workbook.exportSheets = finalExportSheets;
+                            exportExcel(route, extension, sheetType);
+                        }
+                    }
+                }).catch(function (error) {
```
dondi, line 199: To keep the indentation manageable, this `function (response)` (and why isn’t it just `response =>`?) can be defined externally thus allowing this line to become something like `.then(expressionDataHandler)` or whatever name it gets

web-client/public/js/upload.js
```diff
@@ -121,96 +245,71 @@ export const upload = function () {
...
-                    console.log(error.stack);
-                    console.log(error.name);
-                    console.log(error.message);
-                });
+
+                    }).catch(function (error) {
+                        console.log(error.stack);
+                        console.log(error.name);
+                        console.log(error.message);
+                    });
```
dondi, line 306: Second appearance of this function—can be defined as some standard `errorHandler` so that you would just need to say `catch(errorHandler)` here

web-client/public/js/upload.js
```diff
@@ -337,8 +437,9 @@ export const upload = function () {
         ] : [
             (grnState.workbook.meta2 !== undefined && "optimization_diagnostics")
         ];
-        additionalsheets = additionalsheets.filter(x => x !== false && x !== "optimization_parameters");
-        additionalsheets = ["optimization_parameters", ...additionalsheets].sort();
+        additionalsheets = additionalsheets.filter(x => (x !== false && -1 !== optionalAdditionalSheets.indexOf(x)));
```
dondi, line 440: `x !== false` is more verbose than it has to be, yes?

dondi: …unless there is a specific intent here to also allow `x` values of `undefined` and `null` to make this condition true. And if so, that is somewhat irregular so therefore worth a comment

dondi (approved):
So, a good number of style refinements can still be done here, but nothing that can’t be deferred. I’ll go ahead and accept this and perhaps the style issues can be taken on by the team next semester

web-client/public/js/api/grnsight-api.js
```diff
@@ -1,25 +1,14 @@
...
-    return selection.timepoints ?
-    `${baseQuery}&timepoints=${buildExpressionTimepointsString(selection)}` :
-    baseQuery;
+// General DB Access Functions
+const buildQueryURL = function (path, parameters) {
+    const searchParams = new URLSearchParams("");
+    for (let p in parameters) {
+        searchParams.append(p, parameters[p]);
+    }
+    return `${path}?${searchParams.toString()}`;
```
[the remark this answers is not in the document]
ceciliazaragoza: I removed the .toString() from the line

server/dals/network-dal.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
```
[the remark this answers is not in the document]
ceciliazaragoza: I deleted this line and did not see any errors in the console, so it doesn't seem to be needed anymore.

web-client/public/js/upload.js
ceciliazaragoza: I created a function called `expressionExportErrorHandler` that is used in the functions `handleExpressionDataAndExport` and `handleExportExcelButtonExport`. It is called like this: `.catch((error) => expressionExportErrorHandler(error))`

server/dals/network-dal.js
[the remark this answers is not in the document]
ceciliazaragoza: After running `npm run lint` I received this error without `/* eslint-disable max-len */`: `/Users/ceciliazaragoza/Documents/GRNsight/server/dals/network-dal.js 25:1  error  Line 25 exceeds the maximum line length of 120  max-len`. I resolved this by changing the line breaks in the multi-line string in the method `buildNetworkGeneFromSourceQuery`.

web-client/public/js/upload.js
ceciliazaragoza: Found some issues after running `npm run lint`. Revised by fixing indentation and changing method call to be: `.then((expressionData) => expressionDataHandler(expressionData, sheet, route, extension, sheetType, finalExportSheets))` so that helper method gets all necessary information.

web-client/public/js/api/grnsight-api.js
```diff
@@ -33,92 +22,36 @@ const responseExpressionData = (formData, queryURL) => {
                 crossDomain: true
             }) :
             $.getJSON(fullUrl)
-            ).done((expressionData) => {
-                resolve(expressionData);
-            }).error(console.log("Error in accessing expression database. Result may just be loading."));
+            ).done((data) => {
+                resolve(data);
+            }).error( function () {
```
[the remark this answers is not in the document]
ceciliazaragoza: I changed both of the functions to be arrow functions to make the function style consistent.

web-client/public/js/upload.js
ceciliazaragoza: I changed the line to be: ```additionalsheets = additionalsheets.filter(sheet => (sheet && -1 !== optionalAdditionalSheets.indexOf(sheet)));``` to remove the verbose statement and add more clarity, changing `x` to `sheet`.

server/dals/network-dal.js
```diff
@@ -62,16 +63,17 @@ const convertResponseToJSON = function (queryType, query, totalOutput) {
     case "NetworkSource":
         JSONOutput.sources = {};
         totalOutput.forEach(function (x) {
```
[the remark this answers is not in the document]
ceciliazaragoza: I changed the `x` to `connection`.

Issue #1003, opened 2022-12-05
Remove residual hard-coded database information


Onariaginosa:
There were some residual hard-coded database structures (Expression time points, and html code) that should get the information directly from the database. With the inclusion of the Apweiler dataset, it has become necessary to rectify this.

Onariaginosa:
Completed in PR #1004

kdahlquist:
Going to close because of PR from 2022 should have completed this work.

Issue #1002, opened 2022-12-04
Lingering style notes from #996


dondi:
A number of smaller style issues where noted in #996—this can be good onboarding work for new team members in the spring. Go to the closed pull requests list, find #996, then go over the comments from the last review, changing the code accordingly without incurring regressions

kdahlquist:
This will be the next one that @ceciliazaragoza and @akaiap will tackle this week.

ntran18:
I also found repetitive code in scripts in `generate_network.py`, `generate_new_network_version.py` and `generate_sgd_network_from_yeastract_network.py`

![image](https://github.com/dondi/GRNsight/assets/108908370/006568c2-69ee-4f0b-b6bb-bbaf272e034d)

ceciliazaragoza:
I found a lingering `/* eslint-disable max-len */` in line 1 at the top of `server/dals/protein-dal.js`, so removed the comment, fixed multiline string in the file, and ran lint and test with all tests passing.

ceciliazaragoza:
> I also found repetitive code in scripts for `generate_network.py`, `generate_new_network_version.py` and `generate_sgd_network_from_yeastract_network.py`
> 
[...]

PR #1004, opened 2022-12-05
Removed hard-coded database information and replaced them with database calls

Onariaginosa:
#1003


dondi (approved):
LGTM; the hardcode removal is notable and is worth a version bump—will do that after merging

Similar style notes here as in #996 and in similar places also, so I didn’t leave redundant notes here anymore. Going through the #996 comments as stated in #1002 should still suffice

PR #1007, opened 2022-12-06
Moved ajax and getJson function into api module

Sarronnn:
Moved "$.ajax"  and  "$.getJSON" function into api-module


dondi (approved):
Indentation seems to have shifted to two spaces but otherwise looks OK. We can deal with the indentation at some other time

Issue #1018, opened 2023-02-01
Export to Excel isn't actually exporting an unweighted graph when selected


Onariaginosa:
When exporting a workbook or demo that already has weight information in them, if you select unweighted in the first expression menu, the 2nd menu will include the `network_optimized_weights` and the `network_weights` sheets as sheet options. And if you export with those sheets and load the exported workbook it will load as a weighted graph (as it should).

To fix this consistency issue should any network weights sheets be excluded when an unweighted network is selected, or is selecting an unweighted vs weighted graph even necessary when you can select which network sheets to export. Another option is also to keep the buttons and use them to determine which network sheets are selected by default, (weighted will select all network sheets, unweighted will uncheck the network weights sheets)

![Screenshot from 2023-02-01 06-00-56](https://user-images.githubusercontent.com/21343072/216063774-72d04df1-b24f-4a27-89d4-0b4d3be7accb.png)

![Screenshot from 2023-02-01 06-01-05](https://user-images.githubusercontent.com/21343072/216063771-02c03534-e3fb-4748-acf9-cd8ae8136a49.png)

kdahlquist:
After discussion, we decided that the two modals can be collapsed into one.
- We don't need the radio buttons for "weighted" or "unweighted" because it is redundant with selecting sheets.
- So the user would select the expression data source
- Then, they would check the boxes for which sheets to export (default to all checked)
- If there is no weight data to export, that checkbox will be grayed out so you can't select them.

Onariaginosa:
Made the requested changes. 
![image](https://user-images.githubusercontent.com/21343072/217636480-6ffb73df-0ad1-4a18-ae8c-753467f1096f.png)

kdahlquist:
Confirmed in v6.0.7

PR #1019, opened 2023-02-08
#1018 consolidated export to excel

Onariaginosa:
Made the fixes requested in #1018 and fixed spacing issues from when travis was down.


dondi (approved):
LGTM—looks like we should have folks disable Prettier, or maybe install a custom _.prettierrc_ that has indent size of 4?

Issue #1035, opened 2023-04-12
Change order of Yeastmine versions in generate GRN modal


kdahlquist:
Can we change the order of the Yeastmine versions in the generate GRN modal dropdown menue so that the newest is always on top?  That way when a user goes there it will always default to the latest data.

kdahlquist:
This issue is still open.

kdahlquist:
Actually, it is fixed in beta 7.0.0, so closing now!

Issue #1028, opened 2023-03-22
protein-protein interaction UI refinements


kdahlquist:
@kdahlquist has reviewed this and edited this comment.

- [x] generate network modal should look the same as for the GRN, except change the labels
    - I think this is done
- [x] export to SIF and Excel, not graphml
    - [ ]  cell A1 of the exported Excel workbook "network" sheet; does not need to say "Cols regulators, rows targets", instead it should say "Cols protein1, rows protein2"
- [ ] <s>do not color boxes with expression data</s>  Upon reflection, this is more nuanced.  The _default_ is that there should be no node coloring.  However, if the user checks "Enable Node Coloring" either from the side panel or top menu, then it should behave as if it is a GRN.  This will be a change from the current behavior on beta 7.0.2

dondi:
These are in the `beta` branch; also, there are multiple separate issues and based on the magnitude of the issues, one or more may be large enough to become its own issue
1. Initial assignment is both @nchun2 and @ntran18 in order to pair over learning about each issue and assessing them overall
2. Third item (node coloring of PPN) may need further consideration; can consult with @kdahlquist 
3. New issues or this issue can then be assigned directly to @ntran18 (or maybe paired)

ntran18:
Because my schedule was tight last week, I couldn't meet with @nchun2 to do this issue. We plan to meet on Friday this week.

dondi:
Now that @nchun2 also has GRNsight running locally, maybe @ntran18 and @nchun2 can try for this this week.

dondi:
@ntran18 is traveling this week so it may not be feasible for @ntran18 and @nchun2 to do this together for this week; however if @nchun2’s queue reaches this, it will also be OK for her to look into it on her own and report back

ntran18:
I didn't have time to schedule a meeting with @nchun2 so I checked this by myself. 

1. Not solved: "_cell A1 of the exported Excel workbook "network" sheet; does not need to say "Cols regulators, rows targets", instead it should say "Cols protein1, rows protein2"_"

- Example from demo 1:
  
<img width="1065" alt="image" src="https://github.com/dondi/GRNsight/assets/108908370/76f222e4-e80f-41ac-9ff8-52df46b29fe7">


2. For the third checklist (node coloring). I already see that feature in the version 7.0.2 beta. However, there is different options between the top menu and the side panel. 

- Side panel:
[...]

Issue #1020, opened 2023-02-15
Investigate Yeastmine for information on Protein - Protein Interactions Mode


Onariaginosa:
I went on SGD/YeastMine to see if there was anything related to querying for protein protein interactions. In the proteins tab you could query for all proteins. The following screenshot captures all of the information you can receive for each protein result.

![image](https://user-images.githubusercontent.com/21343072/219078447-5b196978-c19f-48b3-8180-65f38d5faaf9.png)

In the interactions tab, I found that you can query based on complex and it returns the complex participants. You can restrict the query such that you return complexes that contain a given protein. The following screenshot captures all of the information you can recieve for each complex -- participant pairing (The database seems to be stored as a many to many)

![image](https://user-images.githubusercontent.com/21343072/219080216-676a3fab-bff0-4c72-844d-1e2d5605341b.png)
![image](https://user-images.githubusercontent.com/21343072/219080305-dc82d044-56a5-44a3-a799-f7492d924b46.png)

My preliminary inclination is to store the protein data in the following schema
![image](https://user-images.githubusercontent.com/21343072/219086087-00332708-b347-43f7-821d-eeb0c6c7a4d0.png)

Is there anything that should be revised?

Side note: I didn't know the language of protein - protein interactions, so I read [this article](https://www.pnas.org/doi/epdf/10.1073/pnas.93.1.13) for preliminary knowledge.

kdahlquist:
@Onariaginosa, we are actually not going to use the "complex" annotations.  There is a different query for "physical interactions" that we will use. See this example
https://yeastmine.yeastgenome.org/yeastmine/results.do?trail=%257Cquery

kdahlquist:
Retain "Gene" and "Protein" tables.  But instead of "Complex" and "Protein-Complex" we will have a table called "Physical-Interactions"

Physical Interactions table would be two-column table with protein1 and protein2

kdahlquist:
Change schema names to be more informative:
- Gene Regulatory Network schema
- Gene Expression schema
- Protein-Protein Interactions schema
- GRNSettings schema

Onariaginosa:
Here is the revised schema:
![image](https://user-images.githubusercontent.com/21343072/219141987-6a191e44-4445-4611-8934-a08baee9fd2e.png)

kdahlquist:
[...]

Issue #1011, opened 2023-01-11
Database setup in Initial Setup wiki


dondi:
We have documented our database setup instructions but they are not linked to from the Initial Setup wiki page. Let’s do that

dondi:
* @ahmad00m and @Onariaginosa to do some initial content/links
* @Sarronnn and @ntran18 can walk through the documentation and provide feedback on how to optimize for new members

Onariaginosa:
Database setup instructions are linked from the Initial Setup wiki page.

Onariaginosa:
We finally fixed the database issue! It wasn't actually a big in the code, Sequelize was rounding the timestamps from PostgreSQL, causing the timestamp to not work when querying for specific genes. I changed the network generation code to  ignore the milliseconds and that did the trick on both @Onariaginosa and @ntran18 's machines. We did a fresh network database install and it worked

kdahlquist:
As noted in issue #1014, @ntran18 is still working on documentation on how to load the database. She's going to edit the PDF from the databases class and select only the items necessary for GRNsight and add it to the wiki.

kdahlquist:
The timestamps included milliseconds, but we are truncating because we don't need this level of precision.

ntran18:
I added one line in the database setup page to go back to initial setup page and continue on the instructions. I also deleted some slide on the instruction page. 
[postgresql-setup-day.pdf](https://github.com/dondi/GRNsight/files/10917355/postgresql-setup-day.pdf)

dondi:
Given the abridged/distilled content of the PDF and some overlapping bullets in the database setup wiki, let’s see if we can skip the need for the separate PDF, so that the user can just read through things

@ahmad00m suggests separating out the commands that are only run once (i.e., for setup) from the commands that one runs every time (placeholder has been included in the https://github.com/dondi/GRNsight/wiki/Initial-Setup wiki)

ntran18:
I have modified the database setup to remove the PDF and combine overlapping bullets. I also added "Reuse Commands" section at the end of wiki page

dondi:
@nchun2 posted some notes on the database documentation in #1046 —please review with her to see if those changes should be made. After resolution, we can close this issue (we will just open a new one if a future documentation glitch is identified)

dondi:
Based on the latest findings in #1046 @nchun2 will check our current database documentation to see if there are critical/immediate changes needed; otherwise, we can close this and open future issues for documentation

[...]

PR #1039, opened 2023-04-27
v7.0.0 Protein-Protein Physical Interactions

Onariaginosa:
New Feature:
    * Protein Protein Physical Interactions Database and UI (#1028, #1020)
Enhancements:
    * Change order of Yeastmine versions in generate GRN modal (#1035)
    * Filename and display nomenclature for YeastMine-generated Networks (#983)
Documentation:
    * Database setup in Initial Setup wiki (#1011)


server/dals/network-dal.js
```diff
@@ -66,7 +66,7 @@ const convertResponseToJSON = function (queryType, query, totalOutput) {
             const timestamp = x.time_stamp;
             const source = x.source;
             const displayName = x.display_name;
-            JSONOutput.sources[`${displayName} : ${timestamp.toISOString().split("T")[0]}`] = {timestamp, source};
+            JSONOutput.sources[`${displayName}: ${timestamp.toISOString().split("T")[0]}`] = {timestamp, source};
```
dondi, line 69: Nice attention to detail 😊

web-client/public/js/generateNetwork.js
```diff
@@ -94,34 +104,58 @@ export const generateNetwork = function () {
...
+                        displayCurrentGenes();
+                    } else {
+                        alert(
+                            `${userGeneProtein}: "${searchGeneProtein}" was not found in this database. Please check for any typos and try again.`
+                        );
+                    }
+                }).catch(function (error) {
+                    console.log(error.stack);
+                    console.log(error.name);
+                    console.log(error.message);
```
dondi, line 156: Mental note for future work: we should probably place this information in our standard error/warning modal right?

web-client/public/js/generateNetwork.js
```diff
@@ -143,15 +177,26 @@ containing "-", "_", and alpha-numeric characters only`);
...
+                geneRegulation: null
+            }
         };
     // get sources from database
+        queryProteinProteinDatabase({type:"NetworkSource"}).then(function (response) {
+            grnState.customWorkbook.sources.proteinProteinInteractions = response.sources;
+        }).catch(function (error) {
+            console.log(error.stack);
+            console.log(error.name);
+            console.log(error.message);
```
dondi, line 193: More of the same…

web-client/public/js/update-app.js
```diff
@@ -477,24 +481,21 @@ const expandLayoutSidebar = () => {
...
-//     }
-// };
+const toggleLayout = (on, off) => {
+    if (!$(on).prop("checked")) {
+        $(on).prop("checked", true);
+        $(off).prop("checked", false);
+        $(`${off} span`).removeClass("glyphicon-ok");
+        $(`${on} span`).addClass("glyphicon-ok");
+    }
+};
```
dondi, line 491: Wait—was the intent here to delete the code outright? I seem to recall this was part of @ntran18’s pull request…

dondi (approved):
Generally LGTM, with a couple of non-blocking comments. The error reporting can be done for another day, but please check if @ntran18’s work is integrated correctly here (mainly whether `toggleLayout` legit needs to be restored or should actually be removed)

Thanks—great work!

web-client/public/js/update-app.js
Onariaginosa: Nope! So that function toggles between whether two items are checked in the top bar menu. While @ntran18 removed it because the Force Graph and Grid Layout were considered to be buttons that provoked an action rather than part of the graphs state. However, when switching between protein-protein mode and grn mode, this is a state change and the mode should be indicated by the check mark. So I brought back this function and used it for the network type switcher. 

In short it was deleted because it wasn't used for the bug, but I ended up using the pre-existing function in my new code

dondi: Got it, thanks for clarifying that!

Issue #1016, opened 2023-01-25
Bug: set layout to grid layout and click on any demo, layout goes back to force graph


kdahlquist:
Bug found by @ntran18: if you set the layout to grid layout and click on any demo, the layout goes back to force graph, but the menu and button still indicate that it is a grid layout.

The layout should not change.  Also, the menu/buttons should reflect the current state.

ntran18:
This bug is interesting. I checked the bug locally, and it didn't appear. It always changes to Force Graph whenever I click on a new Demo, but the bug only appears on our website.

dondi:
Upon discussion the issue is more about whether the layout commands are “sticky” state vs. just one-time actions. The checkbox in the menu is misleading and inconsistent with the buttons on the sidebar. Further, due to manual moving of nodes, the notion of “layout as state” isn’t very applicable in the first place

So the desired fix here includes:
* removing the checks from the menu
* possibly rephrasing the menu items to more clearly state that these are commands

dondi:
On further examination, looks like Grid Layout as a state because it locks out the force graph parameters—there may be a need for a broader investigation of how these settings interact

We should also look at interactions with the sidebar to see what makes sense and what does not

It might be useful to look at the client-side testing document to see what was expected in the past

@dondi will meet with @ntran18 sometime to walk through the codebase

ntran18:
I made the Grid Layout to modify the Force Graph parameters without changing the graph. The Lock Force Parameters are available for both layouts. However, I cannot fix the bug on the topbar when you click on the grid layout and choose any demo, it still shows Grid Layout. I don't know which functions correspond to that feature.

ntran18:
I fixed the bug on the topbar when you click on the grid layout and choose any demo, it will show a Force Graph. However, talking with @dondi, he suggested removing the check of the force graph and grid layout to be matched with the left menu.

ntran18:
I tried to remove the "checked" property of Force Graph and Grid Layout on the topbar. However, even if I delete the property, the check doesn't remove. I figure out it has to do something with "glyphicon-ok" class. If I remove that class, when you click on Demo, it doesn't automatically display the graph, but you have to click on either "Force Graph" or "Grid Layout" then it will display the graph.

dondi:
@ntran18 will check for errors in the log as a next step

ntran18:
[...]

PR #1045, opened 2023-09-04
Fix bug #1016

ntran18:


web-client/public/js/update-app.js
```diff
@@ -491,7 +500,7 @@ const toggleLayout = (on, off) => {
 };
 
 const updatetoForceGraph = () => {
-    $(LOCK_SLIDERS_BUTTON).removeAttr("disabled");
+    // $(LOCK_SLIDERS_BUTTON).removeAttr("disabled");
     console.log("Update to Force graph");
```
dondi, line 504: If we really don’t need the commented-out line anymore, it’s OK to delete it outright (GitHub has it anyway in the history)

Typically we also reserve console logs for debugging. I notice that this will leave the function empty; I think that’s OK, just in case we need to add code in the future when `updateToForceGraph` takes place

dondi (commented):
Looks good, just needs some minor cleanup

ntran18:
Code is clean!

dondi (approved):
LGTM

Issue #1033, opened 2023-04-12
Can only change dataset from bottom option, not top


Sarronnn:
<img width="703" alt="Screenshot 2023-04-11 at 11 38 28 PM" src="https://user-images.githubusercontent.com/89826597/231372175-72a4515d-9eac-4927-811d-093146ac321e.png">
<img width="703" alt="Screenshot 2023-04-11 at 11 39 24 PM" src="https://user-images.githubusercontent.com/89826597/231372678-a78614a9-5c42-4e72-ad7e-8e1aec3b41f1.png">

kdahlquist:
This still has some problems on Beta 7.0.2
- If I select the top and bottom datasets from the side panel, the changes are not reflected in the top menu.  It does weird things
    - If I select the top dataset first from the side panel, then the right top dataset is reflected in the menu, but not the bottom dataset.
    - If I select the bottom dataset first from the side panel, then neither the top dataset or bottom dataset are checked off in the menu.
- Strange things happen when selecting datasets from the top menu.

dondi:
Fundamentally, the core issue is that the selected dataset should be centralized to an internal state value: whenever any control or UI element changes this value, that change should lead to the same code regardless of the source. Doing so will ensure that the internal behavior is always the same independent of whether the change was made on the sidebar or the menu

dondi:
@ntran18 as a PR for this and that is what the “review requested” tag indicates; however she documents that a subsequent bug appears which seems to be more related to the dataset than the UI

@ntran18 will extract this as a separate issue, likely to be addressed after this one is merged into `beta`

dondi:
The UI-level bug appears to be fixed but the fix is disrupted by a new issue that appears to occur at the data level, especially for expression data sets that come from the database and are not read from the network directly. Will consider this item as closed but @ntran18 will write up and investigate the missing data portion as a new issue

The new issue can reference this one in case we need to review how it was discovered

Issue #882, opened 2020-10-02
Integrade code formatting specifications into the GRNsight repo


dondi:
The latest code formatters are getting better at formatting code in a reliable and consistent manner. Some implications in case we adopt this for GRNsight:
1. We will want to configure the code formatter in a manner that use a style that the team agrees upon
2. There will be a “formatting-only” transition phase where the code formatter may need to make a wide array of edits in order to sync up the code to the latest style

Formatter for consideration: https://prettier.io

kdahlquist:
Set up repo so that independently of the code editor being used.  Might need to reckon with 4 spaces of indenting vs new standard for javascript which is 2 spaces of indenting.

dondi:
Prettier has established itself as the de facto code-formatter for many repos, so we will just go with that. We will need a _.prettierrc_ file (or similar mechanism) on the repo, and make sure that our editors are configured to use it

If we change format, step (2) in the original description may still need to take place

dondi:
With the merging of #1148 into `beta`, we now need to observe interactions between Prettier and ESLint and can finalize configurations based on that

akaiap:
## Prettier Integration
- Created .prettierrc.js aligned with Grnsight's style:
4-space indentation
Double quotes for strings
Consistent semicolons
Standardized object formatting

- Then Integrated ESLint and Prettier to avoid style conflicts.
Added prettier dependencies `npm install --save-dev prettier eslint-config-prettier eslint-plugin-prettier`
Really cool --> `eslint-config-prettier`: Disables ESLint rules that conflict with Prettier
`eslint-plugin-prettier:` Runs Prettier as an ESLint rule

- Updated ESLint Config for Prettier Integration
- Add formatting scripts to packjson
- Added to the scripts (formatting)
```
"scripts": {
  "format": "prettier --write \"**/*.{js,jsx,json,md}\"",
  "format:check": "prettier --check \"**/*.{js,jsx,json,md}\"",
  "lint": "eslint \"**/*.js\"",
  "lint:fix": "eslint --fix \"**/*.js\""
}
```


### Formatting 
To format use the command `npm run format`
To check for formatting issues use the command `npm run format:check`



Lastly I ran npm run lint and received errors regarding the depricated **.eslintignore** file and unused eslint-disable directives. 
- Updated eslint.config.js file to include all ignored files and removed eslintignore file
- Cleaned up unused eslint-disable directives using `npx eslint --fix .`

kdahlquist:
[...]

PR #1051, opened 2023-09-24
Maika 1033

ntran18:
Fixing bottom dataset

### DIFF
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -1010,12 +1010,20 @@ export var drawGraph = function (workbook) {
                 .append("g")
                 .selectAll(".coloring")
                 .data(function () {
-                    if (grnState.workbook.expression[dataset].data[p.name]) {
-                        const result = getExpressionData(p.name, dataset, average);
-                        timePoints = result.timePoints;
-                        return result.data;
-                    } else {
-                        return 0;
+                    if(grnState.workbook.expression[dataset]) {
+                        if (
+                            grnState.workbook.expression[dataset].data[p.name]
+                        ) {
+                            const result = getExpressionData(
+                                p.name,
+                                dataset,
+                                average
+                            );
+                            timePoints = result.timePoints;
+                            return result.data;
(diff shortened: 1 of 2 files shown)


ntran18:
I was able to make the selection in the bottom data menu. However, I'm thinking there is problem with database. One notice problem is that in graph.js, `grnState.workbook.expression[dataset]` not always contains data, it can be a null object. I fixed this one, however, sometimes I choose a random expression, the error shows there is no value `p.length` but doesn't show which files it refer to. I wasn't able to look more on this problem this week because I didn't have time. Will take a look at it again next week.

dondi (commented):
Looks like this PR has linter issues—for direct reference, you can look at the Travis report: https://app.travis-ci.com/github/dondi/GRNsight/builds/266145719

You should also be able to do `npm run lint` on the command line in order to see these messages

dondi:
As a side note, these linter errors will be more avoidable if we configure an autoformatter that complies with the current style checks made by the `npm run lint` command. We have an issue for this (#882) which we can bump in priority when the time is right

ntran18:
I expected some errors would happen. However, these errors are out of my scope of understand so I might need to meet with @dondi to discuss about these problems. I also think this is a different issue, related to database, not the UI.  

Problems: When you click in some expressions, especially the one contains ACE node, it would not reflect on the UI, meaning the check doesn't change to your selection. This is because the program can't load the expression. As I mentioned before, this is because there is no value `p.length` (I experienced some expressions, and it happens mostly for ACE node, not sure about the other nodes). However, if you click on that selection again, it works, but I don't know if it is correct or not.

1. Lint errors 
image
image
image
image
--> From what I understand, I think either the version of database I currently have has different expression sheet name. A lot of errors were wrong name, or worksheet not detected. 

2. Example of error from console
image

dondi (approved):
LGTM from a code perspective; but as noted in #1033 and in the comments, merging these changes appears to reveal an internal error due to missing data

Issue #1054, opened 2023-10-11
Modify the name of cell A1 of exporting Excel workbook 


ntran18:
Problem: Cell A1 of the exported Excel workbook "network" sheet does not need to say "Cols regulators, rows targets", instead it should say "Cols protein1, rows protein2"

_Example from demo 1:_
  
<img width="1065" alt="image" src="https://github.com/dondi/GRNsight/assets/108908370/76f222e4-e80f-41ac-9ff8-52df46b29fe7">

--> Cell A1 stills have the name Cols regulators, row targets

nchun2:
I retried this using some chosen proteins instead of one of the demos, and the name of cell A1 was still an issue. 
![Screenshot (117)](https://github.com/dondi/GRNsight/assets/122498296/07c19fcc-e57a-4546-aa1f-8e9bdb949840)

dondi:
@nchun2’s report was done from `beta` but at the time, `beta` did not have the PR yet; will need to revisit after merging

In order to try something out before it is merged to `beta`, one can switch to the git branch that contains the fix(es) locally. Then the local version of the app can be tested. For example, specifically with this issue, the command would be:

    git switch maika-1054-new

Once finished, `switch` can be invoked again to go back to whichever original branch you were on, e.g.:

    git switch beta

dondi:
After noting the test suite failures and reviewing the PR, we noted that the change in heading should not be wholesale but instead conditional on the type of network being exported. For GRNs, we keep the existing heading; the change in heading should only be done for protein-protein networks
* Look into the code base to see how each network type is differentiated
* Restructure the export code so that this network type can determine the exact exported content

Extending the test suite to include protein-protein-specific test files would also be desirable but their absence doesn’t break the build, only limiting the test coverage. It will be separated out into its own issue due to the scale involved with recreating the test suite completely

ntran18:
I haven't had time this week to do this issue, due to busy week. I will try to do this issue this week

ntran18:
I made a commit for this issue. The server is the one doing the export. Therefore, I change the code in the server folder. However, I wasn't able to change the test. I need to talk with Dr. Dondi for help. I also linked the commit below. 

f19f1dc3e0b5e8868764698a56f01623e54789c9

ntran18:
Since the code was merged and no problem was found currently, Is it reasonable to close this issue? @dondi  @kdahlquist

kdahlquist:
[...]

PR #1057, opened 2023-10-13
Modify the name of cell A1 of exporting Excel workbook #1054

ntran18:


ntran18:
This is wrong, created a new branch

Issue #1050, opened 2023-09-19
Search (magnifying glass) button is too small in Load from Database modal


kdahlquist:
The Search (magnifying glass) button is too small in Load from Database modal.  See screenshot below:

![image](https://github.com/dondi/GRNsight/assets/6463867/b49edcdc-b639-48a6-82c8-200ad0749627)

kdahlquist:
Assigning this priority 0 for @ntran18

dondi:
Deployed as beta 7.0.6; upon review we noticed that the border was missing. In addition we were able to track down the specific regression that occurred from `master`, which is that the `.glyphicon` style rule was changed. @ntran18’s PR made a different set of changes

@ntran18 will compare the `master` CSS with the new `beta` to see whether we can restore the CSS more directly

kdahlquist:
confirmed fixed in beta 7.0.7

PR #1064, opened 2023-11-14
Fixing search symbol #1050

ntran18:
Modifying glyphicon class to match with master class, and fix the spacing of Force Graph and Grid Layer option. 

* On the previous PR, one of the issue was removing the check of force graph and grid layout option, thus I modified the glyphicon class. Unfortunately, it modified the css for the search icon too.


dondi (approved):
LGTM

Issue #1072, opened 2023-11-21
Review PPI database-loading script


dondi:
@ntran18 has run into some issues with the PPI loading code/documentation in `beta` (script doesn’t complete; tables do not get populated). We should compare this to what @nchun2 is seeing and in general troubleshoot what might be going on

This may include temporarily (or permanently) revising the scripts to produce better feedback while running so that the user can take note of any progress/issues

ntran18:
Actually `generate_protein_network.py` would not populate any data into the database tables. It will create 5 different files that will later be used to populate data. Therefore, the tables don't contain any data. 

After that, `python3 loader.py | psql postgresql://localhost/postgres` this script actually populates data into the database. It might stop because of duplicate protein `Aad6p` (might want to bold **1** from `Aad16p` to have a better distinguish between `Aad6p` and `Aad16p`.  

After fixing the duplicate protein, if we run the script again, there might be some data already in the table. Thus, having more errors on duplicate keys. My solution was to add Truncate tables to remove existing data, and populate data again.

dondi:
We need to look at the current deployed beta to make sure things are functional. @ntran18 and @nchun2 can re-run the scripts to test functionality on a pre-existing database; @ceciliazaragoza and @akaiap will be running these scripts for the for the first time as a part of #1078 and can also report back on functionality

kdahlquist:
@ceciliazaragoza was able to get the PPI database running OK.  But @akaiap wasn't.  She'll see @dondi in office hours to troubleshoot, so we'll keep this open for a little while longer.

dondi:
@akaiap and @dondi worked on this last Thursday and got the issues resolved

PR #1073, opened 2023-11-28
Adding changes for #1072 issue

ntran18:
Add more feedback for genereate_protein_network.py and add a truncate function for loader.py to make sure there is no duplicate data when running the loader script multiple times.

### DIFF
--- a/database/protein-protein-database/scripts/loader.py
+++ b/database/protein-protein-database/scripts/loader.py
@@ -16,7 +16,7 @@ directly into a database command line utility such as `psql`.
 This function Loads Protein-Protein Network Data Sources into the database
 """
 def LOAD_SOURCES():
-    print('COPY protein_protein_interactions.source (time_stamp, source, display_name) FROM stdin;')
+    print('COPY protein_protein_interactions (time_stamp, source, display_name) FROM stdin;')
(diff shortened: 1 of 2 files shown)


database/protein-protein-database/scripts/loader.py
```diff
@@ -16,7 +16,7 @@
 This function Loads Protein-Protein Network Data Sources into the database
 """
 def LOAD_SOURCES():
-    print('COPY protein_protein_interactions.source (time_stamp, source, display_name) FROM stdin;')
```
dondi, line 19: I think this line may need to stay unchanged; existence of `source` table confirmed in beta GRNsight database

dondi (commented):
Please look into the `source` `COPY` line; may need to redo the loading in order to make sure that the entire script runs correctly, plus test on GRNsight itself to ensure that there are no database errors—ideally, also search the code for instances of `source` in order to determine what use cases will access the `source` table

When all is done, please review the database Markdown documentation and include any clarifications/revisions pertaining to the script change. In particular, make sure to mention that the script truncates existing data to ensure a clean load without duplicates

database/protein-protein-database/scripts/loader.py
ntran18: Fixed!

Issue #1065, opened 2023-11-14
Add PPI demo


dondi:
For parity with GRNs, we need to add PPI demos to the app. @kdahlquist will generate the file and this can be added to the app in the same way that the existing demos are added.

dondi:
We’ll start with an export based on the database then move on to Dr. Dabir’s data when ready

kdahlquist:
[Demo #5 PPI Network (Yeastmine - SGD 2023-03-29; 18 proteins, 81 edges).xlsx](https://github.com/dondi/GRNsight/files/13490802/Demo.5.PPI.Network.Yeastmine.-.SGD.2023-03-29.18.proteins.81.edges.xlsx)

I created this network from the database.  It is all the proteins that interact with Aim32p, which is the protein that Dr. Dabir studies.

I exported it from GRNsight and then modified the filename.

It has only a "network" sheet.  I changed the text in cell A1 to read "protein 1/protein 2".

It loads back into GRNsight OK, but currently, it loads as a GRN, not PPI because there's nothing to tell GRNsight that it is a PPI network.

In doing this, I think I uncovered a bug, but will report in a different issue.

kdahlquist:
I'm open to shortening the filename to something like:

Demo #5 PPI (18 proteins, 81 edges)

"Network" is probably redundant.  If we shorten it this way, we lose the provenance of the data, but that could be put into the documentation.  This could be better due to readability in the dropdown menu.

dondi:
File is ready to go—we will use the content of cell A1 to determine the Network Mode to switch to, as noted in #1066. Label the demo as noted in the comment above

Demos are encoded directly as JSON objects; the easiest way to convert is probably to temporarily add a `console.log` to the code which displays the converted network object. When opening the file, this network JSON will appear in the Developer Console and can then be transcribed to the source code in _demo-workbooks.js_

As noted, we will use cell A1 of the Network sheet to determine the Network Mode. Further, we will need to adjust the warning to adapt to these changes

kdahlquist:
Verified present in beta 7.0.8

PR #1075, opened 2023-11-30
Adding PPI demo #1065

ntran18:
Currently, there is no PPI demo. Therefore, adding another demo that demonstrates protein-protein physical interaction. The demo also changes the network mode to the correct type.


dondi (approved):
LGTM! Great step toward deeper PPI support on GRNsight 👌🏽

PR #1105, opened 2024-03-22
Update network and protein protein database

ntran18:
Adding the scripts for updating protein protein database and clean code


ntran18:
I might need help from @akaiap @ceciliazaragoza and @nchun2 to test those scripts out first before running in the production database.

database/network-database/scripts/loader.py
```diff
@@ -74,6 +79,11 @@ def LOAD_NETWORK():
...
-LOAD_GENES()
-LOAD_NETWORK()
+# load_sources()
+Utils.load_sources(Constants.NETWORK_DATA_SOURCE, Constants.DATABASE_NAMESPACE)
+
+# load_genes()
+Utils.load_genes(Constants.GENE_SOURCE, Constants.DATABASE_NAMESPACE, False)
+
+# load_network()
+Utils.load_network(Constants.NETWORK_SOURCE, Constants.DATABASE_NAMESPACE, False)
```
dondi, line 89: It’s typically an anti-pattern to have unnamed Boolean arguments because other code readers will not clearly know what is being set to true or false; let’s make this a named parameter

Name tbd pending inclusion of _utils_ module in the PR…

dondi (commented):
Let’s make sure to include _utils_ module in the PR, then we can act on the other comment on Boolean parameters

database/network-database/scripts/loader.py
dondi, line 89: With _utils_ now included, we can see that the argument is `is_protein` so we want these to be:
```suggestion
Utils.load_genes(Constants.GENE_SOURCE, Constants.DATABASE_NAMESPACE, is_protein=False)

# load_network()
Utils.load_network(Constants.NETWORK_SOURCE, Constants.DATABASE_NAMESPACE, is_protein=False)
```

dondi (commented):
Cool, let’s go ahead and name all instances of Boolean parameters

Issue #1106, opened 2024-03-26
ERT1 not in network database


kdahlquist:
We are going to start using GRNsight in BIOL 367 this week.  I was writing the protocol and had occasion to look up "ERT1" in the "Load from database" GRN.  It was not found.  Can we check to see that it is not in the network database?  Also, will it be there in the 2024 update?

I then tried to look it up by its systematic name "YBR239C" and got an error saying that it did not conform to the naming convention it was expecting.  But I was unable to reproduce this.

dondi:
Let’s audit this—we can look at both the database directly and also the downloaded original files; the latter can be supplied by @ntran18

dondi:
A quick look at our archived data indicates that `ERT1` was in the original database load from fall 2021 but appears to have been dropped in spring 2022:
```
postgres=> \dn
             List of schemas
             Name             |  Owner   
------------------------------+----------
 fall2021                     | postgres
 gene_expression              | postgres
 gene_regulatory_network      | postgres
 protein_protein_interactions | postgres
 public                       | postgres
 settings                     | postgres
 spring2022_network           | postgres
(7 rows)

postgres=> set search_path=fall2021;
SET
postgres=> \dt
                 List of relations
  Schema  |        Name         | Type  |  Owner   
----------+---------------------+-------+----------
 fall2021 | degradation_rate    | table | postgres
 fall2021 | expression          | table | postgres
 fall2021 | expression_metadata | table | postgres
 fall2021 | gene                | table | postgres
 fall2021 | production_rate     | table | postgres
 fall2021 | ref                 | table | postgres
(6 rows)

postgres=> \d gene
                        Table "fall2021.gene"
     Column      |       Type        | Collation | Nullable | Default 
-----------------+-------------------+-----------+----------+---------
 gene_id         | character varying |           | not null | 
 display_gene_id | character varying |           |          | 
 species         | character varying |           |          | 
 taxon_id        | character varying |           | not null | 
Indexes:
    "gene_pkey" PRIMARY KEY, btree (gene_id, taxon_id)
Referenced by:
    TABLE "degradation_rate" CONSTRAINT "degradation_rate_gene_id_fkey" FOREIGN KEY (gene_id, taxon_id) REFERENCES gene(gene_id, taxon_id)
    TABLE "expression" CONSTRAINT "expression_gene_id_fkey" FOREIGN KEY (gene_id, taxon_id) REFERENCES gene(gene_id, taxon_id)
[...]

PR #1111, opened 2024-04-09
#1106 fixing off-by-one column issue. 

ntran18:
Previously the first column was cols regulator/rows targetETR1. Fix by adding \t to the header.


dondi (approved):
LGTM—we can merge but this is a database script, so in order to deploy, we would actually need to update the database

PR #1113, opened 2024-04-23
#1106 Writing Scripts for Updating Database and Refactor Code

ntran18:
- Refactor Code into reusable code in utils.py
- Create an all in one `loader.py` to load all data into expression, network, and protein-protein interactions database
- Create an all in one `loader-update.py` to load all updates into network and protein-protein interactions database
- Create a union gene table for all expression, network, and protein-protein interactions database


dondi (commented):
Generally looks good, with one historical note and I think we have an opportunity to address it: when we first started populating our database, we had only one network type (GRN), so many names in this code pertain to just “network.” But now that we also support PPIs, the proper naming scheme is no longer “network” vs. “PPI” but “GRN” vs. “PPI.” So please scan for any generic “network” names and adjust to indicate “GRN” instead. Thanks!

dondi (approved):
LGTM

PR #1112, opened 2024-04-18
Issue #1083 When check restrict graph to viewport the D-pad and zoom slider do not disappear and graph cannot exceed viewport

ceciliazaragoza:
When restrict graph to viewport all *nodes*, not edges, are restricted to viewport. Graph does not exceed viewport even when dragged, zoomed, or moved with Dpad.

PR #1114, opened 2024-04-23
Fixing PR #1112 to Restrict D-pad

ceciliazaragoza:
After going through PR checklist for PR #1112 , realized D-pad was not being restricted properly. I had to adjust one line of the code in `move` function because I did not call the viewportBoundsMoveDrag function and instead just had an if statement that said `if ((graphZoom, moveWidth, moveHeight))`. Updated code to `if (viewportBoundsMoveDrag(graphZoom, moveWidth, moveHeight))` and now D-pad is being restricted as expected.


dondi (approved):
LGTM. TIL about `(expr1, expr2, expr3)` 😅

Issue #1077, opened 2023-12-05
Release checklist for v7.1


dondi:
- [x] In the days leading up to the release, all open issues that need to be resolved before the release will be given the label "for next release".
- [x] Any new or changed client-side features need to be documented on the [Client Side Testing Document](https://github.com/dondi/GRNsight/wiki/Client-Side-Testing-Document) page.
    - [x] the [feature JSON files](https://github.com/dondi/GRNsight-archive/tree/main/documents/developer_documents/testing_script_generator) file should be updated
    - [x] the new feature also needs to be added to the [Pull Request Checklist](https://github.com/dondi/GRNsight/wiki/Pull-Request-Checklist)
- [x] Test the candidate build (typically the `beta` branch)
- [x] Merge the branch(es) under development (typically `beta`) into `master`
    - [x] Resolve any version control conflicts
    - [x] Test the final merged version
    - [x] Perform hotfixes on `master` as needed
- [x] Update README.md at top-level directory of master branch, if necessary
    - [x] Update Travis and Coveralls badges to reference the master branch
- [x] Tag the final, tested version as `v#.#.#` (all three numbers are required, even if `0`; see _npm_ step below)
    - [x] Update version numbers in _package.json_
    - [x] Remember that the tag needs to be pushed to GitHub: `git push origin <tag-name>`
- [x] Create draft release item on Github repository (name it exactly like the tag)
    - [x] Make release notes: a list of new features/bug fixes with issue numbers covered by this release.  Note that this at least will be the ones labeled "for next release", with possible others.
- [x] Publish GitHub release item
- [x] Publish release to GRNsight home page
    - [x] Refresh (pull) the `master` clone on _grnsight.lmucs.org_
    - [x] Update the version number and modification date on `gh-pages`
    - [x] Update the About page with release notes (GitHub release item without issue numbers)
        - [x] Update the browser version compatibility statement with the latest version of Chrome and Firefox for which the new GRNsight version has been tested.
        - [x] Update modification date
    - [x] Add news item about the release to the News page
        - [x] Update modification date
        - [x] Link news item to the corresponding release notes in the About page
    - [x] Update Documentation page with any new features or updates to old features
- [x] Increment beta version
    - [x] Merge released `master` with `beta`, especially if hotfixes/merges were made on `master`
    - [x] Update version number and modified date on `gh-pages`
    - [x] Update Travis and Coveralls badges to reference the beta branch
[...]

PR #1119, opened 2024-10-02
Restrict Graph to Viewport Hotfix: Navbar 

ceciliazaragoza:
Hotfix for 7.1 release #1077. Graph is updated to fit within bounds when zoomed into the graph when Restrict Graph to Viewport is selected in navbar. Do this by calling tick() when the restrict graph to viewport option is clicked in navbar, just like how tick is called when click Restrict Graph to Viewport option in sidebar.

### DIFF
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -456,20 +456,18 @@ export var drawGraph = function (workbook) {
             center();
         }
         updateAppBasedOnZoomValue(); // Update zoom value within bounds
+        // Refresh the graph so that nodes and paths are adjusted to fit in viewport
+        tick();
     };
 
     d3.select("#restrict-graph-to-viewport").on("click", function () {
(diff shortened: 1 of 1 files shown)


dondi (commented):
Great find!

One side thought—should invocation of `tick` be a standard action for _all_ cases when `restrictGraphToViewport` is called? Because if so, maybe we can put the `tick()` call inside `restrictGraphToViewport` instead?

ceciliazaragoza:
I placed `tick()` call at the end of restrictGraphToViewport method and it works!

dondi (approved):
Nice! Anytime we can identify repeated code that turns out to be an inherent part of an operation, we same on code lines and are less error-prone (in case someone performs the operation elsewhere but forgets to add that essential repetition)

dondi:
@ceciliazaragoza this has been deployed to production—you can verify if the fix now shows up on our website

PR #1121, opened 2024-10-06
Adding release notes to About page and News page, also LMU Symposium 2024

ntran18:


about.html
```diff
@@ -65,6 +65,54 @@
                         <div id="releaseContent" class="panel-collapse collapse">
                             <div class="panel-body">
                                 <div class="panel-group">
+
+                                    <div class="panel panel-default"></div>
+                                        <div class="panel-heading">
+                                            <p class="panel-title aboutHeaders">
+                                                <a href="#v7_1_0Content" data-toggle="collapse">
+                                                    GRNsight v7.1.0
```
dondi, line 73: We went to 7.1.1 right? (Bumped due to Cecilia’s hotfix)

about.html
```diff
@@ -65,6 +65,54 @@
...
+                                                </a>
+                                            </p>
+                                        </div>
+                                        
+                                        <div id="v7_1_0Content" class="panel-collapse collapse">
+                                            <div class="panel-body">
+                                                GRNsight v7_1_0 includes the following new features and bug fixes:
```
dondi, line 80: I think we need dots here rather than underscores… (plus the version number)

```suggestion
                                                GRNsight v7.1.1 includes the following new features and bug fixes:
```

dondi (commented):
Generally looks ok except for the version number; plus a small punctuation change

about.html
ntran18: Done

dondi (approved):
LGTM

PR #1124, opened 2024-10-10
#1106 Allowing query for any gene in gene table even if it's not in network table

ntran18:
After merging both gene tables from PPI and GRN, we still can't query ERT1 because ERT1 is not in network table. In order to fix that, change the database query to only query in the gene table.


dondi (approved):
LGTM

Issue #1080, opened 2024-01-23
node coloring glitch when loading network from database


kdahlquist:
Do this in beta 7.0.8:
- load a GRN from the database
- uncheck node coloring
- recheck nod coloring--the checkbox gets checked and the node coloring comes back in the display, but the options don't show on the side panel

This bug is also seen in production 6.0.7

ceciliazaragoza:
I tried this on beta 7.0.13 and this behavior no longer seems to be occurring. However, I do see that this is an issue when checking and unchecking node coloring for PPIs. @kdahlquist Should I disable the ability to check or uncheck node coloring for PPIs?

kdahlquist:
I think we should keep the ability to check/uncheck node coloring for PPIs.  People might have protein-expression data that they want to visualize or they might want to overlay the mRNA expression data on the PPI network.

ceciliazaragoza:
I mentioned that the workbook was the issue during the meeting, but after looking over the code again, that was not the issue. There was a check for grnState.mode === NETWORK_PPI_MODE where grnState.nodeColoring.nodeColoringEnabled was set to false. I was able to get the Node Coloring Menu to show with PPIs, but should the options below "Enable Node Coloring" still display when viewing a PPI?
<img width="374" alt="image" src="https://github.com/dondi/GRNsight/assets/105010308/0f2d39de-ccbe-4339-90bb-46d7306302c8">

kdahlquist:
Yes, those options should show because they enable the user to customize how the nodes are colored.

dondi:
We clarified at the 4/23 meeting that we can go ahead and have the warning appear each time. A “don’t show this again” option might be something for the future

ceciliazaragoza:
I am able to display a warning each time a PPI is displayed and node coloring is enabled. I am thinking of making the default behavior for grnState to enable node coloring, but is there any case where nodeColoring should be disabled where we wouldn't want to enable node coloring by default?

kdahlquist:
Here are the use cases:
* User uploads their own PPI network with their own expression data in the workbook; node-coloring should be enabled.
* User uploads their own PPI network with no expression data in the workbook; node-coloring should default to disabled
* User loads a PPI network from the back-end database; node-coloring should default to disabled.  If they then enable node coloring to use our expression database, that is when the warning should pop up.

[...]

PR #1125, opened 2024-10-16
#1080: Display warning when node coloring enabled for PPIs

ceciliazaragoza:
Issue #1080. When user viewing and PPI and node coloring enabled in navbar or sidebar, warning modal displays to let users know they are displaying mRNA-level expression data on a PPI network.

### DIFF
--- a/web-client/public/js/update-app.js
+++ b/web-client/public/js/update-app.js
@@ -548,13 +548,11 @@ const checkWorkbookModeSettings = () => {
     if (grnState.mode === NETWORK_PPI_MODE) {
         grnState.nodeColoring.nodeColoringEnabled = false;
         grnState.colorOptimal = false;
-        // showNodeColoringMenus();
         hideEdgeWeightOptions();
         updateModeViews();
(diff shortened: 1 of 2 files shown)


web-client/public/js/update-app.js
```diff
@@ -933,10 +924,13 @@ export const updateApp = grnState => {
                 setTimeout(() => updaters.renderNodeColoring(), 250);
 
             }
+            if (grnState.mode === NETWORK_PPI_MODE) {
+              displayPPINodeColorWarning();
+            }
         }
+        refreshApp();
```
dondi, line 931: As you might have observed, “refresh” calls are not expressly needed in frameworks like React and SwiftUI—but they still do happen. It is, in fact, part of what these frameworks offer that certain operations are done automatically and at the right moments, rather than leaving it to the programmer to make it happen

dondi (approved):
Looks good from a code-review perspective—for changes like these, the code review mainly looks at style and localized approaches instead of full functionality. We’ll definitely still want to put this through its paces to make sure that the fix truly does what it’s intended

Issue #1055, opened 2023-10-11
Node Coloring Coordination between top menu and side panel


ntran18:
I already see that feature in the version 7.0.2 beta. However, there is different options between the top menu and the side panel. 

- Side panel:
<img width="207" alt="image" src="https://github.com/dondi/GRNsight/assets/108908370/9807903a-d6ac-415e-9121-c1acbeced713">

- Top menu:
<img width="350" alt="image" src="https://github.com/dondi/GRNsight/assets/108908370/acdd85d9-d0c4-4390-bbd5-5bd79ff86af8">

--> Users can still select top data or bottom data from the top menu, not from the side panel.

nchun2:
I tried to replicate this issue on beta, and it allows me to select node coloring from both the top and side menus, but only the side menu seems to act on the change. When I click on either option, it consistently moves the nodes further away from each other. I don't know if this is supposed to happen or if it actually matters, but I just wanted to note that. 
![Screenshot (114)](https://github.com/dondi/GRNsight/assets/122498296/c9edea75-0e16-42b0-9108-44e07c1c65ea)
Clicking the option from the top menu moves the nodes further away from each other, but it does not change their color. After it is clicked it also does not have the check next to it to indicate that it has been clicked. 
![Screenshot (115)](https://github.com/dondi/GRNsight/assets/122498296/583cc3f1-39b2-4825-8230-0f9cb160e520)
![Screenshot (116)](https://github.com/dondi/GRNsight/assets/122498296/de97fc69-b17d-4c0f-98d6-a6764489b6ca)
For the side menu, it does keep the check and it changes the color of the nodes from white to grey.

ntran18:
I think this problem can be closed. This issue was solved in one of the other issues (#1033 probably). The issue that @nchun2 mentioned meaning data pulled from protein-protein interactions allowed node coloring, which shouldn't be. The option is disabled in beta branch currently. 
![image](https://github.com/dondi/GRNsight/assets/108908370/658fcebf-2f3a-49ed-9518-9b15a475a317)

kdahlquist:
In beta 7.0.11, I have verified that node coloring is disabled (completely grayed out) if the first network you choose is a PPI network.  However, if you open a GRN first and then open a PPI, the Node coloring top menu is not grayed out and you can select options there, even though they do not act on the graph.  The Node side panel will allow you to check the "Enable node coloring" box.  If you check the box, it will color the nodes a uniform light gray color.

[...]

PR #1128, opened 2024-10-23
#1055 Make the node coloring menu for top bar and side bar to be cons…

ntran18:
…istent, and fix lint errors

### DIFF
--- a/web-client/public/js/update-app.js
+++ b/web-client/public/js/update-app.js
@@ -51,7 +51,6 @@ import {
   NODE_COLORING_MENU_CLASS,
   NODE_COLORING_NAVBAR_OPTIONS,
   NODE_COLORING_SIDEBAR_BODY,
-  NODE_COLORING_SIDEBAR_PANEL,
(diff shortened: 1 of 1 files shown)


web-client/public/js/update-app.js
```diff
@@ -925,7 +922,7 @@ export const updateApp = grnState => {
 
             }
             if (grnState.mode === NETWORK_PPI_MODE) {
-              displayPPINodeColorWarning();
+                displayPPINodeColorWarning();
```
dondi, line 925: Wow—interesting that Prettier hasn’t fixed these on its own 🤔

web-client/public/js/update-app.js
```diff
@@ -972,7 +969,7 @@ export const updateApp = grnState => {
 
     if (grnState.nodeColoring.showMenu) {
         showNodeColoringMenus();
-    } 
+    }
```
dondi, line 972: This too!

dondi (approved):
LGTM—funny about those formatting-only changes: wonder why Prettier hadn’t done those yet

Issue #1093, opened 2024-02-13
D-pad, zoom slider, and cursor drag do not move graph after export to SVG


ceciliazaragoza:
After exporting a graph to SVG, the D-pad can be clicked and the zoom slider can be moved. However, the cursor hand does not close, and the graph does not move and remains at the same zoom and position as when exported. However, if you reload the graph, choose a demo, or open a file, the cursor grab and zoom slider move the graph again. This issue is in the master branch as well.

dondi:
Suspicion here is that the export routine manipulates the SVG elements in-place to make them export ready but those changes then impact the web app. Ideally the SVG export operates on a deep copy of the SVG tree so that it doesn’t leave any unintended side effects

ceciliazaragoza:
This seems to be the same issue as in #1059.

dondi:
To elaborate further, #1059 may be the root cause of this bug; that is currently assigned to @ntran18. @ceciliazaragoza will monitor both issues and update accordingly if it does turn out that they are bound together

kdahlquist:
Verified fixed in beta 7.1.7

Issue #1059, opened 2023-10-17
Export to SVG modifies the graph itself


dondi:
Choosing _Export to SVG_ results in a visual change to the displayed graph. It appears that prepping the exported data “leaks” some changes to the data in memory.

ntran18:
I can't reproduce this error.

dondi:
Upon testing during the 10/23 meeting, we noted that the root cause of the error appears to be browser-independent but it manifests in different ways:
* In Chrome and Firefox, visual changes do not occur but #1093 does take place
* In Safari, the same root cause (we believe) does change the visual behavior, but does not break dragging (i.e., not #1093)

dondi:
Example in Safari; before export:
<img width="481" alt="image" src="https://github.com/user-attachments/assets/21a00608-e83e-458d-aaa0-cdabaae237b7">

After export:
<img width="530" alt="image" src="https://github.com/user-attachments/assets/35643a32-9784-46c9-ad43-84122df0e6e9">

ntran18:
When I ran locally, I didn't encounter the issue in Safari, but I couldn't drag the nodes. I fixed it by creating a copy of the current graph and exporting it. Hopefully, this would fix the issue in Safari.

kdahlquist:
confirmed fixed in 7.1.17, closing

PR #1132, opened 2024-10-25
 #1059 and #1093 Fixed issue for exporting SVG modify the graph itself. 

ntran18:
Fixing both issues #1059 and #1093


dondi (approved):
LGTM

Issue #1098, opened 2024-02-26
Importing new files has some delays


ntran18:
Importing new files has some delays. For example, if you load an error file and want to reload another file, it still shows the previous error file, which needs to be re-uploaded.

ntran18:
I investigated this issue only in Chrome, not Safari. So the problem happens because of the web browser.

dondi:
Verified that this is a Chrome-only issue (Firefox does not manifest the problem either)

However due to Chrome’s popularity, it will be worth tracing the specific divergent behavior that Chrome is manifesting which cases the file selector to appear twice. Once we pinpoint that, we can determine if there is a way to address the divergent behavior without also regressing the other browsers

ntran18:
The root cause was that when the user clicks on "Select New File," it triggers ".upload".click(). But the class upload is bound to two Open File buttons: one in the sidebar and one at the top bar, which trigger the file selector twice.

ntran18:
[Database Missing Info.xlsx](https://github.com/user-attachments/files/17739080/Database.Missing.Info.xlsx)

dondi:
@ntran18 clarified how to replicate the issue and the fix was confirmed during the 11/13 meeting on beta 7.1.7

PR #1133, opened 2024-10-25
#1098 Fixing file selector appears twice after importing wrong file

ntran18:
The root cause was that when the user clicks on "Select New File," it triggers ".upload".click(). But the class upload is bound to two Open File buttons: one in the sidebar and one at the top bar, which triggers the file selector twice.

### DIFF
--- a/web-client/public/js/setup-load-and-import-handlers.js
+++ b/web-client/public/js/setup-load-and-import-handlers.js
@@ -171,8 +171,7 @@ export const setupLoadAndImportHandlers = (grnState) => {
      * for helping to resolve this.
      */
 
-    $("body").on("change", "#upload-network", uploadHandler(loadGrn));
-    $("body").on("change", "#upload-nav", uploadHandler(loadGrn));
+    $("body").on("change", ".upload", uploadHandler(loadGrn));
 
     const loadDemo = (url, value) => {
         $("#demoSourceDropdown option[value='" + value.substring(1) + "']").prop(
--- a/web-client/public/js/setup-load-and-import-handlers.js
+++ b/web-client/public/js/setup-load-and-import-handlers.js
@@ -171,8 +171,9 @@ export const setupLoadAndImportHandlers = (grnState) => {
      * for helping to resolve this.
      */
 
-    // $(".upload").change(uploadHandler(loadGrn));
-    $("body").on("change", ".upload", uploadHandler(loadGrn));
+    $("body").on("change", "#upload-network", uploadHandler(loadGrn));
+    $("body").on("change", "#upload-nav", uploadHandler(loadGrn));
+
     const loadDemo = (url, value) => {
         $("#demoSourceDropdown option[value='" + value.substring(1) + "']").prop(
             "selected",
--- a/web-client/views/upload.pug
+++ b/web-client/views/upload.pug
@@ -513,7 +513,7 @@ html
             p(id='error')
           div(class='modal-footer')
             input(type='button' class='btn btn-default' data-dismiss='modal' value='Close')
-            input(type='button' id='launchFileOpen' class='btn btn-primary' data-dismiss='modal' value='Select New File' onclick="$('.upload').click()")
+            input(type='button' id='launchFileOpen' class='btn btn-primary' data-dismiss='modal' value='Select New File' onclick="$('#upload-network').click()")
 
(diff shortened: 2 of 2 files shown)


web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -171,8 +171,9 @@ export const setupLoadAndImportHandlers = (grnState) => {
      * for helping to resolve this.
      */
 
-    // $(".upload").change(uploadHandler(loadGrn));
-    $("body").on("change", ".upload", uploadHandler(loadGrn));
+    $("body").on("change", "#upload-network", uploadHandler(loadGrn));
+    $("body").on("change", "#upload-nav", uploadHandler(loadGrn));
```
dondi, line 175: Per the discussion at our meeting, let’s try reverting this change so that we retain having a single line of code to designate this behavior, since it remains the same for both elements
```suggestion
   $("body").on("change", ".upload", uploadHandler(loadGrn));
```

web-client/views/upload.pug
```diff
@@ -513,7 +513,7 @@ html
             p(id='error')
           div(class='modal-footer')
             input(type='button' class='btn btn-default' data-dismiss='modal' value='Close')
-            input(type='button' id='launchFileOpen' class='btn btn-primary' data-dismiss='modal' value='Select New File' onclick="$('.upload').click()")
+            input(type='button' id='launchFileOpen' class='btn btn-primary' data-dismiss='modal' value='Select New File' onclick="$('#upload-network').click()")
```
dondi, line 516: Please see if this works:
```suggestion
            input(type='button' id='launchFileOpen' class='btn btn-primary upload' data-dismiss='modal' value='Select New File')
```

…if this doesn’t work, the general idea is to group this button alongside all of the other controls that trigger the upload handler. That way, we can specify with a single line of code that all of these elements are to invoke the upload handler when clicked/changed

dondi (commented):
Thanks @ntran18 for explaining the nature of the bug—with that in mind, these comments are more of an exercise to see how minimal the fix can be. Please experiment a bit and let me know

web-client/views/upload.pug
ntran18: I tried to give it an upload class, but it doesn't work. However, if we refer to just one of the upload buttons, it will trigger the uploadHandler.

dondi: OK, sounds good—yes that will work; the core problem was the `$('.upload')` selector so as long as we target just one element then we’ll be OK

web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -171,8 +171,8 @@ export const setupLoadAndImportHandlers = (grnState) => {
      * for helping to resolve this.
      */
 
-    // $(".upload").change(uploadHandler(loadGrn));
     $("body").on("change", ".upload", uploadHandler(loadGrn));
```
dondi, line 174: Looks like this just goes back to what it was, but quick question now that I see the commented-out line: are you aware of any difference between the top and bottom approaches? It seems like they would be the same, but clearly someone historically thought that doing `$("body").on("change", .....` would be better. I guess we can look up the commit log

dondi (commented):
This is nice! Just one investigative question on the `change` handling assignment—I can look that up too just in case, but if you already know, just go ahead and respond

web-client/public/js/setup-load-and-import-handlers.js
ntran18: The two approaches are similar, but the `.change` method is deprecated in jQuery from version 3.3, so Ona reconstructed the code to address this. The first approach attempts to bind the event directly to existing elements with the class upload, meaning these elements should already be present in the HTML file. In contrast, the second approach is better because it uses event delegation by binding the `change` event to the body element and specifying `.upload` as the target. This allows for handling events for upload elements that currently exist or will be added to the DOM in the future, meaning we can dynamically add the `upload` class via JavaScript and still manage the events effectively.

dondi: Ah got it—well deprecation is certainly a good reason for changing things, plus yes, the newer version indeed does “late binding” on the selector 👌🏽 Thanks for tracking this down!

dondi (approved):
Thanks for the explanation on the `.upload` handler!

Issue #1068, opened 2023-11-14
Node coloring does not support protein-level expression data


dondi:
Node coloring is currently completely gene expression-centric. For now, the immediate task is to:
* Include a disclaimer that we are displaying gene expression (mRNA-level data) and not protein-level data (separated out into #1071)

Eventually, we can explore to prospect of doing node coloring based on protein-level data

dondi:
@ntran18 notes that this might be a matter of matching up node names when showing a PPI network because `p` is appended to the end and capitalization is different

ntran18:
I'm able to make the node coloring appear for PPI if the user loads it from the database. However, the demos still have issue with displaying the node coloring.

dondi:
Confirmed in `beta`

kdahlquist:
confirmed in beta 7.1.7

PR #1134, opened 2024-10-25
#1068 Allow node coloring for PPI by matching the gene name (MERGE #1080 first)

ntran18:
The cause was that the gene name wasn't matching when querying the expression database. For protein, the gene name has an additional "p" at the end.

I can make the node coloring appear for PPI if the user loads it from the database. However, the demos still have issues with displaying the node coloring, but I think it's the same root cause documented in https://github.com/dondi/GRNsight/issues/1080.

### DIFF
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -11,7 +11,8 @@ import {
     ZOOM_DISPLAY_MAXIMUM_VALUE,
     ZOOM_DISPLAY_MIDDLE,
     ZOOM_ADAPTIVE_MAX_SCALE,
-    NETWORK_GRN_MODE
+    NETWORK_GRN_MODE,
+    NETWORK_PPI_MODE
 } from "./constants";
 
 /* globals d3 */
@@ -1087,11 +1088,15 @@ export var drawGraph = function (workbook) {
                 .selectAll(".coloring")
                 .data(function () {
                     if (grnState.workbook.expression[dataset]) {
+                        var geneName = p.name;
+                        if (grnState.workbook.meta.data.workbookType === NETWORK_PPI_MODE && geneName.endsWith("p")) {
+                            geneName = geneName.slice(0, -1);
+                        }
                         if (
-                            grnState.workbook.expression[dataset].data[p.name]
+                            grnState.workbook.expression[dataset].data[geneName]
                         ) {
                             const result = getExpressionData(
-                                p.name,
+                                geneName,
                                 dataset,
                                 average
                             );
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -1,6 +1,11 @@
 import Grid from "d3-v4-grid";
 import { grnState } from "./grnstate";
-import { modifyChargeParameter, modifyLinkDistanceParameter, valueValidator } from "./update-app";
+import {
+    modifyChargeParameter,
+    modifyLinkDistanceParameter,
+    valueValidator,
+    adjustGeneNameForExpression,
+} from "./update-app";
 import {
     ENDS_IN_EXPRESSION_REGEXP,
     VIEWPORT_FIT,
@@ -12,7 +17,6 @@ import {
     ZOOM_DISPLAY_MIDDLE,
     ZOOM_ADAPTIVE_MAX_SCALE,
     NETWORK_GRN_MODE,
-    NETWORK_PPI_MODE
 } from "./constants";
 
 /* globals d3 */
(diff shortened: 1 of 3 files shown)


web-client/public/js/graph.js
```diff
@@ -1087,11 +1088,15 @@ export var drawGraph = function (workbook) {
                 .selectAll(".coloring")
                 .data(function () {
                     if (grnState.workbook.expression[dataset]) {
+                        var geneName = p.name;
+                        if (grnState.workbook.meta.data.workbookType === NETWORK_PPI_MODE && geneName.endsWith("p")) {
+                            geneName = geneName.slice(0, -1);
+                        }
```
dondi, line 1094: This sequence appears to be the same logic as the version in _update-app.js_ line 396 onward; as such, let’s streamline these to a more consistent approach between versions:

```suggestion
                        const geneName = (
                          grnState.workbook.meta.data.workbookType === NETWORK_PPI_MODE && geneName.endsWith("p")
                        ) ? p.name.slice(0, -1)
                          : p.name;
```

web-client/public/js/update-app.js
```diff
@@ -393,7 +393,12 @@ const loadExpressionDatabase = function (isTopDataset) {
         queryExpressionDatabase({
             type:"ExpressionData",
             dataset,
-            genes : grnState.workbook.genes.map(gene => {return gene.name;}).join(","),
+            genes : grnState.workbook.genes.map(gene => {
+                if (grnState.workbook.meta.data.workbookType === NETWORK_PPI_MODE) {
+                    return gene.name.endsWith("p") ? gene.name.slice(0, -1) : gene.name;
+                }
+                return gene.name;
```
dondi, line 400: In the same spirit as what we’re doing in _graph.js_, we can make this a 100% conditional expression as well:
```suggestion
                return (
                  grnState.workbook.meta.data.workbookType === NETWORK_PPI_MODE &&
                  gene.name.endsWith("p")
                ) ? gene.name.slice(0, -1)
                  : gene.name;
```

dondi: Note also that upon unifying the coding style of the calculation in both _graph.js_ and _update-app.js_, we now see that the code is similar enough to extract as its own function. I won’t go into how that function exactly will be here, but it will probably involve some parameterization; however, doing that will still unify some of the logic

So, next step—recommendation is to do the first wave of changes, then test + commit; then when this approach looks good, make the next change of extracting a function from it

dondi (commented):
Overall approach looks good, but there is also an opportunity to exercise consistency with coding style, then leading to extraction as a function. Let’s give that a shot

web-client/public/js/graph.js
ntran18: Fixed

web-client/public/js/update-app.js
ntran18: Fixed

dondi (commented):
Looking good! I think we are almost done…just a couple more iterative changes and we should be set 🤗

web-client/public/js/update-app.js
```diff
@@ -383,6 +383,13 @@ const enableNodeColoringUI = function () {
     $(LOG_FOLD_CHANGE_MAX_VALUE_HEADER).removeClass("hidden");
 };
 
+const adjustGeneNameForExpression = function(gene) {
+    return grnState.workbook.meta.data.workbookType === NETWORK_PPI_MODE &&
+        gene.name.endsWith("p")
+        ? gene.name.slice(0, -1)
+        : gene.name;
+};
```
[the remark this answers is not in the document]
ntran18: Oh yes! This is better! Thanks for pointing it out!

web-client/public/js/update-app.js
```diff
@@ -393,7 +400,9 @@ const loadExpressionDatabase = function (isTopDataset) {
         queryExpressionDatabase({
             type:"ExpressionData",
             dataset,
-            genes : grnState.workbook.genes.map(gene => {return gene.name;}).join(","),
+            genes : grnState.workbook.genes.map(gene => {
+                return adjustGeneNameForExpression(gene);
+            }).join(","),
```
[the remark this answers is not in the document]
ntran18: Yes! I rare write my code like this, even though Dr. Toal taught me about this ... Thanks for helping me to write the code better!

dondi (approved):
Great revisions! The evolution of this PR is a nice example of iterative programming refinement. Functionality never changed, but I think the code got better with each round, and this is in optimal form now 💯

Issue #1053, opened 2023-10-10
Missing expression data for a gene in database causes UI problems


ntran18:
Problems: When you click in some expressions, especially the one contains ACE node, it would not reflect on the UI, meaning the check doesn't change to your selection. This is because the program can't load the expression. As I mentioned before, this is because there is no value `p.length` (I experienced some expressions, and it happens mostly for ACE node, not sure about the other nodes). However, if you click on that selection again, it works, but I don't know if it is correct or not. This problem only happens for dataset come from the database. 

1. Lint errors 
<img width="659" alt="image" src="https://github.com/dondi/GRNsight/assets/108908370/fab47071-ae48-4b5a-b77e-9ffacac9af3a">
<img width="923" alt="image" src="https://github.com/dondi/GRNsight/assets/108908370/00488539-2d8a-427b-bfb0-2699a33b9dde">
<img width="861" alt="image" src="https://github.com/dondi/GRNsight/assets/108908370/13b7f7a6-e0c1-471c-8047-16c2639a6197">
<img width="754" alt="image" src="https://github.com/dondi/GRNsight/assets/108908370/832d0a74-5b16-4f9b-9125-44b58dddaa5c">
--> From what I understand, I think either the version of database I currently have has different expression sheet name. A lot of errors were wrong name, or worksheet not detected. 

2. Example of error from console
<img width="368" alt="image" src="https://github.com/dondi/GRNsight/assets/108908370/3921b2a5-37ab-4f7e-9d89-5250461fa63d">

ntran18:
The error is caused by d3 graph generated an empty <g> element. 
![Screenshot 2024-11-04 at 9 13 58 PM](https://github.com/user-attachments/assets/c8d0196d-54d4-4543-971e-eb22fddda0d0)



From chatGPT, 
>In D3, if the dataset passed to .data() is either null, undefined, or an empty array, D3 will create an element for each item in the array. If there is no data (e.g., the result of result.data is an empty array or 0), D3 will append an empty <g> element even if it has no meaningful data to display.

To prevent d3 create empty g element, we return an empty list instead

kdahlquist:
We think this is fixed beta 7.1.7

PR #1138, opened 2024-11-05
#1053 fixing p.length error

ntran18:
The error is caused by d3 graph generated an empty <g> element. 
Screenshot 2024-11-04 at 9 13 58 PM

From chatGPT, 
>In D3, if the dataset passed to .data() is either null, undefined, or an empty array, D3 will create an element for each item in the array. If there is no data (e.g., the result of result.data is an empty array or 0), D3 will append an empty <g> element even if it has no meaningful data to display.

To prevent d3 create empty g element, we return an empty list instead

### DIFF
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -1101,9 +1101,7 @@ export var drawGraph = function (workbook) {
                                 average
                             );
                             timePoints = result.timePoints;
-                            return result.data && result.data.length > 0
-                                ? result.data
-                                : [];
+                            return result.data || [];
                         }
                     }
                     return [];
--- a/web-client/public/js/graph.js
+++ b/web-client/public/js/graph.js
@@ -1101,11 +1101,12 @@ export var drawGraph = function (workbook) {
                                 average
                             );
                             timePoints = result.timePoints;
-                            return result.data;
-                        } else {
-                            return 0;
+                            return result.data && result.data.length > 0
+                                ? result.data
+                                : [];
                         }
                     }
+                    return [];
                 })
                 .attr("class", "coloring")
                 .enter().append("rect")


web-client/public/js/graph.js
```diff
@@ -1101,11 +1101,12 @@ export var drawGraph = function (workbook) {
                                 average
                             );
                             timePoints = result.timePoints;
-                            return result.data;
-                        } else {
-                            return 0;
+                            return result.data && result.data.length > 0
+                                ? result.data
+                                : [];
```
dondi, line 1106: Looking to tighten up the code here, it seems that the case of `result.data.length === 0` means that `result.data` is already `[]` so making it branch to return a hardcoded `[]` seems redundant

From the look of it, you really just need to return the fixed `[]` if `result.data` is falsy, right? Would this work?
```suggestion
                            return result.data ?? [];
```

dondi (commented):
Thanks for this! Indeed D3 typically wants arrays back. It’s really small, but again in the spirit of honing code conciseness, please see if the suggestion in the review will suffice for the fix

web-client/public/js/graph.js
ntran18: I tried this approach before, and I got a parser error.

dondi: Ah OK—that probably means our GRNsight parser hasn’t been updated to support `??`

Earlier versions of JavaScript used `||`—it should generally still work here:
```javascript
                            return result.data || [];
```

ntran18: Yes, it works! `&&` is used for when `result.data` is null or undefined. But for `||`, it will return `[]` if `result.data` is null, undefined, 0, or "".

dondi (approved):
LGTM

ceciliazaragoza:
This works as expected for GRNs and PPIs in beta.

PR #1140, opened 2024-11-05
Addition fix to previous fix for #1093

ntran18:
When exporting the svg, using cloneNode making the image to be all black. This is because style didn't pass correctly. This makes sure the style is correct.

### DIFF
--- a/web-client/public/js/setup-handlers.js
+++ b/web-client/public/js/setup-handlers.js
@@ -94,13 +94,11 @@ export const setupHandlers = grnState => {
         window.document.body.appendChild(emptySvg);
         var emptySvgDeclarationComputed = getComputedStyle(emptySvg);
 
-        const traverse = svg => {
-            var tree = [];
-            tree.push(svg);
-            // implement DFS
-            const visit = (node) => {
-                if (node && node.hasChildNodes()) {
-                    var child = node.firstChild;
+        const traverse = (node) => {
+            const tree = [];
+            const visit = (currentNode) => {
+                if (currentNode && currentNode.hasChildNodes()) {
+                    let child = currentNode.firstChild;
                     while (child) {
                         if (child.nodeType === 1 && child.nodeName !== "SCRIPT") {
                             tree.push(child);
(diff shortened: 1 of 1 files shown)


web-client/public/js/setup-handlers.js
```diff
@@ -110,38 +108,38 @@ export const setupHandlers = grnState => {
                     }
                 }
             };
-            visit(svg);
+            visit(node);
             return tree;
         };
 
         const explicitlySetStyle = element => {
             const cSSStyleDeclarationComputed = window.getComputedStyle(element);
```
dondi, line 116: I realize this was pre-existing code, but since you’re already changing this, let’s fix the capitalization issue: the `CSS` here should actually be all lowercase, if we are to follow the convention:
```suggestion
            const cssStyleDeclarationComputed = window.getComputedStyle(element);
```

web-client/public/js/setup-handlers.js
```diff
@@ -110,38 +108,38 @@ export const setupHandlers = grnState => {
...
-                key = cSSStyleDeclarationComputed[i];
-                value = cSSStyleDeclarationComputed.getPropertyValue(key);
+            for (let i = 0; i < cSSStyleDeclarationComputed.length; i++) {
+                const key = cSSStyleDeclarationComputed[i];
+                const value = cSSStyleDeclarationComputed.getPropertyValue(key);
                 if (value !== emptySvgDeclarationComputed.getPropertyValue(key)) {
                     // Don't set computed style of width and height. Makes SVG elmements disappear.
                     if ((key !== "height") && (key !== "width")) {
-                        computedStyleStr += key + ":" + value + ";";
+                        computedStyleStr += `${key}:${value};`;
```
dondi, line 125: Great change! Yes, in the age of template strings I generally change any concatenations `+` into template/formatted strings 💯

web-client/public/js/setup-handlers.js
```diff
@@ -110,38 +108,38 @@ export const setupHandlers = grnState => {
...
             }
-            element.setAttribute("style", computedStyleStr);
+
+            if (element.classList.contains("weight")) {
+                computedStyleStr += "visibility: hidden;";
+            }
+
+            if (computedStyleStr) {
+                element.setAttribute("style", computedStyleStr);
+            }
```
dondi, line 137: The observation here is that we are building a string in order to set what is really an object—style attribute names + their values. In the end, I think this whole block will be more robust if we build the desired style as an _object_ instead of a string. Then, `Object.assign` can be used to set the final style once the object is built. Here’s a link to show how this would work: https://www.bram.us/2017/01/16/using-object-assign-to-quickly-set-multiple-inline-styles/

This does change this block of code quite a bit, but I think it is a good change in the right direction—what do you think?

web-client/public/js/setup-handlers.js
```diff
@@ -110,38 +108,38 @@ export const setupHandlers = grnState => {
...
         };
 
         // hardcode computed css styles inside svg
         var allElements = traverse(svg);
-        var i = allElements.length;
-        while (i--) {
-            explicitlySetStyle(allElements[i]);
-        }
+        allElements.forEach(explicitlySetStyle);
```
dondi, line 142: Another great refactor, using the function-as-iterator trick from a prior PR 🤩 Also nice to see 4 lines become just one, right? 😁

dondi (commented):
Good find! For this one, I realize you are revising pre-existing code, but it looks like the pre-existing code itself could have used some refactors. Please see if we can switch from a concatenated string to `Object.assign` when building up the new inline style for the node(s)—thanks!

web-client/public/js/setup-handlers.js
ntran18: I think `Object.assign` is better!

web-client/public/js/setup-handlers.js
```diff
@@ -110,38 +108,39 @@ export const setupHandlers = grnState => {
...
-            for (i = 0, len = cSSStyleDeclarationComputed.length; i < len; i++) {
-                key = cSSStyleDeclarationComputed[i];
-                value = cSSStyleDeclarationComputed.getPropertyValue(key);
+            const cssStyleDeclarationComputed = window.getComputedStyle(element);
+            const computedStyleObj = {};    
+
+
+            for (let i = 0; i < cssStyleDeclarationComputed.length; i++) {
+                const key = cssStyleDeclarationComputed[i];
+                const value = cSSStyleDeclarationComputed.getPropertyValue(key);
```
dondi, line 122: Whoops, shouldn’t this be edited also?
```suggestion
                const value = cssStyleDeclarationComputed.getPropertyValue(key);
```

dondi (commented):
That `Object.assign` change works really well I think! But I think you missed one more capitalization change?

For things like that, VS Code’s _Rename_ command really helps—I can show it to you if you aren’t familiar with it yet

web-client/public/js/setup-handlers.js
ntran18: I didn't catch this one. Usually, I would use Ctrl + D to select all words, but I think I was fixing it manually. But I would love to learn about VS Code's Rename command!

dondi (approved):
Thanks for the quick fix! LGTM

ceciliazaragoza:
I checked this in beta, and it works as expected!

PR #1141, opened 2024-11-06
Improve current export images (png and pdf)

ntran18:


dondi (approved):
Nice—you are gaining some expertise in both SVG and PDF here!

dondi:
I’ll wait to merge this for now, in case there are dependencies on the earlier PRs

Issue #1144, opened 2024-11-13
If a Excel file is loaded with no expression data, default node coloring to "off"


kdahlquist:
If a Excel file is loaded with no expression data, default node coloring should be "off" (noticed when looking at fix for #1074)

dondi:
Verified by @kdahlquist on v7.1.11 during 12/4 meeting

PR #1153, opened 2024-11-27
#1144 Disable node coloring when uploading file with no expression

ntran18:


web-client/public/js/update-app.js
```diff
@@ -512,6 +512,15 @@ const toggleLayout = (on, off) => {
...
 };
 
+const hasExpressionData = (sheets) => {
+    for (var property in sheets) {
+        if (property.match(ENDS_IN_EXPRESSION_REGEXP)) {
+            return true;
+        }
+    }
+    return false;
+};
```
dondi, line 522: This isn’t functionally incorrect, but along the lines of more current idioms when processing lists or objects, see if this can be written as a `some` call? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

Using `some` will make the code look something like this…

```suggestion
const hasExpressionData = (sheets) => {
    return (/*expression with sheets*/).some(
        value => (/*expression with value */).match(ENDS_IN_EXPRESSION_REGEXP)
    )
};
```

…which to me sounds more concise, and objectively does not use side effects as well as easier for an optimizer to parallelize

I realize `sheets` isn’t an array, but there may be ways to take advantage of the simpler expressiveness of `some` vs. having a `return true` with a `return false` at the end (that’s why I have some placeholder `/* */` comments in there instead of specific code; there may be other places too)

dondi (commented):
Good refactors! I have a suggestion for one more…see if it’s doable (hope you agree about the benefits I list here about the benefits of using the `some` iterator)

web-client/public/js/update-app.js
ntran18: I just modified it to use `some` call instead.

web-client/public/js/graph.js
```diff
@@ -1235,12 +1235,7 @@ export var drawGraph = function (workbook) {
     };
 
     const hasExpressionData = sheets => {
-        for (var property in sheets) {
-            if (property.match(ENDS_IN_EXPRESSION_REGEXP)) {
-                return true;
-            }
-        }
-        return false;
+        return Object.keys(sheets).some(property => property.match(ENDS_IN_EXPRESSION_REGEXP));
```
dondi, line 1238: Thanks, this looks good! But it looks like this function is duplicated in _update-app.js_? (see line 515 below) Should these functions be consolidated into one?

Sorry I didn’t notice this before…

web-client/public/js/update-app.js
dondi: You certainly modified it in _graph.js_, but does that mean that lines 515–522 above are duplicated?

dondi (commented):
The `some` refactor looks great, but I hadn’t realized that there appear to be two copies of `hasExpressionData`? Please take a look at the overall diff; thanks!

web-client/public/js/graph.js
ntran18: I didn't notice it either ....

dondi (approved):
LGTM—it appears we have accidentally found some hidden duplicated code 😁 That was a lucky search-and-replace accident haha

Issue #1148, opened 2024-11-13
Include test coverage results in GitHub Action CI script


dondi:
In addition to running tests, we should start recapturing coverage results from the tests (formerly handled by Coveralls) and include this in the script as well. Per #1146, once we have coverage captured, it would be good to also include these results in the README badge

akaiap:
1. I will modify the Github Actions workflow to generate coverage reports
- run: `npm test -- --coverage --coverageReporters=text-lcov | coveralls
      env:
        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}` _Need to confirm coveralls account in repo secrets_
2. Update README file 
- `[![Coverage Status](https://coveralls.io/repos/github/<OWNER>/<REPO>/badge.svg?branch=master)](https://coveralls.io/github/<OWNER>/<REPO>?branch=master)`
-  Test Test Test

dondi:
Instructions look good; learning about secrets systems comes with the territory so is a good aspect of these tickets

akaiap:
What I have done so far 
- Updated the scripts in mocha 
- Did some tested and updated mocha's version. 
```

-coverage: Uses nyc with mocha to run tests and collect coverage data.
coveralls: Generates a coverage report in LCOV format and pipes it to coveralls.
```
- After removing Istanbul and installed I did some testing using `npm run coverage and npm run coveralls`
- and success ! <img width="1470" alt="Screenshot 2024-12-04 at 12 30 13 PM" src="https://github.com/user-attachments/assets/1b6b6794-4f72-42f9-84ff-fa427f88945c">
- However, adding the coverage badge can get complicated due to GitHub "secrets." These secrets are admin tokens for Coveralls, and the safest approach is to avoid committing the token.


**_next steps to get the token:_** 
Sign in to [Coveralls.io](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) with GRNSight account.
Enable repository if not already active.
Navigate to the repository's settings on Coveralls.
Copy the Repo Token provided.
**Add the Token to GitHub Repository Secrets:**

Click on Settings > Secrets and variables > Actions.
Click on New repository secret.
Name the secret COVERALLS_REPO_TOKEN.
Paste the token you copied from Coveralls.
Save the secret.


This way this token will not be pushed to the repo.

dondi:
[...]

PR #1170, opened 2025-02-05
#1148 include test coverage in actions

akaiap:
Capturing coveralls results to appear in the README.md as the badge.


.github/workflows/node.js.yml
```diff
@@ -1,34 +1,58 @@
-# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
-# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs
-
```
dondi, line 3: I’m curious about the removal of this comment—wouldn’t it be better to document what’s going on in this way? Or is it that the comment contains incorrect information?

test/grnstate-tests.js
```diff
@@ -1,4 +1,5 @@
-import "chai/register-should";
+const chai = require("chai");
+chai.should(); // Activates "should" style assertions
```
dondi, line 2: Is there a reason that this is calling `require` rather than `import`? The latter is the newer approach. Also, we should move `chai.should()` to after all of the `import` statements because the convention is to do all imports first before executing code

dondi (commented):
Generally looks good, but with some questions and a small potential revision before we proceed. Let’s address these then do another round

.github/workflows/node.js.yml
akaiap: Ah yes! It is correct-- I will add that back

test/grnstate-tests.js
akaiap: - Since Mocha and the Node environment were still configured for CommonJS modules, using `require` made the test file run correctly under the regular CommonJS system. Switching to `import` would require the project to be recognized as an ES module (e.g., by setting "type": "module" in package.json).

- I will fix the import!

akaiap: I tested the build again after moving the chai.should() call below all the import statements. While the build still ran, it triggered warnings and linting errors.

- The issue is from an inconsistent module system, where ES Modules (import) are being processed before CommonJS (require). Moving chai.should() below the imports affects the execution order in a way that conflicts with the current module setup.
- Linting failures are due to ESLint 9 deprecating old configuration formats:
ESLint now expects the "ignores" rule inside eslint.config.js instead of relying on the .eslintignore file.
The "languageOptions" object no longer allows "ecmaFeatures", requiring an updated configuration.

We can go down to V8, but may still encounter the same error.

dondi (commented):
Oh oof—I remember typing my comment last week during our meeting, but I failed to submit it…submitting now, sorry for the miss 😬

dondi:
Pinging on this—I think there are still some unresolved comments?

dondi (commented):
Deployment issues aside (as reported in #1148), the PR itself has been updated by running eslint’s automated `--fix` function, reducing existing linter errors down to 11, all of which say the same thing: `Parsing error: 'import' and 'export' may appear only with 'sourceType: module'`

Please look into this so that ideally we can “go green” on this PR then I think we will be ready to merge

akaiap:
Currently, so happy right now!! 
After months of getting the test coverage working, the badge is being updated, and coverage is passing!

Steps:
After cleaning up the code to now take imports/export of grnState, the code was pushed and passed all the checks! 

### Next steps: 
- While the checks pass -- I currently have ` continue-on-error: true # Linter failures won't stop the pipeline` in the yml package when Grnisight is built.  
- When I lint locally, I am receiving `1403 problems (1403 errors, 0 warnings)`
- Realized I was using two active and conflicting config files. _(.eslint.config.js and eslintrc.yml)_
- The next steps are to delete the old format (yml file) and reformat the config file to be able to use the `--fix command`

akaiap:
**ITS WORKING!** 

### Steps taken 
- Fixed module system config: Properly configured ESLint to support ES Modules w

- Resolved quote style issues: Fixed inconsistent quotes (single vs. double quotes)

- Proper configuration: Set up appropriate globals and rules in the eslint.config.js file 

Clean code base: the code now passes all linting checks, making it more maintainable and consistent :))))

dondi (approved):
LGTM

Issue #1150, opened 2024-11-20
Can't load same demo after loading from database


kdahlquist:
Do the following actions on beta:
- load Demo #2 from side panel
- load a grn from the database
- try loading Demo #2 again from the side panel--it doesn't load

I think this is because Demo #2 is still showing up in the side panel and clicking on it again doesn't do anything.  I can load it from the Network menu OK.

dondi:
This appears to break down into two distinct buggy behaviors:
* It should be possible to re-load the same demo; this resets the demo network to the original force-graph setup. The Network menu and Demo menu versions of this dropdown behave in this manner; we should find a way to make the sidebar version do the same
* If a demo is selected but the user then opens something different (e.g., a file or a database network), then the sidebar `select` element should reflect this by reverting to “Select a demo”

If it becomes necessary to separate these into their own issues, we can do so

MilkaZek:
On update-app.js we added a new function to reset the demo dropdown and called it whenever the 'Generate Network' button gets clicked; the bug seems to be fixed. I've attached screenshots of the code snippets we added to update-app.js. We're unsure if we should place this code in the generate-network.js file or keep it here. We also weren't sure if keeping lines 602 & 603 from the screenshot is necessary, for the bug seems to be fixed without them, but the other Network Mode value change listeners include them.

<img width="529" alt="Image" src="https://github.com/user-attachments/assets/b7b5c845-3410-45e4-abf3-fd0bf0562086" />

<img width="556" alt="Image" src="https://github.com/user-attachments/assets/a7492090-aa9e-43db-8500-0b8e22e11274" />

dondi:
Go ahead and create a branch then put these changes there, and open pull request. That will let us do a code review and also try the code out live

dondi:
The recommendation for how to proceed here is to start tracing where the incorrect behavior is taking place; I suggest putting a console log there while replicating the unwanted behavior to see when it gets invoked and by whom. Possible issues include:
* Unintended setting of the `demoDropdownValue` somewhere during the update sequence
* Unintended/overlooked setting of a web app element (via the jQuery `$` function) somewhere in the sequence

kdahlquist:
I just poked around at this.  The problem only happens when selecting a demo from the side panel.  The behavior is correct when using the "network" or "demo" menu to load the demo.

kdahlquist:
This issue also occurs if you upload a user file in between selecting the demos from the side panel.

dondi:
[...]

PR #1171, opened 2025-02-12
Resetting Demo Dropdown #1150

MilkaZek:
We added a function in update-app.js to reset the Demo Dropdown, and call it when the Generate Network button is clicked, which seems to have fixed the bug of the original demo not being able to be reclicked after generating a network. There is a possibility that this code can be condensed/placed in a different place, but we were unsure about trying that out.

### DIFF
--- a/web-client/public/js/update-app.js
+++ b/web-client/public/js/update-app.js
@@ -549,6 +549,11 @@ const updateModeViews = () =>{
     }
 };
 
+const resetDemoDropdown = () =>{
+    $("#demoSourceDropdown option").removeAttr("selected");
+    $("#demoSourceDropdown").val("none");
+};
+
 const checkWorkbookModeSettings = () => {
     const hasExpression = hasExpressionData(grnState.workbook.expression);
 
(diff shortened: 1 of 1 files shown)


dondi (commented):
So after going a bit deeper into the existing code, I think I’ve put my finger on the core underlying issue: `grnState` doesn’t have a notion of “selected demo.” The idea of the selected demo is currently housed exclusively in the value of `#demoSourceDropdown`. To make this follow our MVC flow cleanly, I think we should start with that change, then take it from there

As such, I’d say the steps are as follows:
1. Add a new property to `grnState` (_grnstate.js_) which stores the currently-selected demo. This can be `none` when the current network is from a different source
2. Somewhere in `updateApp`, we need a block of code that updates `#demoSourceDropdown` based on this new property in `grnState`. We can also update the other items as pointed out by @ntran18 (the menu bar and the demo menu) at this point, because they are all based on this new `grnState` property
3. Existing code that touches `#demoSourceDropdown` (including what’s in this PR) should then be restructured such that:
    * They modify the `grnState` property (look at existing code for how this is done by other UI elements)
    * This change to `grnState` should then invoke `updateApp`…I’m not sure at this point if this update is done implicitly or if it has to be explicitly called; check existing code to see how the cycle goes

I realize that this is now a lot more work than what you have in this PR 😅 but the idea is to implement the fix within the proper structure as well. It is fair to say that this bug actually exists precisely because the value of the selected demo was not stored in `grnState` to begin with; had that property been tracked in that centralized way, the original question of where to put this code would have been easier to answer

@ntran18 @ceciliazaragoza how does the plan above sound?

ceciliazaragoza:
That sounds good to me!

dondi:
Work on this has been shifted to `amelie-1150`

PR #1175, opened 2025-02-19
Updating database [Cecilia's PR]

ntran18:


server/dals/network-dal.js
```diff
@@ -18,13 +18,20 @@ var sequelize = new Sequelize(
...
+const buildNetworkGeneFromSourceQuery = function (gene, timestamp) {
+    const namespace =
+      timestamp < new Date("2025-01-01")
+        ? "gene_regulatory_network.gene"
+        : "gene_regulatory_network_new.gene";
     return `SELECT DISTINCT gene_id, display_gene_id FROM 
-    gene_regulatory_network.gene WHERE (gene.gene_id ='${gene}'
-    OR gene.display_gene_id ='${gene}')`;
+    ${namespace} WHERE (gene.gene_id ='${gene}'
+    OR gene.display_gene_id ='${gene}') AND gene.time_stamp ='${timestamp}'`;
```
ntran18, line 34: Old namespace doesn't have timestamp so this would cause error

server/dals/protein-dal.js
```diff
@@ -18,15 +18,22 @@ var sequelize = new Sequelize(
...
+    const namespace =
+      timestamp < new Date("2025-01-01")
+        ? "protein_protein_interactions"
+        : "protein_protein_interactions_new";
     return `SELECT DISTINCT gene_id, display_gene_id, standard_name, length, molecular_weight, PI FROM
- protein_protein_interactions.gene, protein_protein_interactions.protein WHERE
- (LOWER(gene.gene_id)=LOWER('${geneProtein}') OR LOWER(gene.display_gene_id)=LOWER('${geneProtein}') 
- OR LOWER(protein.standard_name) =LOWER('${geneProtein}')) AND
- LOWER(gene.gene_id) = LOWER(protein.gene_systematic_name);`;
+    ${namespace}.gene, ${namespace}.protein WHERE
```
ntran18, line 33: Similar here, old namepsace doesn't have timestamp. Additionally, if we query from protein table, then we should check the timestamp from the protein table. If we query from gene table, then we should check the timestamp from the gene table.

ntran18:
I verified the code is working

dondi (approved):
LGTM

PR #1178, opened 2025-02-25
#1150 updated solution to align with the existing structure

Amelie1253:


MilkaZek:
We created a new property in grnstate.js (demoDropdownValue) to hold the value of the Demo Dropdown and in update-app.js we check to see if demoDropdownValue is equal to null. If it is, we reset the dropdown selection. When a custom workbook is being displayed, demoDropdownValue is set to null. If a demo dropdown value is selected, the value of the new grnstate property is changed so that it matches what was selected and isn't null anymore. 

We think we need to find out how the three demo dropdowns are connected in their values to move forward because with the addition of our new code, when the top dropdown is changed, the one on the sidebar changes to match but then is set to null.

dondi (approved):
Great, thanks for the linter fixes!

Issue #863, opened 2020-04-30
Need indicator to user to differentiate between user-uploaded and Expression Database data


kdahlquist:
I see how the functionality works now.  The user uploaded data and Expression Database data are both accessible from the drop-down menus, which is what we wanted.

However, there is no indicator to the user that this is the case.  Also, our Expression Database is only for yeast right now.

We need to provide some more info to the user about this.  I'm pretty braindead right now so I don't have any suggestions off the top of my head, but I wanted to note this as an issue.

dondi:
Possibly prepending text or looking into `optgroup` may make this low-hanging fruit

ntran18:
I think we need to *sit down* and think where to put it before I can implement anything 😅.

kdahlquist:
There would be two optgroups:
- User-uploaded
- Expression Database

MilkaZek:
We went through the various pages and identified this section (referring to the attached screenshot/code snippet) of the update-app.js file to be where we need to separate the drop-down menu options - we (@Amelie1253 and I) plan to attempt that fix this upcoming week after clarifying which list options fall under the category of User-uploaded or Expression Database.

<img width="741" alt="Image" src="https://github.com/user-attachments/assets/5c106bd3-83fb-4d4c-a453-014f7f437714" />

dondi:
In response to this comment…

> We went through the various pages and identified this section (referring to the attached screenshot/code snippet) of the update-app.js file to be where we need to separate the drop-down menu options - we ([@Amelie1253](https://github.com/Amelie1253) and I) plan to attempt that fix this upcoming week after clarifying which list options fall under the category of User-uploaded or Expression Database.

The code snapshot points to the right place; the values listed in line 714 should be what eventually gets wrapped in an `opt-group` labeled “User-uploaded” and the line 721 block is what gets wrapped in an “Expression database” `opt-group`

dondi:
For question regarding the long comment, that pertains to #865 —we have backlogged it because the workaround appears to remain stable and we haven’t had the bandwidth to investigate this further for a “real” solution

We can help future developers by cross-linking between this comment and the issue (i.e., mention the issue number in the comment, and perhaps in the issue, mention the file/approximate location of the workaround)

kdahlquist:
Upon testing, we realized that we needed clarification of the behavior:
[...]

PR #1179, opened 2025-02-26
#863 Separated options into user-uploaded and expression database

Amelie1253:
This issue #863 is fully functional and has been resolved but this PR contains code for the solution for #1150 which hasn't been fully resolved yet. However, the solution for this issue is independent of the code from #1150.


dondi:
The work toward #863 has been cherry-picked out and placed in `amelie-milka-863`

PR #1181, opened 2025-02-26
#863 separated options into User-Uploaded and Expression Database

Amelie1253:


web-client/public/js/update-app.js
Amelie1253, line 1: I realized that I accidentally initialized workbookExpressions and databaseExpressions twice, once in grnstate.js and once in update-app.js. The lines 703 and 704 in update-app.js can be deleted.

dondi (commented):
Similarly to the other PR, this looks good but let’s also resolve the linter issues first

This allows you to also make the changes you mentioned in the other comment about the double initializations 😊

dondi (approved):
LGTM with the linter fixes; thanks!

PR #1184, opened 2025-03-12
Improving upon #1183 hotfix

ceciliazaragoza:
Was using hardcoded namespace instead of constants, so exported namespaces from dbConstants. Changed variable names for constants to include `with_timestamp` instead of `_old`.


dondi (approved):
Agreed, those constant names line up better with their values 👌🏽

Issue #1164, opened 2025-01-22
Attach timestamps to genes


dondi:
Emerging from the discussion of #1120, we have resolved that the gene table should itself have a timestamp. Even if a high percentage of genes will be the same from year to year, producing nearly the same records, we also don’t want to include genes that didn’t exist at a particular time with the interactions from that time

dondi:
@ceciliazaragoza reported that initial queries that include the timestamp now run without errors, but for `ZAP1`, which regulates itself, the updated query does not return the expected self-regulation adjacency item

@kdahlquist checked the 2025 data and confirmed that `ZAP1` does still regulate itself, so @ceciliazaragoza will investigate the query further. @ntran18 can investigate the latest loaded data to see if the issue resides in the database scripts instead

ntran18:
The issue with `ZAP1` is not from the database queries but from the scripts that getting data from AllianceMine. I checked the script, and we don't get data that have `ZAP1` self-regulated. I have to investigate the queries more.

dondi:
See comment on #1120 for next steps

Issue #1120, opened 2024-10-02
Port database loading scripts to AllianceMine


dondi:
Now that InterMine is no more, we need to port our loading scripts to AllianceMine

ntran18:
Add documents to the [Wiki](https://github.com/dondi/GRNsight/wiki/Database-Migration-From-YeastMine-to-AllianceMine). I wanted to take a break from it last week, so I didn't investigate PPI. I will do it this week.

ntran18:
I got response from SGD support about the API or service they have. I will test look at it and see if it works. 

<img width="913" alt="image" src="https://github.com/user-attachments/assets/6604c20b-70f3-47e2-8fb5-6d3bec10f1f6">

ntran18:
I updated the [Wiki](https://github.com/dondi/GRNsight/wiki/Database-Migration-From-YeastMine-to-AllianceMine) with all analyses for GRN and PPI and the proposed solutions for each.

ntran18:
@kdahlquist https://www.alliancegenome.org/bluegenes/alliancemine/templates. Currently AllianceMine has some templates for Yeastmine. Could you please check over them?

kdahlquist:
I looked at the Organism > All Genes query results from the link given in the above comment.  This looks like what we need to retrieve all the genes in yeast.  There are a couple of observations:

- The default query as written does not retrieve the systematic names, only the standard names.  In order to get the systematic names, you need to click the "Edit Query" button and select systematic name in the model browser panel on the far left.  I don't know whether the revised query can then be converted into a web service URL like the default query can.
- The All Genes query retrieves 7261 records.  We only need the Gene > Featuretype "ORF" records (6700 records).  You could probably play with the query to only retrieve those instead of filtering them afterwards.

The Literature > Regulation template is what we need to use to get the network information.  
- We should capture this entire table.
- Regulator > Standard Name or Systematic name is the regulator
- Gene > Standard Name or Systematic name is the target
- I think for this one, you can use the default query

One more thing to note.  I looked at the downloaded results files and it looks like when a gene doesn't have a standard name, they leave that field blank.  In that case, we need to copy the systematic name and put it in the standard name field.

Let me know if you have more questions.

kdahlquist:
Literature > Regulation is for the GRNs.

For PPI, use Literature > Interaction and just use the "physical" interaction type.

ntran18:
I will take a look at these

ntran18:
[...]

PR #1185, opened 2025-03-12
#1120 #1164 Fix for updating data

ntran18:
I found a bug in my code that requires me to change the current schema. I removed code related to filtering because we don't need to filter for updates anymore. However, the current schema doesn't allow adding genes/proteins at different timestamps because it doesn't consider timestamps and sources for primary keys. 

This PR:
- Remove code related to filtering data
- Change schema of gene and protein tables to make timestamp and source as primary key
- Remove unnecessary columns (gene_systematic_name1, gene_systematic_name2) in protein_protein_interactions_with_timestamp.physical_interactions.


dondi (approved):
LGTM

PR #1196, opened 2025-04-25
Fixing genes' orders in network sheet for ppi and grn

ntran18:


web-client/public/js/generateNetwork.js
```diff
@@ -205,6 +205,12 @@ containing "-", "_", and alpha-numeric characters only`);
         $(CREATE_NETWORK_MODAL).modal("show");
     };
 
+    const sortGenes = function (genes, idToSort) {
```
dondi, line 208: Solely for code readability, let’s define constants which assign a name to what the `0` and `1` IDs mean; something like:
```suggestion
    const GRN_STANDARD_NAME_ID = 1;
    const PPI_STANDARD_NAME_ID = 0;

    const sortGenes = function (genes, idToSort) {
```

web-client/public/js/generateNetwork.js
```diff
@@ -242,6 +248,7 @@ containing "-", "_", and alpha-numeric characters only`);
  ${genesAmount} genes. Please remove some genes from your proposed network.`);
         } else {
             if (grnState.customWorkbook.type === NETWORK_GRN_MODE) {
+                grnState.customWorkbook.genes = sortGenes(grnState.customWorkbook.genes, 1);
```
dondi, line 251: That way, over here, we have named constants rather than hardcoded numbers:
```suggestion
                grnState.customWorkbook.genes = sortGenes(grnState.customWorkbook.genes, GRN_STANDARD_NAME_ID);
```

dondi (commented):
Looks good, just needs a quick readability tweak

dondi (approved):
LGTM—thanks for making the revisions!

web-client/public/js/generateNetwork.js
```diff
@@ -242,6 +251,7 @@ containing "-", "_", and alpha-numeric characters only`);
  ${genesAmount} genes. Please remove some genes from your proposed network.`);
         } else {
             if (grnState.customWorkbook.type === NETWORK_GRN_MODE) {
+                grnState.customWorkbook.genes = sortGenes(grnState.customWorkbook.genes, GRN_STANDARD_NAME_ID);
                 const genes = Object.keys(grnState.customWorkbook.genes);
                 const displayGenes = Object.keys(grnState.customWorkbook.genes).map(g => grnState.customWorkbook.genes[g]);
```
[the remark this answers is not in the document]
ntran18: I agree that Object.keys doesn't preserve order. I think it's best if we can switch Object.keys to an array. I will create a new PR for this proposed change.

PR #1199, opened 2025-04-30
Amelie milka 863

Amelie1253:
I tested our solution and it seems to work. I just have a question about what opt-groups should be visible when the user loads from the database? Currently my solution only shows the Expression Database data as the opt-group when the user loads from the database and enables node coloring.


dondi (commented):
pretty good almost there:
- Let’s delete the _.pyc_ files. Side issue will be to review _.gitignore_ to see why these were allowed through
- Merge the latest `beta` into this branch so that older changes don’t get mixed in

Amelie1253:
Thank you @ntran18 for helping with deleting files and merging beta! I reversed what I had added so that the code aligns with what was discussed during the meeting. There will always be two opt-groups either Demo or User-Uploaded and Expression Database, even if there is no data in User-Uploaded. That seems to be it, everything is in order otherwise!

web-client/public/js/setup-load-and-import-handlers.js
```diff
@@ -17,6 +17,7 @@ import {
     NETWORK_GRN_MODE
 } from "./constants";
 import { getWorkbookFromForm, getWorkbookFromUrl } from "./api/grnsight-api";
+import { updateLocale } from "moment";
```
dondi, line 20: I think I missed this from before—this looks like an unused import? Either way I won’t let it block the PR; it has no effect as of now and can be fixed easily later

dondi (approved):
Thanks for cleaning out the files! I spotted something I missed earlier, but nbd

web-client/public/js/setup-load-and-import-handlers.js
Amelie1253: Not sure where that came from but I took it out!

PR #1225, opened 2025-09-14
Switching from Object.keys to Object.entries to preserve order of genes

ntran18:


dondi (approved):
LGTM, this looks like it will preserve the “array”-ness of the data structure throughout, which should then preserve the order

Issue #1231, opened 2025-09-19
Adding scroll bars for container displays warning messages


ntran18:
I couldn't see the whole warning messages when there are multiple messages and it overflows the height of the container for those warning messages and there is no scroll bars. 

<img width="1907" height="512" alt="Image" src="https://github.com/user-attachments/assets/241328e2-2e46-4581-9a2f-4d106cf69b48" />

kdahlquist:
@ntran18, do you have a test file that I can use to review this fix?   Also, it would be good to add this test file to our testing regime.

ntran18:
This happened when I was testing locally with this test file.

The process was:
- Right click and click Inspect to debug
- Upload the file
- Import the file to Excel
- Click on warnings window
- Problem: I can't scroll to see the full list of warnings

However, when I tried to see it on beta production, I don't see this error

[18-genes_0-edges_no-deg-prod-rates.xlsx](https://github.com/user-attachments/files/22524191/18-genes_0-edges_no-deg-prod-rates.xlsx)

dondi:
After reviewing the `beta` and local behaviors during the meeting, @dondi concluded that this was an artifact of how the two setups determine the size of the GRNsight app. To line up the conditions more closely, it was suggested that @ntran18 look at @ceciliazaragoza’s local-iframe work from last year and run the app locally within an iframe. This may help foster a clearer understanding of when the dialog scrollbar appears vs. not

PR #1232, opened 2025-09-19
#1231 adding overflow: auto to warning models and error models

ntran18:


web-client/public/js/warnings.js
```diff
@@ -52,18 +52,7 @@ export var displayWarnings = function (warnings) {
...
-    var setPanel = screenHeight - BORDER + "px";
-    var minPanel = MIN_SCREEN_HEIGHT - BORDER + "px";
-    if (screenHeight > MIN_SCREEN_HEIGHT) {
-        $("#list-frame").css({ height: setPanel });
-    } else {
-        $("#list-frame").css({ height: minPanel });
-    }
-
-    $("#warningsModal").modal("show");
+    showWarningsModel();
```
dondi, line 55: Good duplication catch and code consolidation!

web-client/public/js/warnings.js
```diff
@@ -79,15 +68,19 @@ export var displayPPINodeColorWarning = function (warningDisplayed) {
         ].join(" ")
     );
 
+    showWarningsModel();
+};
+
+var showWarningsModel = function () {
```
dondi, line 74: Minor thing, won’t block merging: we actually mean the _modal_ dialog box here right? So I think the spelling on this function should be:
```suggestion
var showWarningsModal = function () {
```

Issue #1211, opened 2025-09-03
Opt groups for Expression Datasets are not displaying in Top menu


kdahlquist:
The Opt groups are only displaying in the side panel and are not shown in the Top menu.

Amelie1253:
Fixed! I forgot to update the createHTMLforDataset function which create the options for node coloring. Since these are technically <li> elements and not <option> in a <select> like in the sidebar, I couldn't add opt groups but added some headings to mimic opt groups and show that differentiation.

MilkaZek:
I opened a [PR (#1237)](https://github.com/dondi/GRNsight/pull/1237) with a fix implementing the process described above and while it fixed the issue on my local branch, the PR is failing one of the Node.js CI / build workflow checks. I'm not sure if it's something wrong with my code or Coveralls or our workflow files, but am hoping to figure that out this week and/or go to @dondi's office hours.

kdahlquist:
Actually, I have a cosmetic request, if possible:
1. Can the text for the opt-group headings be black instead of gray?
2. Can you put some padding on the left side so that the opt-group headings aren't flush with the left side of the drop-down menu?

dondi:
As a side note, @MilkaZek asked about the existing framework for finding CSS then encountered the _.styl_ file format; @dondi responded with how this was an early decision of the team predating existing web app frameworks (Jade then Pug)

kdahlquist:
Confirmed that this is working in beta 7.3.0, so closing.

PR #1237, opened 2025-09-24
Displaying Expression Database Opt Groups in Top Menu - for Issue #1211

MilkaZek:
Adapted a fix from our old amelie-milka-863 branch to make sure the expression database opt groups show in the top menu with headings separating the `<li>` elements (because unlike the sidebar, they aren't `<option>` elements in a `<select>`). The headings are also cleared each time the app is updated so that the duplication issue we once had with the opt-groups doesn't occur again.


dondi (approved):
LGTM

Issue #1192, opened 2025-04-08
Need a warning to the user when exporting workbook if a gene does not have a degradation/production rate


kdahlquist:
We discovered today that when a user exports data to an Excel workbook with degradation rate and prodution rate sheets, a gene will be missing from those sheets if there is no value for the degradation rate/production rate.  It took us a while to figure out why GRNmap wouldn't run.  We need to remedy this in two ways:
1. Give a warning to the user upon export that says: "<gene id> does not have a degradation rate in our database.  A degradation rate will need to be supplied to run this workbook in GRNmap." and/or "<gene id> does not have a production rate in our database.  A production rate will need to be supplied to run this workbook in GRNmap."
2. The gene ID needs to included in the degradation_rates and production_rates sheets, but without a value.

I am going to look into a couple more tweaks for this that will take some additional time:
1. I'm going to try to find a longer list of degradation rates/production rates so that this is less likely to happen in the future.
2. We may end up providing a replacement value (instead of null), if it is missing from our database.  We would then maybe need to have some logic where the user is asked whether they want to use the replacement value.

dondi:
Because this involves export code, @Amelie1253 and @MilkaZek have been assigned for coordination, with @ntran18 taking the lead

The database part though is definitely in @ntran18’s wheelhouse

ntran18:
I added a PR #1195 for code review

dondi:
May be a follow-up after resolution of #1120, #1164, #1165

ntran18:
For the expression database, I believe all of us are getting the data directly from [GRNsight Box Folder > GRNSight Expression](https://lmu.box.com/s/4s6yw8009fmc9pq8zn17uu4iyitdskm2)

ntran18:
I believe PR #1195 added a fix to this but the scenario was:
1. User loads genes from database
2. Exporting the graph to Excel
3. Selected degradation rate and/or production rate option
4. If any genes don't have degradation rate and/or production rate -> add a warning message

However, @kdahlquist tested it with a file that doesn't have degradation/production rate, which is a different scenario. I added a fix to this in PR #1230

kdahlquist:
I'm reviewing the text of the Warning given to the user in this scenario.  The top text reads:

[...]

PR #1243, opened 2025-09-26
#1192 Modify the export warning message to desired message

ntran18:


web-client/public/js/export-constants.js
```diff
@@ -6,8 +6,8 @@ module.exports = {
             errorDescription:
                 "GRNsight has detected that there are missing degradation rates in the exported" +
                 " workbook's 'degradation_rates' sheet. These degradation rates are missing in our database." +
-                " A degradation rate will need to be supplied to run this workbook in GRNmap." +
-                " The missing genes are: " +
+                " A degradation rate will need to be supplied to use this workbook as an input file for GRNmap." +
+                " The missing values are for the genes: " +
                 missingGenes,
         }),
```
dondi, line 12: Purely a style note for future long strings—no need to change this now, but just file away for the future: we do the `+` concatenation to avoid exceeding the maximum line length, which is more important now that we have automated formatting. For this purpose, I have come to prefer `join(" ")` because it feels more structured and it alleviates the programmer from remembering whether to add a leading or trailing space to the string

So, expressions like this would one day look like:
```javascript
errorDescription: [
    "GRNsight has detected that there are missing degradation rates in the exported",
    "workbook's 'degradation_rates' sheet. These degradation rates are missing in our database.",
    "A degradation rate will need to be supplied to use this workbook as an input file for GRNmap.",
    `The missing values are for the genes: ${missingGenes}`
].join(" ")
```

Note also the preference to use template strings when the string needs to be built from variables/expressions (you use this in this PR so that’s good!)

Between template strings and `.join(" ")`, I have found that there is virtually no need for string concatenation via `+` anymore. Purely an aesthetic and style choice, but as mentioned, it feels more structured than `+` and I think automated formatters work with them better

dondi (approved):
Minor style note for _future_ work—not a big deal at all, but figured I’d document the preference while I was thinking about it

LGTM

Issue #1212, opened 2025-09-04
When no graph is loaded menu Export > Export Data > to Unweighted GraphML should NOT be available


kdahlquist:
I am going through the client-side testing document (#1158) on the wiki  and discovered this issue.

When no graph is loaded menu Export > Export Data > to Unweighted GraphML should NOT be available.  The rest of the options are properly grayed out, but not this one weirdly enough.

Amelie1253:
I deleted the id `unweightedGraphmlContainer` and checked that the id isn't used elsewhere. It doesn't seem to touch anything else and allows the Unweighted GraphML option to be properly grayed out.

Amelie1253:
As referenced in **#1244,** I thought I had checked but I missed the reference to the id `unweightedGraphmlContainer` in grnstate where it is then labeled `EXPORT_TO_UNWEIGHTED_GML_MENU` in constants. `EXPORT_TO_UNWEIGHTED_GML_MENU` is then referenced in update-app where it listens for clicks. On each click, it checks the grnstate mode and if the mode is not a GRN, then it disables the Unweighted GraphML option. Otherwise, it removes the disabled class that was initialized on the `unweightedGraphmlContainer` id. This led to the issue of the Unweighted GraphML option not being disabled because the grnstate mode is initialized to GRN and removes the disabled class immediately. I have changed the initial value of the grnstate mode to null but this resulted in this error. 

<img width="692" height="209" alt="Image" src="https://github.com/user-attachments/assets/43417606-413c-4177-90e9-5502ae3e4e42" />

For future work, I will look into an alternative solution such as checking something other than the grnstate mode, perhaps something that records whether or not something is being displayed and can be exported and then checking the grnstate mode.

kdahlquist:
The odd thing about this is that all of the other options for exporting _are_ properly grayed out, it's just this one that is not. Would you be able to investigate why those are working properly?

Amelie1253:
Sorry for the late response! Yes, the other options do not have the id unweightedGraphmlContainer, only the Unweighted GraphML option has the id. All options started disabled (grayed out) but since this id is used to check for the grnstate mode, the disabled class is specifically removed from the Unweighted GraphML option.

That does bring up a good question of when other options' disabled classes get removed. Perhaps, the Unweighted GraphML option should follow suit with when those options' disabled classes get removed. I will look into this!

dondi:
[...]

PR #1244, opened 2025-10-01
#1212 Unweighted GraphML option properly grayed out

Amelie1253:


dondi (commented):
Although the ID has been removed from the .pug file, it is still referenced in the code so I think it’s important to look at that in order to gain an understanding of why eliminating that ID seems to have fixed the bug

The ID is referred to here, so I suspect that understanding what this code does/did will help explain the role of this ID. once that understanding is gained, if appropriate we can then remove this block too: https://github.com/dondi/GRNsight/blob/569c7bbaaa6e3baea7477b015815b5fefa9381c9/web-client/public/js/update-app.js#L577

PR #1245, opened 2025-10-01
Opt Groups Cosmetic Change - for Issue #1211

MilkaZek:
I updated the opt group labels so that the left side has more padding and the text is now black instead of gray.


dondi (approved):
LGTM

Issue #1214, opened 2025-09-04
Grid layout button/menu item should be disabled when no network graph is loaded


kdahlquist:
I think we need to review the behavior of the Grid Layout button when no graph is loaded.  Currently the button is available to click when no graph is loaded, but the default behavior when you load a graph is Force Graph layout, even if you previously clicked the Grid Layout button.

If you click the Grid Layout button before a graph is loaded, the force graph parameter sliders will gray out and become unavailable, but the graph still loads as a force graph.

I think maybe the simplest way to set the behavior is to only allow force graph layout when a user first loads the graph.  In this case, the Grid Layout button on the side bar and menu would be disabled until a graph is loaded. Once a graph is loaded, then the Grid Layout button/menu item would be available.

Alternating between force graph and grid is working properly once a graph is loaded, so this seems the simplest way to deal with this problem.

MilkaZek:
Grid layout button and menu item are both disabled until a graph is loaded now -  fixed in [PR #1247](https://github.com/dondi/GRNsight/pull/1247)!

PR #1247, opened 2025-10-01
Disabled Grid Layout button/menu item when no graph is loaded for issue #1214

MilkaZek:
When no network graph is loaded, the Grid Layout button and menu item are now both disabled until a graph is loaded. CSS styling has been applied to show that those items are disabled to begin with as well.


dondi (approved):
LGTM

Issue #1239, opened 2025-09-24
Establish how the React app will invoke the existing API


dondi:
Continuing our buildout, we should establish how the React app communicates with the existing API

dondi:
@ceciliazaragoza has set this up to interact with the API locally, and now needs to ensure that it works in both local and deployed environments #1265

PR #1255, opened 2025-10-17
#1239: Establish how the React app will invoke the existing API

ceciliazaragoza:
#1239 
- can `console.log()` all of the demos.
- can access the local server from multiple hosts with an updated CORS configuration. 
- added `constants.js` to hold constant values like the demo names.
- put the `theme` for Grommet styling in a separate `theme.js`.


ceciliazaragoza:
To satisfy the coveralls requirements, I may need to add some tests to our current test suite because I am modifying code in `web-client-classic`.

dondi (approved):
LGTM—we will let the Coveralls callouts slide for now and look into ways to address them at a later time

Issue #1213, opened 2025-09-04
Export > Print not working for PPI networks


kdahlquist:
Load a PPI network (either Demo #5 or load from database or open file) and select menu item Export > Print.  The print dialog does not open.

MilkaZek:
I was trying to isolate where the issue is stemming from but wasn't able to figure it out yet, so I will try to keep digging. I first tried to see if the print button just wasn't getting re-enabled once the demo was loaded but I don't think that's the issue. Will try to stop by @dondi's office hours after the Career Trek!

MilkaZek:
Ok so I think what was happening is that with the PPIs, there was code for the hideEdgeWeightOptions that removed the `startDisabled` class from weighted export elements and added `disabled` - but the print window functionality was checking if all the things in the `startDisabled` class were enabled. This didn't work for PPIs, so I changed it so that it checked if there was a current workbook open, and this solves the issue. I've opened a PR [(PR #1257)](https://github.com/dondi/GRNsight/pull/1257) with this implementation.

dondi:
@dondi walked @Amelie1253 and @MilkaZek through the testing code base and talked briefly about TDD (writing tests first) so we’ll see if we can go through this process firsthand with this ticket

MilkaZek:
I pulled from beta and created a new branch but am having some issues hosting GRNsight locally now - when I run it on localhost none of the changes I make in my code editor are reflected in the locally run webapp. I don't want to open a PR without testing and making sure it works locally, so I'll have to stop by office hours for help in deducing what's wrong since I haven't yet figured it out.

MilkaZek:
I have a pull request implementing the same change as before - [PR #1256](https://github.com/dondi/GRNsight/pull/1261)

PR #1257, opened 2025-10-22
Fixed print window functionality for PPIs - for Issue #1213

MilkaZek:
I updated the print window functionality so that it checked if there was a current workbook open, and this makes the print button work for all demos, including PPIs.


dondi (commented):
Code change looks good, but due to the recent big folder rename to _web-client-classic_, PR merge will be cleaner if @MilkaZek creates a new branch from a fresh `beta` pull. The code change is a single line so isn’t too costly to redo

PR #1259, opened 2025-10-24
#1239: Fix for beta server config

ceciliazaragoza:
- Removed callback func in app.js that returned an error if there was no origin -> `!origin`
- Callback func not needed since dynamically return the correct port in the `module.exports` in `helpers.js`.


dondi (approved):
LGTM…laissez les bons temps rouler!

PR #1263, opened 2025-10-31
Added wrap-around styling for demo name in export to excel window

MilkaZek:
I don't think there's an official issue/ticket for this but I remember during our last meeting that Dr. Dhalquist mentioning wanting to fix how the demo name starts below the select button (in the export to excel pop-up) when the browser gets too narrow. I added wrap styling for when the browser is too narrow - see attached screenshot for what it looked like before vs now. If we want it to look differently I can change that as well but this is what I came up with!

Screenshot 2025-10-31 at 4 50 25 PM
Screenshot 2025-10-31 at 4 42 01 PM

### DIFF
--- a/web-client-classic/public/stylesheets/grnsight.styl
+++ b/web-client-classic/public/stylesheets/grnsight.styl
@@ -94,7 +94,7 @@ p
   margin-bottom: 15px
 
 form label, form span
-  margin: 5px
+  margin: 0 5px 5px 5px
 
 .navbar
   top: 1px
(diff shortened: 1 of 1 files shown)


web-client-classic/public/stylesheets/grnsight.styl
```diff
@@ -618,6 +618,21 @@ path.mousezone
 .export-excel-workbook-sheet-option-subheader
   margin-left: -20px
 
+.export-radio-group li
+  display: flex;
+  align-items: center;
```
dondi: Per UA (user acceptance) feedback, let’s change this to `align-items: flex-start`

```suggestion
  align-items: flex-start;
```

This change usually has a side-effect where the radio will then look off when the item is just one line, so…

dondi (commented):
Code looks good but we have some UA feedback that we can address here

dondi (approved):
LGTM

Issue #1266, opened 2025-11-05
Investigate supporting different _.prettierrc.js_ configurations between Classic and React


dondi:
While we are transitioning to React, we have run into the need to have different auto format configurations. We either need to find a way to support different ones between the two code bases, or we may have to transition _web-client-classic_ to new rules (primarily the indent size)

ceciliazaragoza:
This pertains to PR #1274.

ceciliazaragoza:
This also relates to issue #1411 and PR #1420 .

dondi:
@ceciliazaragoza will look into a combination of rebase, cherry-pick, and/or manual recommits in other to refocus the PR strictly on the changes pertaining to this issue

ceciliazaragoza:
Did manual recommits to ensure only relevant changes included on PR.

PR #1274, opened 2025-11-12
#1266: React Prettier Config

ceciliazaragoza:
- Added prettier config to web-client with npm install.
- Formatted all React files with prettier.
- Added .prettierignore to both GRNsight/ and GRNsight/web-client so that correct files can be ignored by prettier when open up GRNsight or GRNsight/web-client in separate VScode windows.


dondi (approved):
LGTM

ceciliazaragoza:
This pertains to issues #1266.

Issue #1265, opened 2025-11-05
Configure prod server so that can serve web-client and web-client-classic


ceciliazaragoza:
Currently, the `web-client` demo API setup can only access the local server. Configure the web-client API so that it can access the prod server. Ensure that the prod server can serve both web-client and web-client-classic at the same time.

ceciliazaragoza:
I am configuring the deployed react app to be able to access the prod server. I think I may be running into a larger question of where the requests from the react app should originate from.
- The request from the react app to the server originates from `https://dondi.github.io`
- However, in prod, the requests originate from `https://grnsight.lmucs.org`

In web-client-classic, GRNsight is in an iframe at `https://grnsight.lmucs.org` within the page `https://dondi.github.io`. But the react app is currently directly being served at `https://dondi.github.io`.

I am going to put the app in an iframe so that I can begin simulating the same framework as prod. Which port should I configure the react app to serve at for `https://grnsight.lmucs.org`? Should it be 3000 or another port?
- I could also allow `https://dondi.github.io` to access the server, but that would stray from our current setup.

My draft PR can be viewed here: #1279
<img width="1498" height="905" alt="Image" src="https://github.com/user-attachments/assets/da471dc5-357d-4ea1-8d62-7f367186ae4e" />

PR #1279, opened 2025-11-13
#1265: Configure Prod React App to Access Server

ceciliazaragoza:
- Make API_URL dynamic based on the environment.
- Environment handled by vite. When working locally, vite is in DEV mode, or environment, and when deployed, vite is in PROD mode, or environment.
  - When run `npm run build` and vite build is created, that build is in PROD mode.
  - We can dynamically have environment variables depending on the environment which are used to create API_URL
- The dondi.github.io was added as a CorsOrigin so that the react app can access the server.


dondi (approved):
LGTM

PR #1281, opened 2025-11-16
Fix for double downloading when exporting to excel and changing different expression sources

ntran18:


dondi:
Oh, and thank you Copilot for catching that `console.log` 🤖

PR #1282, opened 2025-11-16
Disable option to export a "network_optimized_weights" sheet for an unweighted network

ntran18:


dondi (commented):
General approach looks good, but a few refinements are called for to improve robustness and consistency

dondi (approved):
Thanks for the fixes! LGTM

Issue #1240, opened 2025-09-24
Port the graph display to React


dondi:
There will be exploration here to determine how much of the existing D3 implementation can be ported over directly vs. new work due to how the available technology stack may have changed in the interim

PR #1307, opened 2026-01-21
#1240: Graph + Demo Dropdowns

ceciliazaragoza:
- The unweighted, weighted, and PPI graphs can be displayed with colored Bezier curve eges and varying marker-ends. 
- The gray threshold, charge, link distance, and edge coloring can be customized in the navbar and sidebar. 
- There are repressor marker ends and arrowheads, and the repressors have both vertical and horizontal arrowheads.
  - However, further work needs to be done to ensure that the switch between horizontal and vertical repressor ends happens at the correct place on the node, since currently the vertical repressor marker end still displays towards the right and left side of the top/bottom of the node.
- Improvements to Demo dropdowns across navbar and sidebar so that demo can be chosen in different places.
  - Further work needs to be done to customize dropdowns to be exactly like web-client-classic. 
- Because changes were made to the graph, some tests began to fail and are commented out. Tests will be updated in a future PR to ensure that the graph is being tested correctly.

### DIFF
--- a/web-client/src/components/Graph.jsx
+++ b/web-client/src/components/Graph.jsx
@@ -2,6 +2,7 @@ import { useEffect, useRef, useContext, useState } from "react";
 import * as d3 from "d3";
 import { GrnStateContext } from "../App";
 import { getDemoWorkbook, getDemoEndpoint, getNetworkMode } from "../services/api";
+import ScaleAndScroll from "./ScaleAndScroll";
 import {
   BOUNDARY_MARGIN,
   ZOOM_DISPLAY_MINIMUM_VALUE,
@@ -28,13 +29,14 @@ import {
   calcMaxWeight,
 } from "../helpers/graphHelpers";
 import { createEdgeMarker } from "../helpers/markerHelpers";
-
 import "../App.css";
 
 export default function Graph() {
   const svgRef = useRef(null);
   const containerRef = useRef(null);
   const simulationRef = useRef(null);
+  const zoomRef = useRef(null);
+  const zoomContainerRef = useRef(null);
 
   // The workbook or sheetType are not needed in global state outside of Graph, so keep them local
   const [workbook, setWorkbook] = useState(null);
@@ -43,25 +45,17 @@ export default function Graph() {
   const [maxWeight, setMaxWeight] = useState(0);
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState(null);
+  const [zoomScale, setZoomScale] = useState(null);
 
   const {
     colorOptimal,
-    setColorOptimal,
     linkDistance,
-    setLinkDistance,
     charge,
-    enableNodeColoring,
-    setEnableNodeColoring,
-    logFoldChangeMax,
-    edgeWeightVisibility,
     demoValue,
-    setDemoValue,
-    adaptive,
-    setAdaptive,
     networkMode,
     setNetworkMode,
     grayThreshold,
-    setGrayThreshold,
+    zoomPercent,
   } = useContext(GrnStateContext);
 
   // Load workbook data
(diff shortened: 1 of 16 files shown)


web-client/src/components/Graph.jsx
```diff
@@ -58,6 +68,12 @@ export default function Graph() {
     getDemoWorkbook(demoEndpoint)
       .then(data => {
         setWorkbook(data);
+        setSheetType(data.sheetType);
+        console.log("data", data);
```
dondi: I’m thinking we won’t really want to keep this debug statement?

web-client/src/helpers/graphHelpers.js
```diff
@@ -132,6 +173,8 @@ export function createPath(d, width, height) {
   // However, this values means that nodes are within 202 pixels of each other, so will need to adjust this
   // When set orthOffset to 0, then creates issues with paths going through border of boudning box
   const isStraightLine = orthoOffset < 0.5;
+  const isRepressor = d.value < 0;
+  const strokeWidth = d.strokeWidth || 2;
```
dondi: The `||` operator here will also trigger for a `d.strokeWidth` of `0`—is that intentional here? These days, you can distinguish between `0` vs. `null` or `undefined` with `??` rather than `||`. Would it be appropriate to use that here, or stick with `||`? (meaning you _do_ want a value of `0` to be replaced by `2`)

web-client/src/helpers/graphHelpers.js
```diff
@@ -218,18 +275,90 @@ function straightLineBoxIntersection(
...
+    " " +
+    xRotation +
+    "," +
+    largeArc +
+    "," +
+    sweep +
+    " " +
+    x2 +
+    "," +
+    (y2 + offset)
```
dondi: For string concatenations that are ported from the classic code, you might want to consider converting these to template strings—they’re more modern and may be more readable

web-client/src/helpers/markerHelpers.js
```diff
@@ -0,0 +1,348 @@
...
+    8: 3.5,
+    9: 4,
+    10: 4.5,
+    11: 5,
+    12: 5,
+    13: 5.5,
+    14: 6,
+  };
+
+  console.log("Creating horizontal repressor marker. d printout:", d);
```
dondi: Another potentially expendable debug statement, but if this is still in active use for a bit, we can keep it until it’s no longer needed

web-client/src/helpers/markerHelpers.js
```diff
@@ -0,0 +1,348 @@
...
+      return d.strokeWidth;
+    })
+    .attr("markerUnits", "userSpaceOnUse")
+    .attr("orient", 180)
+    .append("rect")
+    .attr("width", 25 + d.strokeWidth)
+    .attr("height", d.strokeWidth)
+    .attr("rx", 10)
+    .attr("ry", 10)
+    .attr("style", "stroke:" + d.stroke + "; fill: " + d.stroke + "; stroke-width: 0");
```
dondi: You know at this point 😊

web-client/src/helpers/markerHelpers.js
```diff
@@ -0,0 +1,348 @@
...
+    12: 5,
+    13: 5.5,
+    14: 6,
+  };
+
+  console.log("Creating horizontal repressor marker. d printout:", d);
+
+  defs
+    .append("marker")
+    .attr("id", "repressorHorizontal" + selfRef + "_StrokeWidth" + d.strokeWidth + minimum)
```
dondi: Another template string candidate

dondi (approved):
Phew, for the amount of code involved, there really wasn’t much to say 😅

As such, please take note of the comments, but I’ll go ahead and merge so you can start fresh from `beta` again. Just keep an eye out for the things noted in the PR and revise as you go in future work

Very awesome, congrats on this big milestone!

Issue #1310, opened 2026-01-22
Move Controls and Zoom Scrollbar


ceciliazaragoza:
After implementing #1309, the temporary zoom controls of pinching two fingers on the touchpad had to be removed. Implement the move controls with a direction pad and zoom scroll bar like in web-client-classic.

PR #1319, opened 2026-01-29
#1310: Zoom and Move with D-pad

ceciliazaragoza:
- The zoom bar on the graph can be used to zoom between 25% and 200%.
- The d-pad on the graph can be used to move the graph.
- Further work: 
  - Ensure the graph can be moved by dragging the graph background and that the zoom value can be typed in by the user. 
  - Since I am using a simplified way of calculating the zoom value by using a Zoom Percent, ensure that zoom bar starts in the center where the gray vertical line is.
  - Adjust the styling of the d-pad and zoom to be exactly like classic.

### DIFF
--- a/web-client/src/components/Graph.jsx
+++ b/web-client/src/components/Graph.jsx
@@ -105,7 +105,6 @@ export default function Graph() {
 
     const defs = svg.append("defs");
 
-    // Create zoom container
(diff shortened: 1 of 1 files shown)


dondi (approved):
LGTM—a bunch of already-in-`beta` changes were noted, but those were easy to distinguish based on the files involved

Issue #1276, opened 2025-11-13
Demo 3 should not display expression data on nodes


kdahlquist:
The intention behind demo 3 is that it does not display expression data on the nodes (see documentation webpage).  I'm not sure when this changed, but we need to make it do that again.  So demo 3 should default to node coloring turned off.

We don't store the Schade data as one of our options in the expression database because it's not a fully public dataset.  We only had permission to share the expression data for the genes in the demo network.

MilkaZek:
The method discussed during our meeting worked - we eliminated the workbook expression data contents from demoWorkbook3 in demo-workbooks.js, but had to keep the `expression` property as an empty property. [Fixed in PR #13221](https://github.com/dondi/GRNsight/pull/1321)

PR #1321, opened 2026-01-30
Removed workbook expression data and concentration_sigmas for Demo 3 - fixes Issue 1276

MilkaZek:
We removed the expression data and concentration_sigmas for Demo 3 in demo-workbooks.js. When we tried to get rid of the expression object completely, the browser crashed when we tried to load Demo 3, so we kept expression as an empty object. Fixes [Issue #1276](https://github.com/users/dondi/projects/2/views/1?sliceBy%5Bvalue%5D=MilkaZek&pane=issue&itemId=138978664&issue=dondi%7CGRNsight%7C1276)

### DIFF
--- a/server/controllers/demo-workbooks.js
+++ b/server/controllers/demo-workbooks.js
@@ -3035,8 +3035,7 @@ var demoWorkbook3 = function (path, res, app) {
             },
         },
         twoColumnSheets: {},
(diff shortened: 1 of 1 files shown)


dondi (approved):
LGTM

Issue #1314, opened 2026-01-22
Length at which edge straightens in React version


kdahlquist:
I'm looking at the Classic and React versions side by side.  It seems that for vertical edges, in Classic, the edge becomes straight at 5 cm, but in the React version, it straightens at ~4.5 cm.

For the horizontal direction, Classic straightens at ~4.5 cm and React straightens at ~3.5 cm.

For a ~45 degree angle, Classic straightens at 5 cm and React straightens at about 4 cm.

I also noticed that the arrowhead end marker is not rotating properly.  See below:

<img width="940" height="1032" alt="Image" src="https://github.com/user-attachments/assets/28c78e85-e16d-4073-a5b2-b216399cb124" />

PR #1331, opened 2026-02-05
#1314: Length at which edge straightens and repressor orientation

ceciliazaragoza:
- Edges straighten at consistent distance away from node.
- Adjusted logic to calculate intersection with node back to `web-client-classic` logic and no longer using Bernstein Basis polynomials.
- Fixed lingering issue from classic with gap on right side of target node when source node to right of target node. This required moving a `LEFT_ADJUSTMENT` constant that was being added incorrectly to arrowheads.
- Repressor edges switch between vertical and horizontal at the correct position on node.

### DIFF
--- a/web-client/src/components/Graph.jsx
+++ b/web-client/src/components/Graph.jsx
@@ -225,7 +225,7 @@ export default function Graph() {
 
     function dblclick(event, d) {
       d.fx = null;
-      d.fy = null;``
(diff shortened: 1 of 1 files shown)


dondi (approved):
I went ahead and removed the stray template string, so now, LGTM 👌🏽

Issue #1275, opened 2025-11-12
File naming when exporting and always export "network" sheet


kdahlquist:
As per #1174, following @Amelie1253 's suggestion of breaking out the file naming of files Exported to Excel as a separate issue.

kdahlquist:
I closed the previous issue (#1217) because we will use this issue to do a general audit of file-naming.  I think the filenames need to be simplified.  I think originally, they were modeled the names that display in the menu bar when you either load a demo or load from the database.  What is displayed in the menu bar should not change because it is useful information for the user.  However, we don't need that exact phrasing to be part of the exported filename.

Here is the simplified scheme that I propose:
- Begin the filename with either GRN or PPI, depending on whether it is a GRN or PPI.
- followed by an underscore
- followed by how many genes/proteins and edges as in "6-genes_8-edges" or "6-proteins_8-edges"
- follwed by an underscore
- followed by either weighted or unweighted according to the rules above
- followed by an underscore
- followed by the expression data source, only if the user is exporting expression data.  For example, "Dahlquist-2018-data" or "Apweiler-2012-data".  If it is user-uploaded data then it would be "user-data"

Note that we are avoiding all special characters except for underscore and hypen.  No spaces, parentheses, commas, etc.

So here are some examples:
- PPI_25-proteins_30-edges_unweighted
- GRN_15-genes_32-edges_unweighted
- GRN_6-genes_8-edges_weighted
- GRN_7-genes_14-edges_unweighted_Dahlquist-2018-data
- GRN_8-genes_16-edges_weighted_Apweiler-2012-data
- GRN_9-genes_18-edges_unweighted_user-data

These rules should apply whether the file is user-uploaded, loaded from database, or a demo.  So, this is going to de-couple the exported filename from what is displayed in the menu bar.  We are not including the network source info in the export filename.

This also means that when a demo is exported, it's not going to say "Demo #" in the filename, it will just follow the rules above.  

If the demos are exported, and the user does not change the expression data, then those files would be:
- Demo 1: "GRN_15-genes_28-edges_unweighted_Dahlquist-2018-data"
- Demo 2: "GRN_15-genes_28-edges_weighted_Dahlquist-2018-data"
- Demo 3: "GRN_21-genes_31-edges_unweighted"
- Demo 4: "GRN_21-genes_31-edges_weighted_Schade-2004-data"
- Demo5: "PPI_18-proteins_81-edges_unweighted"

[...]

PR #1335, opened 2026-02-06
#1275 New File Naming Scheme Implemented

Amelie1253:


web-client-classic/public/js/upload.js
```diff
@@ -87,23 +87,40 @@ export const upload = function () {
...
+            if (source === "userInput") {
+                source = "user-data";
+            }
+        }
+        if (mode !== "grn") {
+            mode = "PPI";
+        }
+        if (mode != null && genes != null && edges != null && type != null) {
+            filename =
+                mode.toUpperCase() + "_" + genes + "-genes" + "_" + edges + "-edges" + "_" + type;
```
dondi: Template strings are pretty well-established now, and these are generally more readable than string concatenation. Compare the two versions:

```suggestion
                `${mode.toUpperCase()}_${genes}-genes_${edges}-edges_${type}`;
```

web-client-classic/public/js/upload.js
```diff
@@ -87,23 +87,40 @@ export const upload = function () {
...
         }
 
+        if (source) {
+            filename += "_" + source;
```
dondi: Two possibilities for this one:

```suggestion
            filename += `_${source}`;
```

…or…

```suggestion
            filename = `${filename}_${source}`;
```

Maybe you can look at this as a pair and discuss the differences in clarity/readability, then make a choice. More eyes/opinions can help sometimes!

web-client-classic/public/js/upload.js
```diff
@@ -87,23 +87,38 @@ export const upload = function () {
...
         var filename = $("#fileName").text();
+
         var currentExtension = filename.match(/\.[^\.]+$/);
         if (currentExtension && currentExtension.length) {
             filename = filename.substr(0, filename.length - currentExtension[0].length);
         }
-        if (suffix) {
-            filename = filename + "_" + suffix;
+        if (Object.keys(grnState.workbook.expression).length > 0) {
+            var source = $("input[name=expressionSource]:checked")[0].value;
```
dondi: Just realized this on this reading: the `source` variable is being defined _inside_ the `if` clause. i.e., When the condition is `false`, the `source` variable technically is never defined

This works OK in JavaScript because of its specific behavior of “hoisting” variables, but in general isn’t something that all languages do. As such, where possible, I prefer not to write code with this kind of dependency on a very specific language feature

Can you think of a why to revise this code so that `source` is defined at the level of the function instead of inside the `if` clause? It will make the code look more consistent with how it might be written in other languages that don’t do hoisting

web-client-classic/public/js/upload.js
```diff
@@ -87,23 +87,38 @@ export const upload = function () {
...
         }
-
-        return filename + "." + extension;
+        if (mode !== "grn") {
+            mode = "PPI";
+        }
+        if (mode !== null && genes !== null && edges !== null && type !== null) {
+            filename = `${mode.toUpperCase()}_${genes}-genes_${edges}-edges_${type}`;
+        }
+        if (source) {
```
dondi: So to some readers, this reference to `source` will stick out because `source` is declared in a block that technically this line can’t “see.” Again, JavaScript specifically behaves in a way that this is OK, but that isn’t how other languages operate. For example, Java wouldn’t like this structure (to name an example)

dondi (commented):
Almost there! There’s just one note remaining and it’s a bit of a JavaScript deep cut that illustrates how sometimes, just because a language permits something doesn’t mean it should be leveraged (because it actually wouldn’t be OK in other languages). Hoisting is an example of this and ideally we can rewrite this portion in a way that lines up with other languages better

dondi (approved):
Thanks for the quick update! LGTM

Issue #1296, opened 2025-12-10
Consistent Cell A1 nomenclature for GRN and PPI "network" and "network_optimized_weights" sheets and fix Warning message


kdahlquist:
As I was testing #1273, I figured out that there is a discrepancy between the nomenclature for cell A1 in the "network" or "network_optimized_weights" sheets between a GRN and PPI.
- There is a discrepancy between the Warning modal and the required text in cell A1 for a gene regulatory network's "network" or "network_optimized_weights" sheet.  In the documentation and functionality, cell A1 should read "cols regulators/rows targets" with NO space character after the "/".  If the cell A1 is named this way, the graph loads.  However, in the Warning modal, it says it should be "cols regulators/ rows targets" with the inclusion of a space character after the "/".  If I do include the space character, it actually triggers the warning.  *So, the Warning modal message needs to be corrected.*
- However, there is no consistency between GRN and PPI where "cols protein1/ rows protein2" WITH a space character is required to load while "cols protein1/rows protein2" does not load.  In this case the Warning modal message is currently correct. This difference is the origin of me thinking that GRNsight was not able to detect the network type based on cell A1.  *I want to get rid of that space character after the "/" so that the syntax for cell A1 is consistent between a GRN and PPI.  Both the functionality and the Warning modal message would need to be updated.*

kdahlquist:
Pullling this comment from #1204: what happens whent there is an "optimization_parameters" sheet with a "networkType" parameter and the network types don't match?  Is the "networkType" parameter even used?

dondi:
@MilkaZek and @pichuT looked into this issue with @dondi and we have a clarifying question: in the intended-format file, cell A1 says `protein 1/protein 2` (i.e., no “cols” nor “rows”):


<img width="334" height="217" alt="Image" src="https://github.com/user-attachments/assets/8cba01fb-a4d9-4600-9315-7527d5e0cba9" />


Shall we then be standardizing to exactly this label (`protein 1/protein 2`) or should this label actually be (`cols protein 1/rows protein 2`)?

kdahlquist:
Ack sorry, it should be `cols protein1/rows protein2`.  I'll fix the Documentation to match this when the fix is in (and do a lot of testing!).

When I export Demo 5, it shows `cols protein1/ rows protein2` with the space character.  So it looks like we have even more of a mix if there are some places where it says just `protein 1/protein 2`.

MilkaZek:
[...]

PR #1337, opened 2026-02-11
Fixed cell A1 naming consistency for Issue #1296

MilkaZek:
Edited Cell A1 constants and warning message references to standardize naming conventions and logic to "cols protein1/rows protein 2" for PPIs, matching it to the "cols regulators/rows targets" template for GRNs.


ntran18:
I looked at Milka's issue, and it's similar to what I did before, so I wonder why it failed. The test is now passing for Milka's PR. Old test files have cell A1 = "rows genes affected/cols genes controlling". And since she is now adding errors instead of warnings, any test that checks errors has failed because a new error was introduced.

PR #1368, opened 2026-02-24
Fixing warnings issue for two column sheets (headers & importing) #1192

JustAlex3125:
This PR added fixes for:
- Incorrect headers in additional sheets (import)
- Missing headers in additional sheets (import)
- Tests for incorrect data values


server/controllers/workbook-constants.js
```diff
@@ -274,6 +274,26 @@ module.exports = {
                 " GRNsight defaults to Saccharomyces cerevisiae.",
         },
 
+        additionalSheetIncorrectColumnHeaderWarning: function (sheetName, expectedA1, expectedB1) {
+            return {
+                warningCode: "INCORRECT_COLUMN_HEADER",
```
ntran18: For easier debugging in the future, you can also make the `warningCode` explicit for each sheet

test/test.js
```diff
@@ -693,6 +626,46 @@ var incorrectMSEHeaderWarning = function (input, frequency) {
...
+            x => x.errorCode === "INCORRECT_COLUMN_HEADER"
+        ).length;
+    }
+    assert.equal(additionalSheetIncorrectColumnHeaderWarningCount, frequency);
+};
+
+var additionalSheetMissingColumnHeaderWarning = function (input, frequency) {
+    var sheet = xlsx.parse(input);
+    var workbook = parseAdditionalSheet(sheet);
+    var additionalSheetMissingColumnHeaderWarningCount = 0;
```
ntran18: Same comment like above ^

test/test.js
```diff
@@ -693,6 +626,46 @@ var incorrectMSEHeaderWarning = function (input, frequency) {
...
+    additionalSheetIncorrectColumnHeaderWarningCount += workbook.meta.errors.filter(
+        x => x.errorCode === "INCORRECT_COLUMN_HEADER"
+    ).length;
+    if (workbook.meta2.errors !== undefined) {
+        additionalSheetIncorrectColumnHeaderWarningCount += workbook.meta2.errors.filter(
+            x => x.errorCode === "INCORRECT_COLUMN_HEADER"
+        ).length;
+    }
+    assert.equal(additionalSheetIncorrectColumnHeaderWarningCount, frequency);
+};
```
ntran18: To improve test readability and provide more helpful error messages when a test fails, we can add `sheetName` to the parameter and add more `assert.exists` check. Adding more `assert.exists` check can ensure we catch exactly where the workbook structure is breaking (e.g., if a sheet or the warnings array is missing) before checking the warning count.

```suggestion
const additionalSheetIncorrectColumnHeaderWarning = function (input, frequency, sheetName) {
    const sheet = xlsx.parse(input);
    
    const workbook = parseAdditionalSheet(sheet);
    
    assert.exists(workbook.twoColumnSheets, "Expected two column sheets to exist on workbook");
    assert.exists(
        workbook.twoColumnSheets[sheetName],
        `Expected ${sheetName} sheet to exist on workbook`
    );
    assert.exists(
        workbook.twoColumnSheets[sheetName].warnings,
        `Expected warnings array to exist on ${sheetName} sheet of workbook`
    );
    assert.equal(workbook.twoColumnSheets[sheetName].warnings.length, frequency);
    assert.isTrue(
        workbook.twoColumnSheets[sheetName].warnings.some(
            warning =>
                warning.warningCode ===
                `INCORRECT_COLUMN_HEADER_${sheetName.toUpperCase()}`
        )
    );
};
```

server/controllers/workbook-constants.js
JustAlex3125: How would I go about doing this and how would this make debugging easier?

ntran18: The current `warningCode` is generic for any sheet. Meaning if the warning is for `degradation_rates` sheet, it would be `INCORRECT_COLUMN_HEADER`, and it would be the same for `production_rates`. In the future, when you want to debug and print out the `warningCode`, you would not know which sheet that`warningCode` is for unless you read the `errorDescription`. For testing, you can check whether the `warningCode` is matched for that specific sheet too.

To make `warningCode` explicit for each sheet, you can use `sheetName` argument in the `warningCode` like this:
```suggestion
                warningCode: `INCORRECT_COLUMN_HEADER_{sheetName.toUpperCase()}`,
```

server/controllers/additional-sheet-parser.js
```diff
@@ -360,56 +371,24 @@ const parseTwoColumnSheet = (sheet, genesInNetwork) => {
...
-                    output.data[currentGene] = undefined;
+            const hasMissingHeader = output.warnings.some(
+                w => w.warningCode === `MISSING_COLUMN_HEADER_${sheet.name.toUpperCase()}`
+            );
+
+            if (hasMissingHeader) {
+                const cellA1 = sheet.data[row][0];
+                const cellB1 = sheet.data[row][1];
+
+                if (cellA1 == null && cellB1 == null) {
```
dondi: IYKYK 😉

```suggestion
                if (cellA1 === null && cellB1 === null) {
```

server/controllers/workbook-constants.js
```diff
@@ -289,6 +289,26 @@ module.exports = {
                 " GRNsight defaults to Saccharomyces cerevisiae.",
         },
 
+        additionalSheetIncorrectColumnHeaderWarning: function (sheetName, expectedA1, expectedB1) {
+            return {
+                warningCode: `INCORRECT_COLUMN_HEADER_${sheetName.toUpperCase()}`,
+                errorDescription:
+                    `GRNsight has detected that the headers are incorrect in the imported workbook's ${sheetName} sheet. ` +
+                    `The headers will need to be corrected to use this workbook as an input file for GRNmap, but will not affect the display of the graph in GRNsight. ` +
+                    `Cell A1 should contain the text ${expectedA1}, and cell B1 should contain the text ${expectedB1}, exactly.`,
```
dondi: NBD but since changes are pending anyway, may as well port this to the recommended `.join(" ")` convention for better readability and formatting control

```suggestion
                errorDescription: [
                    `GRNsight has detected that the headers are incorrect in the imported workbook's ${sheetName} sheet.`,
                    "The headers will need to be corrected to use this workbook as an input file for GRNmap,",
                    "but will not affect the display of the graph in GRNsight.",
                    `Cell A1 should contain the text ${expectedA1}, and cell B1 should contain the text ${expectedB1}, exactly.`
                ].join(" "),
```

(some line lengths may need to be tuned, but you get the idea)

dondi (commented):
Good progress so far! Almost there, just need some `==` adjustments and while we’re at it let’s initiate the transition to `.join(" ")` for long strings

server/controllers/additional-sheet-parser.js
```diff
@@ -179,6 +154,47 @@ const parseMetaDataSheet = sheet => {
...
+    const expectedCellA1 = getSheetHeader(sheetName, 0, 0);
+    const expectedCellB1 = getSheetHeader(sheetName, 1, 0);
+
+    const hasData = sheet.data && sheet.data[0];
+    const cellA1 = hasData ? sheet.data[0][0] : undefined;
+    const cellB1 = hasData && sheet.data[0].length > 1 ? sheet.data[0][1] : undefined;
+
+    const isMissing =
+        cellA1 == null ||
+        cellB1 == null ||
```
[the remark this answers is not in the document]
ntran18: I think the purpose we used `==` instead of `===` was because we want to check whether cellA1 or cellB1 is null or undefined.

So `cellA1 == null` => `cellA1 === null || cellA1 === undefined`

Here is where I read about this example: https://es5.github.io/#x11.9.3, when `x==y`:
```
2. If x is null and y is undefined, return true.
3. If x is undefined and y is null, return true.
```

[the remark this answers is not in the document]
ntran18: I think we can go with the second bullet. cellA1 can go with option 1, but for cell B1, `threshold_b` can have a value of 0, and checking by `!cellB1` will result in false. For consistency, I think we can go with bullet 2.

ntran18:
Regarding the coverage drop, Alex accidentally deleted the tests for the special characters error. However, when I brought back the test, the coverage was still low. Alex added a check for duplicates when adding a new warning. But I can't think of any test for that. In addition, it also shows this line of code is new and isn't covered. [Coveralls]( https://coveralls.io/builds/78025322/source?filename=server%2Fcontrollers%2Fadditional-sheet-parser.js#L64)
image

So I added new tests for wrong data types

On the new [Coveralls](https://coveralls.io/builds/78025669/source?filename=server%2Fcontrollers%2Fworkbook-constants.js#L477) report, we still have one uncovered existing line:
image

For two-column sheets, this error is not applicable anymore. However, other sheets still use this error (e.g. expression sheets)

Issue #1370, opened 2026-02-24
Hardcode optimization_diagnostics sheet into Demo 2


kdahlquist:
In spreadsheet-controller.js I used console logs below line 344

sheet = xlsx.parse(path);
console.log(sheet);

and line 353

var workbook = crossSheetInteractions(sheet);
console.log(workbook);

to investigate what GRNsight is storing when a file is uploaded. 

I found out that the optimization_diagnostics sheet is called "meta2" in the code.  I would like to hardcode this worksheet into demo2, but I can't quite tell from the console log how it should be formatted.

<img width="466" height="148" alt="Image" src="https://github.com/user-attachments/assets/9a122948-842a-48bb-b725-ca261f701834" />

kdahlquist:
Code is now in beta v7.5.8.  Since this is entirely "under the hood", I'm closing this issue.

PR #1376, opened 2026-02-26
Add optimization_diagnostics sheet to demo 2 as per #1370

kdahlquist:
The data is there and my local version of GRNsight seems to function properly after my changes.  However, I don't have any good way to verify that I did things correctly.

While it is working locally, the build failed, and I'm not sure what to do about it.  I ran prettier, so I think the formatting is correct.


dondi (approved):
LGTM, and it looks like the build issues were just transient Coveralls hiccups

Issue #1363, opened 2026-02-19
Import and export of "network_weights" sheet


kdahlquist:
I'm breaking this issue off from #1275 .  There are several issue that all have to do with the "network_weights" sheet and should be considered together, but separate from that issue.

- [x] The network_weights sheet data do not appear to be stored in the `demo-workbook.js` for demos 1-4.  We need to figure out the format for storing the data and then copy it over there.  Looking in the `spreadsheet-controller.js` file for `sheet = xlsx.parse(path);` and ` var workbook = crossSheetInteractions(sheet);` will help with figuring out the format.  @kdahlquist volunteers to make the actual changes to the hardcoding since I am familiar with the data.
- [x] When exporting a demo file (1-4), the network_weights sheet is not clickable.  The simplest explanation is that because the network_weights sheet is not present in `demo-workbook.js`, the data are not available,  Hard-coding the sheet fixes this: PR #1366 .  
  - [x] In this case, whatever is stored in the demo should be the data exported for the "network_weights" sheet.
- [x] When exporting a user-uploaded file that does have a "network_weights" sheet, the checkbox becomes available.  However, there is a bug. As I first reported in #1290, the values in the imported "network_weights" sheet are not the ones exported.

Imported file: 

[3-genes_6 edges_complete-GRNmap-output.xlsx](https://github.com/user-attachments/files/25428168/3-genes_6.edges_complete-GRNmap-output.xlsx)

This same file then exported by GRNsight:

[3-genes_6 edges_complete-GRNmap-output_exported.xlsx](https://github.com/user-attachments/files/25428170/3-genes_6.edges_complete-GRNmap-output_exported.xlsx)

As you can see in the first file that I imported, I used values of "5" in the network_weights sheet.  However, in the sheet that was exported, all the values were "1".

The user-uploaded data needs to be stored and then exported.

- [x] If the user-uploaded spreadsheet does not have a "network_weights" sheet or if the user loads a network from the database, they should still be able to export a "network_weights" sheet (the reason is that they need this sheet for running GRNmap).  In this case, a "network_weights" sheet can be created by GRNsight by copying the contents of the "network" sheet exactly.

PR #1385, opened 2026-03-10
#1363 updated network options and fixed overwriting of network weights values

Amelie1253:


web-client-classic/public/js/upload.js
```diff
@@ -366,14 +364,18 @@ export const upload = function () {
         const twoColumnSheets = grnState.workbook.twoColumnSheets
             ? Object.keys(grnState.workbook.twoColumnSheets)
             : [];
-
+        // is this where exports get created?
```
dondi: If this is figured out, we can remove/revise the comment

Issue #1344, opened 2026-02-12
Restrict Graph to Viewport padding issue when restrict graph to viewport


ceciliazaragoza:
When Restrict Graph to Viewport is checked, ensure that all nodes and edges stay within the viewport of the graph even when zooming or dragging the graph.

kdahlquist:
I was playing around with this today.  I think the behavior is working in v0.0.7.  One little tweak.  Move a node outside the viewport to the right and then check "Restrict Graph to Viewport".  There needs to be a little padding between the node and the right edge of the viewport.  See screenshot of YOX1.  The padding is there if you move a graph to the edge, but it's not when the node automatically gets moved by checking the button.

<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/a4e7dfb0-f84e-49d7-a13d-f9f9581e97c1" />

kdahlquist:
Awaiting PR #1584 .

MilkaZek:
This issue seems to be fixed - aside from the one note about padding between nodes that were previously outside the viewport and the edge, but that behavior is present in beta as well.

kdahlquist:
I am renaming this issue because the padding issue is still a bug in React and needs to be addressed, even if it occurs in Classic.

PR #1419, opened 2026-03-26
#1344: Restrict graph to viewport

ceciliazaragoza:
- related to #1344.
- fixed self-referring edges so that arrowheads do not go behind nodes for unweighted graphs.
- when enable restrict graph to viewport and switch between different viewport sizes, nodes and edges clamp to viewport boundaries.
- when restrict graph to viewport not enabled, nodes can drift outside of viewport bounds.


web-client/src/components/ScaleAndScroll.jsx
```diff
@@ -26,7 +26,7 @@ export default function ScaleAndScroll({ getViewportBoundsData }) {
 
   const handleSliderChange = event => {
     const sliderInput = parseFloat(event.target.value);
-    const viewportBoundsData = getViewportBoundsData();
+    const viewportBoundsData = !adaptive ? getViewportBoundsData() : undefined;
```
dondi, line 29: Super minor, for something like this the negation doesn’t really do anything because you’re handling both cases anyway, so you can remove the negation and flip the values:

```suggestion
    const viewportBoundsData = adaptive ? undefined : getViewportBoundsData();
```

web-client/src/helpers/graphHelpers.js
```diff
@@ -214,16 +209,17 @@ export function getEdgeColor(workbook, edge, grayThreshold, maxWeight, colorOpti
...
   if (!colorOptimal) {
-    for (let i = 0; i < allWeights.length; i++) {
+    for (var i = 0; i < allWeights.length; i++) {
       if (allWeights[i] !== 0) {
         allWeights[i] = 1;
       }
     }
   } else {
-    for (let j = 0; j < allWeights.length; j++) {
+    for (var j = 0; j < allWeights.length; j++) {
```
dondi: Let’s revert these to `let`s since that’s the more modern flavor

dondi (approved):
Other comment resolved before the review was done! 😊 As such, lgtm

Issue #1352, opened 2026-02-12
React: port tests from Classic


kdahlquist:
Opening this as a separate issue to remind us to port the tests from Classic to React as new features are coded in React, especially #1347.  AI might be useful for mass conversion of syntax that would then need to be checked.

ceciliazaragoza:
Add open file tests from classic with the different files.

kdahlquist:
@dondi, can you comment on this issue to say what has been completed and what still needs to be done?

dondi:
PR #1352 lists the tests that have been implemented, broken up into separate files based on functionality. Coverage screenshot was provided

PR #1421, opened 2026-03-26
#1352: Test Coverage

ceciliazaragoza:
- related to #1352.
- Ported over API tests from Classic.
- Implemented tests for frontend by triggering DOM events.
- Need to wait to port over file tests until implement open file functionality.
- This is current coverage. It is a bit harder to test Graph.jsx because of the integration with D3.

image


dondi (approved):
LGTM—this is a big milestone actually, something that isn’t viable at all in `classic` 🙌🏽

PR #1424, opened 2026-03-27
Fixes Issue 1273 - Updated edge coloring enabling logic to be attached to presence of network edge weights

MilkaZek:
Removed logic that attached default edge coloring enablement to whether or not expression data was present, and instead attached it to the network sheet type (which connects to if the "network_optimized_weights" tab is present). Also made it so that when edge weights data isn't present and the "Edge" nav/dropdown is disabled, the edge coloring is shown as disabled as well.


dondi (approved):
LGTM, with reported decrease in coverage noted; I will merge anyway since there isn’t a clear way right now to make up for that lost test coverage

PR #1427, opened 2026-04-03
Fix AllianceMine URL in data_fetcher_service.py

kdahlquist:
When setting up the development environment on my laptop, I encountered an error trying to load the network database with main.py.  It turns out that the AllianceMine URL changed since the last time I did this (2/11/26).  I found the new URL and replaced it in data_fetcher_service.py.  In the future, maybe we should make this configurable?

### DIFF
--- a/database/network-database/data_services/data_fetcher_service.py
+++ b/database/network-database/data_services/data_fetcher_service.py
@@ -6,7 +6,7 @@ from io import StringIO
 
 class DataFetcherService(ABC):
     def __init__(self):
-        self.service = Service("https://www.alliancegenome.org/alliancemine/service")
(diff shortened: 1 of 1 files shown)


kdahlquist:
Note that I made the branch `kdahlquist` from `main`, not ` beta` because I forgot about that.  Because this is a simple change, it shouldn't matter.  I'll be more careful in the future.

dondi (approved):
LGTM—having cut the branch from `main` is viable here, because we have recently deployed to production and there haven’t been any changes to `main` since then, so the only tracked change is indeed precisely the change to the URL

In cases where more undesired/unintended changes are involved, typically the course of action will be to “re-cut” the branch from the right base branch. But yes, as observed that was not necessary here

Issue #1347, opened 2026-02-12
Open file and load from database


ceciliazaragoza:
Enable ability to open a file and load from database

kdahlquist:
In v0.0.8 on 4/23/26, there is an issue with PPI networks.  If I load a PPI, the edges are duplicated.  This was a bug in Classic, see issue #1317 .

kdahlquist:
PR #1544 approved and awaiting merge.

MilkaZek:
It looks like I am able to upload and display weighted and unweighted xlsx, sif, and graphml files - and edge coloring looks like it's working. I can't seem to load from the database though.

MilkaZek:
It seems that load from database works locally (once I have GRNsight Classic also running in the background to the backend running) but not on the deployed web app. I've opened a new issue for this: [#1656](https://github.com/dondi/GRNsight/issues/1656)

kdahlquist:
Agreed that .xlsx, .sif, .graphml open file is working.  Since there is a new issue about the load from database, I'm closing this one.

PR #1544, opened 2026-04-23
#1347: Load from Database

ceciliazaragoza:
- related to issue #1347.
- Can load from database.
  - Deviate slightly from classic because instead of doing an alert when an invalid gene is input, display a red error box in modal.
- Align styling of demo selection in navbar and sidebar to align more with classic.
- Navbar menus close when anything is clicked within the navbar menu.


web-client/src/components/LoadFromDbModal.jsx
```diff
@@ -0,0 +1,487 @@
...
+  const [searchValue, setSearchValue] = useState("");
+  const [errorMessage, setErrorMessage] = useState("");
+  const [customWorkbook, setCustomWorkbook] = useState(EMPTY_WORKBOOK);
+
+  const sourceOptions = useMemo(() => {
+    if (customWorkbook.type === NETWORK_PPI_MODE_SHORT) {
+      return Object.keys(customWorkbook.sources.proteinProteinInteractions || {});
+    }
+
+    return Object.keys(customWorkbook.sources.geneRegulation || {});
```
dondi: If the left-hand operands are expected to be objects or arrays, the latest best practice is to use null-coalescing vs. or:

```suggestion
      return Object.keys(customWorkbook.sources.proteinProteinInteractions ?? {});
    }

    return Object.keys(customWorkbook.sources.geneRegulation ?? {});
```

web-client/src/components/LoadFromDbModal.jsx
```diff
@@ -0,0 +1,487 @@
...
+          queryProteinProteinDatabase({ type: "NetworkSource" }),
+          queryNetworkDatabase({ type: "NetworkSource" }),
+        ]);
+
+        if (isCancelled) {
+          return;
+        }
+
+        const proteinSources = proteinResponse?.sources || {};
+        const geneSources = networkResponse?.sources || {};
```
dondi: Ditto on `||` vs. `??`

web-client/src/components/LoadFromDbModal.jsx
```diff
@@ -0,0 +1,487 @@
...
+  };
+
+  const addGene = async () => {
+    const userEntry = String(searchValue || "").trim();
+    const normalized = isValidGene(userEntry.toUpperCase());
+
+    if (!normalized) {
+      setErrorMessage(
+        `${geneProteinLabel}: "${userEntry}" is not to GRNsight specifications. ` +
+          `${geneProteinLabel}s must be 12 characters or less, containing "-", "_", and alpha-numeric characters only.`
```
dondi: Let’s use the `[].join(' ')` technique for long strings for better readability (and automatic space separation!)

```suggestion
        [
          `${geneProteinLabel}: "${userEntry}" is not to GRNsight specifications.`,
          `${geneProteinLabel}s must be 12 characters or less, containing "-", "_",`,
          'and alpha-numeric characters only.'
        ].(' ')
```

web-client/src/components/LoadFromDbModal.jsx
```diff
@@ -0,0 +1,487 @@
...
+      setErrorMessage("Network must have at least 1 gene or protein.");
+      return;
+    }
+
+    if (genesAmount > 75) {
+      setErrorMessage(
+        `GRNsight can handle at most 75 genes/proteins. This network contains ${genesAmount}.`
+      );
+      return;
+    }
```
dondi: The `75`s here represent the same value, so will benefit from a constant (not sure if this shows anywhere else, so worth a search)


```suggestion
    if (genesAmount > MAX_GENES) {
      setErrorMessage(
        `GRNsight can handle at most ${MAX_GENES} genes/proteins. This network contains ${genesAmount}.`
      );
      return;
    }
```

web-client/src/components/LoadFromDbModal.jsx
```diff
@@ -0,0 +1,487 @@
...
+          timestamp: sourceData?.timestamp,
+        });
+
+        const edgesAmount = countEdges(response?.links);
+        if (edgesAmount > 100) {
+          setErrorMessage(
+            `GRNsight can handle at most 100 edges. This network contains ${edgesAmount}.`
+          );
+          return;
+        }
```
dondi: Ditto `100`

web-client/src/components/LoadFromDbModal.jsx
```diff
@@ -0,0 +1,487 @@
...
+
+        const links = [];
+        Object.entries(response?.links || {}).forEach(([regulator, targets]) => {
+          (targets || []).forEach(target => {
+            if (customWorkbook.genes[regulator] && customWorkbook.genes[target]) {
+              links.push(`${customWorkbook.genes[regulator]}->${customWorkbook.genes[target]}`);
+            }
+          });
+        });
```
dondi, line 249: I realize that this is ported over, but now that we are doing so, this is a `??` opportunity

```suggestion
        const links = [];
        Object.entries(response?.links ?? {}).forEach(([regulator, targets]) => {
          (targets ?? []).forEach(target => {
            if (customWorkbook.genes[regulator] && customWorkbook.genes[target]) {
              links.push(`${customWorkbook.genes[regulator]}->${customWorkbook.genes[target]}`);
            }
          });
        });
```

web-client/src/components/Navbar.jsx
```diff
@@ -447,21 +501,28 @@ export default function Navbar({}) {
 
             <DottedLine />
             <Box pad={{ horizontal: "20px", vertical: "3px" }} direction="row">
-              <Text>
+              <Text color={isZoomControlDisabled ? "#ccc" : undefined}>
```
dondi: Does Grommet have a theme color for disabled? Or maybe the `Text` component itself can just be labeled as `disabled`? Just seeing if we can abstract out the hardcoded color

web-client/src/helpers/loadFromDbHelpers.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export function countEdges(links) {
+  return Object.values(links || {}).reduce((total, targets) => {
+    return total + (Array.isArray(targets) ? targets.length : 0);
+  }, 0);
+}
+
+export function sortWorkbookGenes(genes, workbookType) {
+  const entries = Object.entries(genes || {});
```
dondi: Looks like file has more `||` --> `??` opportunities

web-client/src/App.css
```diff
@@ -100,27 +115,21 @@ header span {
...
+  align-items: center;
+  justify-content: flex-start !important;
+  text-align: left;
 }
 
-button {
-  border: none;
-  box-shadow: none;
+.demo-dropdown-navbar button:hover {
+  border: none !important;
```
dondi: Quite a few `!important`s here—can we track down what necessitates these? If we can minimize them with more precise CSS selectors then that would be the ideal; if there are any cases with no other solution, let’s leave a comment explaining what’s getting in the way

web-client/src/App.css
```diff
@@ -296,8 +309,178 @@ input[type="radio"]:checked {
...
+  margin-top: 8px;
+}
+
+.generate-network-modal-footer {
+  border-top: 1px solid #d2d2d2;
+  padding: 20px 20px 20px 20px;
+}
+
+.generate-network-action-button {
+  padding: 6px 12px !important;
```
dondi: Wow, quite a few more…might these be holdovers from classic?

dondi (commented):
Overall structure looks good—all just little style/convention things. Some of these might have been holdovers from classic of course…like some past tweaks, this happens to be a good opportunity to update them

web-client/src/components/LoadFromDbModal.jsx
```diff
@@ -226,16 +231,17 @@ export default function LoadFromDbModal({ margin }) {
...
-            `GRNsight can handle at most 100 edges. This network contains ${edgesAmount}.`
+            `GRNsight can handle at most ${MAX_EDGES} edges. This network contains ${edgesAmount}.`
           );
           return;
         }
 
         const links = [];
-        Object.entries(response?.links || {}).forEach(([regulator, targets]) => {
-          (targets || []).forEach(target => {
+        console.log("Response links:", response?.links);
```
dondi, line 242: Not a blocker, but it looks like we have a `console.log` straggler 😅

web-client/src/helpers/theme.js
```diff
@@ -95,6 +96,17 @@ export const theme = {
       radius: "4px",
     },
     extend: `
+      /* Keep navbar buttons borderless without using !important */
+      .navbar &,
+      .navbar &:hover,
+      .navbar &:focus,
+      .navbar &:focus-visible,
+      .navbar &:active {
```
dondi, line 104: Lengthy, but more explicit and easier to debug later on when styles are being traced 💯

web-client/src/App.css
```diff
@@ -94,13 +94,31 @@ header span {
...
+
+.dropdown-menu button:hover,
+.dropdown-menu button:focus,
+.dropdown-menu [role="button"]:hover,
+.dropdown-menu [role="button"]:focus {
+  border: none;
+  background-color: transparent;
+  box-shadow: none;
+}
+
```
dondi, line 114: Worth it!

dondi (approved):
LGTM except for the lingering `console.log` but will wait for #1546 to get merged into here before merging into `beta`

Issue #1348, opened 2026-02-12
Add links to Help menu in GRNsight React


kdahlquist:
As part of scoping the work #1334, I think you should also add to your list finishing the Help menu.  You just need to add hyperlinks to the appropriate pages: Getting Started (goes to Documentation page), GRNsight wiki (goes to wiki home page in repo), About GRNsight (goes to GRNsight About page).

kdahlquist:
This issue has a pending PR #1546.

MilkaZek:
This issue seems to be completed as well! Is it ready to be closed?

kdahlquist:
This is not yet working in the deployed version v0.0.10 (05/11/2026), so I don't know where the discrepancy is.

dondi:
Needs manual deployment to `gh-pages` as of now, and when the automatic deployment PR has been merged, this should then happen with every push

Reference: https://vite.dev/guide/static-deploy

PR #1546, opened 2026-04-23
#1348: Help menu links

ceciliazaragoza:
- Related to #1348.
- Implemented Help menu links in navbar.


web-client/src/components/Navbar.jsx
```diff
@@ -94,6 +94,11 @@ export default function Navbar({}) {
     setOpenMenu(null);
   };
 
+  const openExternal = url => () => {
+    window.open(url, "_blank", "noopener,noreferrer");
```
dondi: Consider rendering these as genuine links instead of buttons that invoke `window.open` because `window.open` is sometimes flagged by browsers as popup behavior. See below…

web-client/src/components/Navbar.jsx
```diff
@@ -508,9 +513,20 @@ export default function Navbar({}) {
...
           <div className="dropdown-menu" onClickCapture={handleDropContentClick}>
-            <DropdownMenuButton text="Getting Started" />
-            <DropdownMenuButton text="GRNsight Wiki" />
-            <DropdownMenuButton text="About GRNsight" />
+            <DropdownMenuButton
+              text="Getting Started"
+              onClick={openExternal(
+                "https://dondi.github.io/GRNsight/documentation.html#gettingStarted"
+              )}
+            />
```
dondi, line 549: Depending on the footprint of `DropdownMenuButton`, you can either:
* Change the rendering of `DropdownMenuButton` so that it renders a link (`a`) instead. `_blank`, `noopener`, and `noreferrer` can be assigned as attributes to the `a`
* If `DropdownMenuButton` is used in other places that genuinely need a click, then you can implement a component link `DropdownMenuLink` that explicitly renders an `a`

dondi: The `a` tag would look something like this:
```html
            <a
              href="......"
              target="_blank"
              rel="noopener noreferrer"
            >
```

dondi (commented):
Overall looks good but we should try to avoid `window.open` when we can

dondi (commented):
The actual changes relating to the issue LGTM but it appears there’s some interaction with the `!important` work from the parent branch? Let’s reconcile these so that the PR merge is clean (and then when this is merged, we can merge the parent)

ceciliazaragoza:
Merge conflicts are resolved and are ready for review.

dondi (approved):
All cleaned up! LGTM

Issue #1383, opened 2026-03-10
Zoom slider doesn't reset to 100% when switch between demos


ceciliazaragoza:
When switch between demos, the zoom percentage resets to 100%, but the zoom slider does not reset back to 100% and remains at the previous zoom slider position.

Zoom value in navbar also remains at 100 despite the zoom percentage changing.

kdahlquist:
I tested this with Firefox on React v0.0.10 (05/11/2026), and the zoom slider in the d-pad is still stuck on the previous value, even though the new demo file is at 100%.

kdahlquist:
This needs more work, lowering priority level.

PR #1547, opened 2026-04-23
#1383: Zoom slider resets to 100% when switch between demos

ceciliazaragoza:
- related to #1383 
- zoom slider resets to 100% when switch between demos.
- zoom text input in navbar matches the zoom input slider, and the slider matches the zoom text input.


dondi (approved):
LGTM

Issue #1418, opened 2026-03-24
Unchecking all but the sigmas in Demo 2 causes Expression Data Sources to select None


dondi:
As of now, individually deselecting all but the sigma sheets in Demo 2 will set the Expression Data Sources radio button to None, which in turn them makes all of the expression sheets disappear

We need to think through how these controls should behave together, as well as how it relates to the Select All checkbox. Further specs will be supplied when ready

kdahlquist:
Upon reflection, I think the simplest way to think about it is to treat the `*_sigmas` sheets as expression sheets with regard to the checkbox behavior in the modal.

Note that this behavior would need to be implemented for Demo 2 and also any imported workbook that had `*_sigmas` worksheets in it.

- You need to uncheck _all_ of the `*_log2_expression`, `*_log2_optimized_expression`, and `*_sigmas` sheets before the Expression Data source switches to `None` and the checkboxes disappear.

I think that seeing the _checked_ `*_sigmas` disappear is what triggered us thinking that this was a bug, so if we require them to be unchecked before they disappear, it will be more in alignment with the behavior that we expect.

- The corollary to this is that if the `Select All` option is unchecked, then the Expression Sheets options should disappear and the Expression Data Source should change to `None`.

I'll repeat this comment in #1387 so it's clear in both issues what the expected behavior is.

kdahlquist:
We decided in the meeting that @jiacelyne will tackle this issue solo this week.

jiacelyne:
The bug came from expression value being the only one in the if statement below.

`if (
                        typeof allSheets[i] === "object" &&
                        allSheets[i].id !== "exportExcelWorkbookSheet-All" &&
                        allSheets[i].value &&
                        allSheets[i].value.includes("expression") &&
                        allSheets[i].checked
                    )`

to 

`if (
                    typeof allSheets[j] === "object" &&
                    allSheets[j].id !== "exportExcelWorkbookSheet-All" &&
                    allSheets[j].checked &&
                    allSheets[j].value &&
                    (allSheets[j].value.includes("expression") ||
                        allSheets[j].value.includes("sigma"))
                )`

<img width="428" height="682" alt="Image" src="https://github.com/user-attachments/assets/4f62636b-2315-4085-9bb2-a6ea0e774c08" />

[...]

PR #1551, opened 2026-04-24
 #1418 Unchecking all but the sigmas in Demo 2 causes Expression Data Sources to select None

jiacelyne:


dondi (commented):
Looks like you derived your branch from `main`—this results in conflicts when I try to merge it into `beta`. Please “replay” your change on a branch that is created from `beta` so that the PR can merge there instead; thanks!

Issue #1339, opened 2026-02-11
For the network database, separate the fetching/processing steps from the loading into PostgreSQL steps


kdahlquist:
As part of issue #1229 and #1302, I am following the instructions to load the GRNsight databases on my new research computer [found here](https://github.com/dondi/GRNsight/blob/main/database/README.md).

The main.py script does everything--it fetches data from AllianceMine, processes it, and loads it into the network database.  In keeping with best practices for data integrity, we keep all prior versions of the network database that users can access using the load from database feature on GRNsight.  In keeping with this, we need to do a few things:
1. We need to keep back-ups of all the released versions of the network database on Box.
  - I know we did a major refactor of our database recently so older versions would not work with the current code (I think), but we need to try to back them up anyway.
2. We need to separate the functions of fetching and processing new data from AllianceMine from populating the network database in postgres.
  - This is because within our group, we all need to be doing development and testing on the same version of the network database, as we do for the expression database.  For example, a bug might be introduced with new data and would be hard to track down if everyone is using a different version of the database.
  - People outside of our group who want to build GRNsight should have access to our production versions of the database for their own use, and would only need to populate their database from our backed-up files.
  - Databases are a moving target, and even a day's difference in downloading data could result in different versions of the data.

kdahlquist:
Not quite this issue, but I've created a PR #1427 because the URL to get data from AllianceMine changed.  Is this something that could be configurable?

kdahlquist:
Created new issue and then closed it #1434 because PR #1427 does not directly address this issue.

ntran18:
I uploaded a folder in our [Box](https://lmu.app.box.com/folder/378840928315) that contains all the production network data (including YeastMine and AllianceMine). The new script I just wrote would be able to load data into Postgres from:
1. Script-results folder (reserved for data downloaded from AllianceMine API)
2. Custom folder (data downloaded from Postgres, refer to [Backup database folder in Box](https://lmu.app.box.com/folder/378840928315) for the format

kdahlquist:
I changed the folder name in Box to "Network Database" to distinguish it from the "Expression Database" folder (which I also changed from "GRNsight Expression").  I made the corresponding changes to the documentation in the READMEs on the beta branch.  The hyperlinks remain the same.

kdahlquist:
I'm trying to check this.  I started out running `main.py` with the following command which had worked for me previously:

[...]

PR #1553, opened 2026-04-27
#1339 Adding script to allow populate data by a specific folder

ntran18:
This script allows users to populate data that was downloaded directly from Postgres for #1339


ntran18:
Hmmm I'm not sure why coverage dropped .... I only touched `database` folder, which is not related to `server` folder

dondi (approved):
LGTM…yes, not sure how coverage would have dropped too 🤷🏻‍♂️

Issue #1456, opened 2026-04-20
Create a warning for when missing all the expression sheets warning


JustAlex3125:
Need to create a warning message that warns the user when they are importing/exporting a GRNmap file that has no expression sheet. This relates to expression sheet file called "a_3-genes_6 edges_missing-all-expression-sheets" mentioned in #1450 files"

JustAlex3125:
@kdahlquist Here is what I think the warnings should be for importing/exporting for this issue. Let me know if there are any alterations that I should make to the warnings. I also went in and edited the [Onboarding CheckList](https://github.com/dondi/GRNsight/wiki/Onboarding-Checklist) to add the section about the project board collaborators. Let me know if there are any changes you want me to make on that. 

Import -> Warning
- GRNsight has detected that the expression sheets are missing in the imported workbook. This will affect the GRNsight graph but will not affect the functionality of GRNsight. Please reimport a workbook that contains the proper expression sheets so GRNsight will function fully.

Export -> Do nothing

kdahlquist:
For this, let's keep in mind that the only thing that GRNsight needs to load a graph is either a `network` or `network_optimized_weights` sheet.  All other worksheets are optional.  So, if the `*_log2_expression` sheet is missing, the only issue for the user is that they would not be doing any node coloring, which would not prevent the display of the graph.

So in this case, upon import the action is "do nothing" and for export the action is "do nothing".  Note that for the export, they have the option to select from among the expression data supplied by the expression database.

This test file was created to make sure that GRNsight was functioning properly in the absence of the `*_log2_expression` sheet,  We should still upload this workbook to the repo with a note that a graph should load without node coloring in this case.

My recommendation is to look at [this comment](https://github.com/dondi/GRNsight/issues/1192#issuecomment-3880061708) on issue 1192 where the expected behavior for all the cases is described.  There will be many parallels with the expression sheets.

kdahlquist:
Also, I saw the note you added to the Initial Setup wiki :)

JustAlex3125:
That makes sense. I'll add this file to the repo and move to taking on another issue. Is there a certain way you want me to go about adding it?

kdahlquist:
I'm not sure. You will need to ask @ntran18 what to do. I don't know how all the test files are organized.

JustAlex3125:
Created a pull request for this issue. All I did was add a file that when imported should display a graph with no node coloring.

kdahlquist:
[...]

PR #1554, opened 2026-04-27
Added file for issue #1456

JustAlex3125:
Added a file that is a control file. When this file is imported into GRNsight it should display a graph with no node coloring. This is in reference to issue #1456.


dondi (approved):
LGTM

PR #1556, opened 2026-04-28
#1418 changing code within checkbox export popup to support sigma checkboxes

jiacelyne:
now comparing against beta


dondi (approved):
LGTM—thanks for the quick redo!

Issue #1454, opened 2026-04-16
Automate deployment of GRNsight React to beta


ceciliazaragoza:
- When a user pushes `web-client` code to beta, then there is automatic deployment to gh-pages

dondi:
@ceciliazaragoza and @dondi looked at the GitHub Actions settings for the deployment step and encountered a few aspects that may bear investigation, especially in relation to how Coveralls integrates with our repository. There appear to be multiple ways to configure the deployment integration, including `ssh-key` and `token`. We also learned of the built-in `GITHUB_TOKEN` secret (https://docs.github.com/en/actions/concepts/security/github_token) although this isn’t documented by the deployment action

Recommendation is for @ceciliazaragoza to experiment with the various options and we can settle on the preferred one after we know more about how/whether each option works

kdahlquist:
This has a PR #1563.

PR #1563, opened 2026-04-29
#1454: Automatic Deployment of React App

ceciliazaragoza:
- related to #1454.


ceciliazaragoza:
- This is ready for review. 
- Instead of using someone's marketplace app, we deploy the app with GitHub Actions. We can remove SSH key that created for other strategy since we just use `secrets.GITHUB_TOKEN`. 
- I also updated the package.json version access so that it is not `undefined` when we do the automated build with GitHub Actions.

ceciliazaragoza:
This is ready for review again. I made changes to ensure that the file structure of gh-pages is not disrupted and changes are only made to the `react-thesis-4081` folder. Also, we ensure that if multiple branches push at the same time, we add the jobs to a queue with `concurrency: gh-pages-deploy`.

This page and deployments build shows a successful automatic deployment, and I confirmed that Home and Beta are still up and running: https://github.com/dondi/GRNsight/actions/runs/25690035571

Issue #1557, opened 2026-04-28
Special handling for warnings and Excel export of optimized_production_rates and optimized_threshold_b test files b, c, d


kdahlquist:
For these test files:
- [x] b_3-genes_6 edges_optimized_threshold_b_sheet_blank.xlsx
- [x] b_3-genes_6 edges_optimized-prod-rates-sheet-blank.xlsx

We need special handling of the warnings upon Export to Excel (noting, of course if the sheet in question is not selected for export we do nothing).

Currently, in beta v7.7.6 that incorporated changes from #1422, when you import those files and then Export to Excel, selecting the `optimized_threshold_b` sheet, the warning says: _The "optimized_threshold_b" sheet was empty in the exported workbook. GRNsight has supplied the default value of 0 for each gene._

However, in this case, we do not want to fill in the missing values with "0".  The `optimized_threshold_b` sheet is GRNmap output and we should not be placing values in it.  Instead the warning should simply say:   The "optimized_threshold_b" sheet was empty in the exported workbook. GRNsight is checking because a threshold b value should have been provided as GRNmap output, but will not affect the display of the graph in GRNsight.

Same thing for the `optimized_production_rates`.  Currently it says: _The "optimized_production_rates" sheet was empty in the exported workbook. GRNsight has supplied production rates from the backend database which are 2X the degradation rates reported in Neymotin et al. (2014)._

However, we should not be filling in the missing values.  The warning should just say: _The "optimized_production_rates" sheet was empty in the exported workbook. GRNsight is checking because a production rate should have been provided as GRNmap output, but will not affect the display of the graph in GRNsight._

- [x] c_3-genes_6 edges_missing-all-optimized-prod-rate-genes-and-values.xlsx
- [x] c_3-genes_6 edges_missing-all-optimized-threshold_b_genes-and-values.xlsx
- [x] d_3-genes_6 edges_missing-all-optimized-prod-rate-values.xlsx
- [x] d_3-genes_6 edges_missing-all-optimized-threshold_b-values.xlsx

For these test files, the import warning is correct, but the export warnings need to be changed to the ones described above.

JustAlex3125:
I have corrected the warnings for this and created a [pull request](https://github.com/dondi/GRNsight/pull/1565). @kdahlquist Let me know if there are any other changes on this issue.

kdahlquist:
In beta v7.7.7, these two workbooks have correct warnings:
[...]

PR #1565, opened 2026-05-02
Adjusted the warnings as described in #1557

JustAlex3125:
Corrected the warnings that had to do with exporting optimized expression sheets and threshold b sheets. Also corrected a couple of minor spelling errors.


dondi (approved):
LGTM

Issue #1558, opened 2026-04-28
Special handling for optimized_production_rates and optimized_threshold_b test files n, o, p, q


kdahlquist:
For these test files:

- [x] n_3-genes_6 edges_wrong-prod-rate-header-opt-prod-rates-sheet.xlsx
- [x] n_3-genes_6 edges_wrong-threshold_b-header-opt-threshold_b-sheet.xlsx

- [x] o_3-genes_6 edges_wrong-geneID-header-opt-prod-rates-sheet.xlsx
- [x] o_3-genes_6 edges_wrong-geneID-header-opt-threshold_b-sheet.xlsx

- [x] p_3-genes_6 edges_missing-headers-opt-prod-rates-sheet.xlsx
- [x] p_3-genes_6 edges_missing-headers-opt-threshold_b-sheet.xlsx

- [x] q_3-genes_6 edges_blank-headers-opt-prod-rates-sheet.xlsx
- [x] q_3-genes_6 edges_blank-headers-opt-threshold_b-sheet.xlsx

Current warning upon import: 
- GRNsight has detected that the headers are incorrect in the workbook's **optimized_production_rates** sheet. The headers will need to be corrected to use this workbook as an input file for GRNmap, but will not affect the display of the graph in GRNsight. Cell A1 should contain the text **id**, and cell B1 should contain the text **production_rate**, exactly.
- GRNsight has detected that the headers are incorrect in the workbook's **optimized_threshold_b** sheet. The headers will need to be corrected to use this workbook as an input file for GRNmap, but will not affect the display of the graph in GRNsight. Cell A1 should contain the text **id**, and cell B1 should contain the text **threshold_b**, exactly.

These need to be changed to:
- GRNsight has detected that the headers are incorrect in the workbook's **optimized_production_rates** sheet. Cell A1 should contain the text **id**, and cell B1 should contain the text **production_rate**, exactly. GRNsight is checking because these headers should have been provided by the GRNmap output, but they will not affect the display of the graph in GRNsight. 
- GRNsight has detected that the headers are incorrect in the workbook's **optimized_threshold_b** sheet. Cell A1 should contain the text **id**, and cell B1 should contain the text **threshold_b**, exactly. GRNsight is checking because these headers should have been provided by the GRNmap output, but they will not affect the display of the graph in GRNsight. 

No warnings are given upon export, which is correct because GRNsight has internally corrected the headers before storing the data.

JustAlex3125:
[...]

PR #1566, opened 2026-05-03
Adjusted the warnings as described in #1558

JustAlex3125:
Corrected the warnings that had to do with exporting optimized expression sheets and threshold b sheets in issue #1558


dondi (approved):
LGTM

Issue #1559, opened 2026-04-28
Special handling for optimized_production_rates and optimized_threshold_b test files r


kdahlquist:
For these test files:
- [x] r_3-genes_6 edges_extra-row-opt-prod-rates-sheet.xlsx
- [x] r_3-genes_6 edges_extra-row-opt-threshold_b-sheet.xlsx

Current import warning: 
- GRNsight has detected that there are extra genes in the imported workbook's 'optimized_production_rates' sheet. **The genes in the 'optimized_production_rates' sheet need to match the genes in the 'network' sheet to use this workbook` as an input file for GRNmap**, but will not affect the display of the graph in GRNsight. The extra genes are: ZAP1.
- GRNsight has detected that there are extra genes in the imported workbook's 'optimized_threshold_b' sheet. GRNsight is checking because the genes in the 'optimized_threshold_b' sheet should match the genes in the 'network' sheet when outputted from GRNmap, but will not affect the display of the graph in GRNsight. The extra genes are: ZAP1.

These need to be changed to:
- GRNsight has detected that there are extra genes in the imported workbook's 'optimized_production_rates' sheet. **GRNsight is checking because the genes in the 'optimized_production_rates' sheet should match the genes in the 'network' sheet when outputted from GRNmap**, but will not affect the display of the graph in GRNsight.  The extra genes are: ZAP1.
- GRNsight has detected that there are extra genes in the imported workbook's 'optimized_threshold_b' sheet. **GRNsight is checking because the genes in the 'optimized_threshold_b' sheet should match the genes in the 'network' sheet when outputted from GRNmap**, but will not affect the display of the graph in GRNsight. The extra genes are: ZAP1.

If these worksheets are not selected for export, the current version correctly does nothing.  However if they are exported, the current warnings are:
- GRNsight has detected that there are extra genes in the exported workbook's 'optimized_production_rates' sheet. **GRNsight is checking because the genes in the 'optimized_production_rates' sheet should match the genes in the 'network' sheet when outputted from GRNmap,** but will not affect the display of the graph in GRNsight. The extra genes are: ZAP1.
- GRNsight has detected that there are extra genes in the imported workbook's 'optimized_threshold_b' sheet. **GRNsight is checking because the genes in the 'optimized_threshold_b' sheet should match the genes in the 'network' sheet when outputted from GRNmap**, but will not affect the display of the graph in GRNsight. The extra genes are: ZAP1.

JustAlex3125:
[...]

PR #1568, opened 2026-05-04
Updated the warning message described in #1559

JustAlex3125:
Updated the warning message that had to do with extra rows in order to properly account for extra rows present in the optimized sheets. This should address issue #1559.


dondi (approved):
LGTM

PR #1584, opened 2026-05-14
#1344: Whitespace between node and right boundary

ceciliazaragoza:
When the nodes are exceeding bounds and then restrict graph to viewport is enabled, there is whitespace between the node and the graph right boundary.


ceciliazaragoza:
Need to fix so that only most recent 2 commits are on branch. It should only be a two-line fix.

dondi (approved):
LGTM

PR #1599, opened 2026-06-15
Updated warning message to account for issue #1557

JustAlex3125:
Whenever a user attempts to export a file that is missing any data from any of the two column sheets, instead of being told that anything is being entered now, they are told that data is missing. This is a more accurate warning than what was being displayed before.

### DIFF
--- a/web-client-classic/public/js/export-warning-constants.js
+++ b/web-client-classic/public/js/export-warning-constants.js
@@ -55,7 +55,7 @@ module.exports = {
                 warningCode: `MISSING_OR_EMPTY_${sheetName.toUpperCase()}_SHEET`,
                 errorDescription: [
                     isMissing
-                        ? `There was no "${sheetName}" sheet in the imported workbook.`
+                        ? `There was no "${sheetName}" sheet in the exported workbook.`
                         : `The "${sheetName}" sheet was empty in the exported workbook.`,
                     prefix
                         ? `GRNsight is checking because a ${sheetName.split("_")[1] + " " + sheetName.split("_")[2]} value should have been provided as GRNmap output, but will not affect the display of the graph in GRNsight.`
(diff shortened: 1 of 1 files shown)


web-client-classic/public/js/export-warning-constants.js
dondi, line 68: The `split` function is being called twice as written below, so first we should ensure that calculation is performed just once:

```suggestion
            const missingType = isAllGenesMissing ? "genes and values" : "values";
            const sheetNameSegments = sheetName.split("_");
```

web-client-classic/public/js/export-warning-constants.js
```diff
@@ -71,7 +71,7 @@ module.exports = {
                 errorDescription: [
                     `There were no ${missingType} supplied`,
                     `in the "${sheetName}" sheet in the exported workbook.`,
-                    `GRNsight has supplied ${dataSourceForTwoColumnSheet[sheetName]}.`,
+                    `GRNsight is checking because a ${sheetName.split("_")[0] + " " + sheetName.split("_")[1]} value should have been provided as GRNmap output, but will not affect the display of the graph in GRNsight.`,
```
dondi: Then here, the + concatenation is redundant within the template string and the template string itself is longer than it needs to be. I’d change as follows, to address all of these points:

```suggestion
                    'GRNsight is checking',
                    `because a ${sheetNameSegments[0]} ${sheetNameSegments[1]} value`,
                    'should have been provided as GRNmap output,',
                    'but will not affect the display of the graph in GRNsight.',
```

There are other ways to break things up of course, but I tried to make it so that each line ends at a reasonable pause

dondi (commented):
Overall content looks good, but we need to eliminate the duplicate `split` call and refine the way we segment the message string

JustAlex3125:
@dondi I have implemented the changes you have suggested as well as fixed the warnings to only display if the sheets are optimized.

dondi (approved):
Thanks for the refactors! LGTM

Issue #1561, opened 2026-04-28
Special handling for import/export warnings for optimized_production_rates and optimized_threshold_b test files t


kdahlquist:
For the test files:
- [x] t_3-genes_6 edges_wrong-geneID-opt-prod-rates-sheet.xlsx
- [x] t_3-genes_6 edges_wrong-geneID-opt-threshold_b-sheet.xlsx

Current import warning gives two warnings:
1. GRNsight has detected that there are missing genes and production rates in the imported workbook's optimized_production_rates sheet. GRNsight is checking because a production rate should have been provided as GRNmap output, but will not affect the display of the graph in GRNsight. The missing genes and values are: CIN5.
2. GRNsight has detected that there are extra genes in the imported workbook's 'optimized_production_rates' sheet. The genes in the 'optimized_production_rates' sheet need to match the genes in the 'network' sheet to use this workbook` as an input file for GRNmap, but will not affect the display of the graph in GRNsight. The extra genes are: ACT1.

The top warning message is correct, but the bottom warning needs to be reworded to:
2. GRNsight has detected that there are extra genes in the imported workbook's 'optimized_production_rates' sheet. The genes in the 'optimized_production_rates' sheet should match the genes in the 'network' sheet from the GRNmap output, but will not affect the display of the graph in GRNsight. The extra genes are: ACT1.

Current export warning:
- GRNsight has detected that there are extra genes in the exported workbook's 'optimized_production_rates' sheet. The genes in the 'optimized_production_rates' sheet need to match the genes in the 'network' sheet to use this workbook` as an input file for GRNmap, but will not affect the display of the graph in GRNsight. The extra genes are: ACT1.

Change this to: 
- GRNsight has detected that there are extra genes in the exported workbook's 'optimized_production_rates' sheet. The genes in the 'optimized_production_rates' sheet should match the genes in the 'network' sheet from the GRNmap output, but will not affect the display of the graph in GRNsight. The extra genes are: ACT1.

JustAlex3125:
Created a [pull request](https://github.com/dondi/GRNsight/pull/1570) that updated the warnings for optimized expression sheets. @kdahlquist Let me know if there are any changes you want me to make to the warnings.

dondi:
I left a review on this PR last night: https://github.com/dondi/GRNsight/pull/1570#pullrequestreview-4226144697

[...]

PR #1600, opened 2026-06-15
Updated two column warning to fix issue #1561

JustAlex3125:
The import and export warnings have both been updated to fix the wording and tone.


dondi (approved):
LGTM

PR #1601, opened 2026-06-22
Updated warning to deal with optimized sheets

JustAlex3125:
Fixed the warning for issue #1561 to change the warning message whenever an optimized sheet is input to GRNsight.


dondi (approved):
LGTM, with one tweak that I went ahead and made: I suspect that there was an erroneous comma in the second message, which would have read:
> …need to match genes in the 'network' sheet to use this workbook, as an input file for GRNmap, …

I think that comma in the middle shouldn’t actually be there, so that this clause more clearly states that the match is specifically needed by GRNmap but not GRNsight:
> …need to match genes in the 'network' sheet to use this workbook as an input file for GRNmap, …

Hope I read this correctly…we’ll see upon review…

PR #1604, opened 2026-06-30
Fixed the warning that had to do with issue #1557

JustAlex3125:
The warning messages have been updated to be the correct warning for this issue and the code has been fixed to properly display the warning when the sheet is blank.

### DIFF
--- a/web-client-classic/public/js/export-warning-constants.js
+++ b/web-client-classic/public/js/export-warning-constants.js
@@ -50,18 +50,21 @@ module.exports = {
 
         MISSING_OR_EMPTY_TWO_COLUMN_SHEET: function (sheetName, isMissing) {
             const prefix = sheetName.includes("optimized");
+            const nameSegment = sheetName.split("_");
 
             return {
                 warningCode: `MISSING_OR_EMPTY_${sheetName.toUpperCase()}_SHEET`,
                 errorDescription: prefix
                     ? [
                           `There were no genes and values supplied in the ${sheetName} in the exported workbook.`,
-                          `GRNsight is checking because a ${sheetName.split("_")[1] + " " + sheetName.split("_")[2]} value should have been provided as GRNmap output,`,
+                          `GRNsight is checking because a ${nameSegment[1]} ${nameSegment[2]} `,
+                          `value should have been provided as GRNmap output,`,
                           "but will not affect the display of the graph in GRNsight.",
                       ].join(" ")
                     : [
                           `The "${sheetName}" sheet was empty in the exported workbook.`,
-                          `GRNsight is checking because a ${sheetName.split("_")[1] + " " + sheetName.split("_")[2]} value should have been provided as GRNmap output,`,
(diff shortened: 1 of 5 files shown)


web-client-classic/public/js/export-warning-constants.js
```diff
@@ -53,14 +53,17 @@ module.exports = {
...
-                        ? `There was no "${sheetName}" sheet in the exported workbook.`
-                        : `The "${sheetName}" sheet was empty in the exported workbook.`,
-                    prefix
-                        ? `GRNsight is checking because a ${sheetName.split("_")[1] + " " + sheetName.split("_")[2]} value should have been provided as GRNmap output, but will not affect the display of the graph in GRNsight.`
-                        : `GRNsight has supplied ${dataSourceForTwoColumnSheet[sheetName]}.`,
-                ].join(" "),
+                errorDescription: prefix
+                    ? [
+                          `There were no genes and values supplied in the ${sheetName} in the exported workbook.`,
+                          `GRNsight is checking because a ${sheetName.split("_")[1] + " " + sheetName.split("_")[2]} value should have been provided as GRNmap output,`,
```
dondi: We have a repeated `split` call here again, so we want to do the same consolidation as before in order to call the function just once

Further, there’s a mix of `+` concatenation and template strings—for consistency, the latter is favored in modern JavaScript

So overall, I’d change this as follows: (using `nameSegment` as the variable name for the `split` result—feel free to pick the right name)

```suggestion
                          `GRNsight is checking because a ${nameSegment[1]} ${nameSegment[2]}`,
                          'value should have been provided as GRNmap output,',
```

And as you can see, finally, you’ll want to continue to be mindful of string lengths so that they interact with the line length gracefully

These attention-to-detail factors get noticed in industry and will build certain impressions with human reviewers one way or another. Automated reviewers might not remember habitual feedback but from a global picture that would mean token expenditure (and their environmental impact) that could have been avoided. So converging on good habits is beneficial either way

web-client-classic/public/js/export-warning-constants.js
```diff
@@ -53,14 +53,17 @@ module.exports = {
...
+                          "but will not affect the display of the graph in GRNsight.",
+                      ].join(" ")
+                    : [
+                          `The "${sheetName}" sheet was empty in the exported workbook.`,
+                          `GRNsight is checking because a ${sheetName.split("_")[1] + " " + sheetName.split("_")[2]} value should have been provided as GRNmap output,`,
```
dondi: Ditto on this line

dondi (commented):
Similar note to last time, plus we also have the failing test to look into. Examining the specific test that failed, the failure is due to a mismatched warning count, which this PR could have introduced due to the new warning. We can examine the result more closely when we meet in order to determine whether this calls for a change in the test or genuinely flags wrong behavior

JustAlex3125:
@dondi I've applied the proper changes that you suggested. The tests will still break because we need to decide how to approach the additional warnings which we can talk about in our meeting.

dondi (approved):
Thanks for the updates! lgtm

Issue #1602, opened 2026-06-22
Unknown workbook file causes `Cannot read properties of undefined` error


dondi:
An unknown workbook file (or files with the same characteristic) is throwing an unrecoverable in both production and beta. When this error happens, the server has to be restarted

We’ll need to (a) determine what file this is so that (b) we can pinpoint the exact issue that causes the error

This appears to be a problematic expression sheet: when read, the `data` array, or one of its elements, appears to be `undefined` when it actually shouldn’t be. Hope this narrows down the possible range of workbook files that may be at issue

> The scroll window on our running servers is very small, but I managed to capture this error on the production log:
> ```
> /data/GRNsight/server/controllers/expression-sheet-parser.js:61
>     const idLabel = sheet.data[0][0];
>                                  ^
> 
> TypeError: Cannot read properties of undefined (reading '0')
>     at parseExpressionSheet (/data/GRNsight/server/controllers/expression-sheet-parser.js:61:34)
>     at /data/GRNsight/server/controllers/expression-sheet-parser.js:176:48
>     at Array.forEach (<anonymous>)
>     at module.exports (/data/GRNsight/server/controllers/expression-sheet-parser.js:174:14)
>     at crossSheetInteractions (/data/GRNsight/server/controllers/spreadsheet-controller.js:131:26)
>     at processGRNmap (/data/GRNsight/server/controllers/spreadsheet-controller.js:368:20)
>     at /data/GRNsight/server/controllers/spreadsheet-controller.js:469:24
>     at /data/GRNsight/node_modules/multiparty/index.js:137:9
>     at /data/GRNsight/node_modules/multiparty/index.js:116:9
>     at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
> ```
> 
> On the beta log, the entire error was almost scrolled away, but I spotted the last four lines. So it appears the same error affects beta
> 
> The upshot: there is a “mystery file” which, when uploaded, causes our server to [permanently] crash. I’ll turn this into a new issue 

 _Originally posted by @dondi in [#1460](https://github.com/dondi/GRNsight/issues/1460#issuecomment-4771870706)_

dondi:
@JustAlex3125 has a lead as indicated in https://github.com/dondi/GRNsight/issues/1460#issuecomment-4771917218 so assigning this to him

kdahlquist:
[...]

PR #1605, opened 2026-07-02
Add new tests to check for if a file is blank issue #1602

JustAlex3125:
This file will break, and the tests will fail, but I am currently working on the fix to make sure that these tests pass.


server/controllers/expression-sheet-parser.js
```diff
@@ -57,8 +57,17 @@ var parseExpressionSheet = function (sheet) {
...
 
+    // Check to see if the sheet is blank
+    let idLabel = "";
+
+    try {
+        idLabel = sheet.data[0][0];
+    } catch (err) {
+        addExpError(expressionData, constants.errors.emptyWorkbookError());
+        return expressionData;
+    }
```
dondi: This works in principle, but we’ll want to revise the approach due to something we can talk about more at the next meeting, which is the “culture” of a certain language. “Culture” here means the typical idioms and approaches that gravitate more to certain languages vs. others. It’s a somewhat heady concept but it becomes clearer as one learns more and more programming languages

Here, the idiom at play is when to use a `try` clause to handle an error after the fact vs. using `if` to check for a potential error before it happens. Although both approaches can function correctly in multiple languages, some languages are inclined more toward one approach vs. the other. In this case, the former is something that Python gravitates more toward, but JavaScript leans toward the latter

So here, let’s explore what the equivalent of this statement would be via an `if` statement. How would you “rephrase” this so that it looks like this instead:

```suggestion
    if (/* what condition is equivalent to sheet.data[0][0] throwing an error? */) {
        addExpError(expressionData, constants.errors.emptyWorkbookError());
        return expressionData;
    }
```

This approach is more idiomatic to JavaScript that a `try` clause (for this situation)

server/controllers/network-sheet-parser.js
```diff
@@ -334,7 +334,12 @@ exports.workbookType = function (workbookFile) {
             sheet.name.toLowerCase() === "network" ||
             sheet.name.toLowerCase() === "network_optimized_weights"
         ) {
-            const cellA1 = sheet.data[0][0];
+            let cellA1 = "";
+            try {
```
dondi: This is similar—how would you approach this as an `if` instead of a `try`?

server/controllers/network-sheet-parser.js
```diff
@@ -334,7 +334,12 @@ exports.workbookType = function (workbookFile) {
...
+                cellA1 = sheet.data[0][0];
+            } catch (err) {
+                return (workbookType = undefined);
```
dondi: I’m unclear on the assignment here; what role does the `workbookType` variable play in the overall code? An assignment to something that is being returned anyway is generally looked at as an anti-pattern. It feels like this can just be a `return`?

```suggestion
                return;
```

dondi (commented):
Looks like the overall functionality is where we want it, but the under-the-hood implementation is a learning opportunity—see the notes for additional details

Ultimately, we want to keep the intended behavior, but how we achieve it is best done in a manner that this particular language (JavaScript) mostly does it (again see the details)

server/controllers/expression-sheet-parser.js
```diff
@@ -57,8 +57,17 @@ var parseExpressionSheet = function (sheet) {
         columnGeneNames: [],
     };
 
+    // Check to see if the sheet is blank
+    let idLabel = "";
+
+    if (sheet.data == null || sheet.data.length === 0 || sheet.data[0].length === 0) {
```
dondi: Thanks for this! Yes, this is more “JavaScript-like” than “Pythonic.” In JavaScript, there are multiple variants for this condition also, like with using the optional operator (`?`), and we can explore those as we proceed. This approach works perfectly well, particularly because JavaScript does Boolean short-circuiting (and if that is unfamiliar to you, you’ll get to it eventually or I can also talk you through it)

With all that said, there is one last adjustment: notice that, with this change, there is no longer a need to declare `idLabel` at the point where it’s being done here. The declaration can be done _after_ the `if` statement, when we are sure that `sheet.data[0][0]` is legit and so… (see comment below)

server/controllers/expression-sheet-parser.js
```diff
@@ -57,8 +57,17 @@ var parseExpressionSheet = function (sheet) {
...
+        addExpError(expressionData, constants.errors.emptyWorkbookError());
+        return expressionData;
+    }
+
+    idLabel = sheet.data[0][0];
+
     // Check that id label is correct. Throw error if not.
-    const idLabel = sheet.data[0][0];
```
dondi, line 61: …you can delete lines 60, 61, and 68, and revert back to this line, which will now be guaranteed to not throw thanks to the `if` check

server/controllers/network-sheet-parser.js
```diff
@@ -334,7 +334,11 @@ exports.workbookType = function (workbookFile) {
             sheet.name.toLowerCase() === "network" ||
             sheet.name.toLowerCase() === "network_optimized_weights"
         ) {
-            const cellA1 = sheet.data[0][0];
```
dondi, line 337: Same here—you can restore this line; just make sure it happens _after_ the `if` guard

dondi (commented):
Great, we’re very close! Just one last loose end coming out as a consequence of the new `if` check—and interestingly, this one _reduces_ the number of changes needed in the PR 😁

server/controllers/expression-sheet-parser.js
dondi: Oops missed one more thing—the first clause is best written as `!sheet.data`

JustAlex3125:
After my discussion with @kdahlquist I realized that this fix while does fix the issue does not display the proper warnings and errors for expression sheets and network sheets respectively. This issue is not ready to be pushed to beta yet. Once I correct the warnings and errors than it should be good and ready for review again.

dondi (commented):
Thanks for the updates! I saw your most recent comment about how this PR needs more work. The previously-requested changes have indeed been applied correctly, so that’s good. lmk when this is ready for `beta`. Thanks!

server/controllers/expression-sheet-parser.js
```diff
@@ -57,8 +57,18 @@ var parseExpressionSheet = function (sheet) {
         columnGeneNames: [],
     };
 
-    // Check that id label is correct. Throw error if not.
+    // Check to see if the sheet is blank
+    if (!sheet.data || sheet.data.length === 0 || sheet.data[0].length === 0) {
+        addExpWarning(
```
dondi, line 62: I realize that this function isn’t new to this PR, but seeing this now in this context, I think we should spell out its name. We have two active “exp” terms in GRNsight: “export” and “expression.” With this function name, it isn’t immediately clear whether we mean `addExportWarning` or `addExpressionWarning`

I mean, it isn’t hard to eventually figure it out from context, but if we can alleviate even that bit of cognitive load from the code reader, then all the better

Recognizing though that this isn’t new to this branch, so will write this up as a separate issue

server/controllers/workbook-constants.js
```diff
@@ -478,6 +494,30 @@ module.exports = {
...
+            };
+        },
+
+        missingGeneNamesWarning: function (sheetName) {
+            return {
+                warningCode: "MISSING_GENE_NAME",
+                errorDescription: `Gene names in column A of the "${sheetName}"
+                    sheet are missing one or more genes from the network sheet. This will not affect the functionality of the graph in GRNsight.`,
+            };
+        },
```
dondi, line 519: Not a huge deal, but these descriptions all look long enough to merit the `[].join(" ")` treatment, to get shorter code lines. Can be dealt with later, but should also be an on-going standard practice when writing out long strings

dondi (approved):
lgtm functionality-wise, with some notes for future rounds of changes. One has been written up as a distinct issue #1623 while the other should be more of a standard practice as we touch long string literals in the future (see individual comments for details)

Issue #1387, opened 2026-03-11
Select All checkbox in Export to Excel modal not behaving properly


kdahlquist:
Now that we have further refined what is happening in the Export to Excel functionality, we need to take a look at the behavior of the checkboxes in the modal.

Since we are defaulting to `network_weights` not being checked, then the `Select all` checkbox should not be checked by default either.

The `Select all` checkbox should only be checked when literally all possible boxes are checked.  So, if `network_optimized_weights` is not available, then the `Select all` checkbox should not be available, unless we want to exclude that sheet from the all behavior.

If all sheets are checked, then `Select all` should be checked.  Unchecking a single sheet should then uncheck `Select all`.  

It's behaving sort of wonky right now.

kdahlquist:
As noted in [this comment](https://github.com/dondi/GRNsight/issues/1418#issuecomment-4245640043) on issue #1418, this is the expected behavior with regard to the `*_sigmas` sheets:

>Upon reflection, I think the simplest way to think about it is to treat the `*_sigmas` sheets as expression sheets with regard to the checkbox behavior in the modal.
>
>Note that this behavior would need to be implemented for Demo 2 and also any imported workbook that had `*_sigmas` worksheets in it.
>
>- You need to uncheck _all_ of the `*_log2_expression`, `*_log2_optimized_expression`, and `*_sigmas` sheets before the Expression Data source switches to `None` and the checkboxes disappear.
>
>I think that seeing the _checked_ `*_sigmas` disappear is what triggered us thinking that this was a bug, so if we require them to be unchecked before they disappear, it will be more in alignment with the behavior that we expect.
>
>- The corollary to this is that if the `Select All` option is unchecked, then the Expression Sheets options should disappear and the Expression Data Source should change to `None`.
>
>I'll repeat this comment in #1387 so it's clear in both issues what the expected behavior is.

kdahlquist:
We decided in the meeting on 4/14 that @jiacelyne will tackle this issue solo along with #1418 this week.

One more observation: I loaded Demo 2 and then selected the menu item Export to Excel.  In the modal, all of the boxes were checked _except_ `network_weights`.  When I checked `network_weights`, then the `Select All` became _unchecked_ even though now all checkboxes were legitimately checked.

jiacelyne:
The issue seems to be stemming from the same location around #1418, so I will continue looking around that area. The network_weights continuously acts as the checkbox that does not cooperate with the Select All option.

[...]

PR #1607, opened 2026-07-07
#1387 Correctly branching from beta 

jiacelyne:
Pull Request Checklist
- [x] Print works as expected

### DIFF
--- a/database/schema/README.md
+++ b/database/schema/README.md
@@ -23,10 +23,6 @@ For example, to access your local PostgreSQL database, use the following command
 ```
 psql postgresql://localhost/postgres
 ```
-or
-```
-psql -U postgres postgresql://localhost/postgres
-```
 
 ## Creating Schemas and Adding Table Specifications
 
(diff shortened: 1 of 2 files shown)


database/schema/README.md
dondi, line 1: The provenance of the branch looks good now, but it looks like your first commit (0793d5ef9662f1d062d733e79c1be89f99e04c31) piggy-backed this file unintentionally. We can talk about how that might have happened this time and what can be done to avoid it, but in terms of finalizing this PR, can you redo the branch one more time and make sure that your committed changes have just your intended changes _exclusively_. The redo will also help you determine how/why this file found its way to your commit

web-client-classic/public/js/upload.js
```diff
@@ -55,14 +55,24 @@ export const filenameWithExtension = function (mode, genes, edges, type, extensi
...
+            source = $("input[name=expressionSource]:checked")[0].value;
+            if (source === "none") {
+                source = null;
+            } else if (source === "userInput") {
+                // only demos will have an expression source
+                source = grnState.workbook.expression.source
+                    ? grnState.workbook.expression.source
+                    : "user-data";
+            }
+        } else {
```
dondi, line 74: For now, we can keep this code as-is. But for future reference—and we’ll see if there’s time to discuss this at the meeting—what the code is doing here represents a higher-level programming pattern of imperative code whose sole result is the value of a single variable (`source`). When you have code that does this, there is an alternative approach that does not rely on side effects (and again if you’re unfamiliar with the term, we can discuss that)

In broader practice, this latter alternative (which will also be good for Alex to know) is generally considered safer and more robust, so even though what you have here is functionally equivalent, in terms of programming craft, the alternative is recommended. We can discuss this at some future juncture; it doesn’t need to block this PR

dondi (commented):
OK, things are definitely cleaner here but there appears to be one last loose end (_database/schema/README.md_). See if you can produce one last instance of this branch that is 100% _just_ your intended change for #1387

dondi:
OK, so I see now what happened, which is that you did a `cherry-pick` of commit 7a235b879a615b54b3cceb6c072d8c5a9fdc9edd—and yes, this was the instruction from last time ✅

But also stepping back, note the _context_ of the instruction: the purpose was to produce a clean `jia-1387` branch that exclusively includes just your work. So in a broader sense, upon following the instruction, spotted the oversight, and either addressed it beforehand or asked a question on how to resolve it. On the one hand, I could have looked at my directive more closely to be more precise about what to do…and yep, sorry, I own that miss 😕

On the other hand, you also have agency here and can continue to hold your own counsel as you follow steps. This summer work is a great opportunity for you to advance in that regard

dondi:
This has been superseded by #1609, so closing this one directly

PR #1608, opened 2026-07-09
Swapped most of the assert.equals in test.js

JustAlex3125:
Went in and adjusted most of the assert.equals in the test.js file to properly show the expected and actual values in the correct place.


dondi (approved):
LGTM

PR #1609, opened 2026-07-09
#1387 unchecked expression sheet and radio button expression data source is…

jiacelyne:
… consistent with exported filenames


dondi (approved):
All cleaned up! Thanks for sticking with this—lgtm

Issue #1613, opened 2026-07-13
Export to Excel: if network_optimized_weights box defaults to unchecked, then it should not be an option in the list


kdahlquist:
Demos 1 and 3 do not have a network_optimized_weights sheet.  Do the following:
- Load Demo 1 or 3
- Export Data > To Excel
- The checkbox for the network_optimized_weights sheet correctly defaults to unchecked because there is no data for it in these demo files, and the box is correctly unavailable to check.  Because this is unchecked, the Select All checkbox should also be unchecked, but currently in beta v7.9.5, it is checked.

kdahlquist:
Alternate solution.  If there is no network_optimized_weights available, let's not even include it as a checkbox at all in the list.

jiacelyne:
[PR created](https://github.com/dondi/GRNsight/pull/1621)

kdahlquist:
Complete in beta v7.9.6

PR #1621, opened 2026-07-17
#1613 excel export removes network optimized weight box if unapplicable

jiacelyne:
Pull Request Checklist
- [x] Print works as expected

### DIFF
--- a/web-client-classic/public/js/upload.js
+++ b/web-client-classic/public/js/upload.js
@@ -663,9 +663,10 @@ export const upload = function () {
             </li>
             `;
         let networkOptimizedWeights = networks[1];
-        result =
-            result +
-            `
+        if (networkOptimizedWeights[0]) {
+            result =
+                result +
+                `
             <li class=\'export-excel-workbook-sheet-option\'>
                 <input type=\'checkbox\' name=\'workbookSheets\' ${networkOptimizedWeights[0] ? 'checked="true"' : ""} value=\"${networkOptimizedWeights[1]}\" id=\'exportExcelWorkbookSheet-${networkOptimizedWeights[1]}\' class=\'export-checkbox\' ${networkOptimizedWeights[0] ? "" : "disabled"}/>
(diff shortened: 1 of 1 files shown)


web-client-classic/public/js/upload.js
```diff
@@ -663,16 +663,18 @@ export const upload = function () {
             </li>
             `;
         let networkOptimizedWeights = networks[1];
-        result =
-            result +
-            `
+        if (networkOptimizedWeights[0]) {
+            result =
+                result +
+                `
```
dondi, line 669: Template strings capture string generation operations more generally than plain concatenation (`+`) these days, and so this is now the favored approach. You’re already using a template string here anyway, so you may as well adopt it fully:

```suggestion
        if (networkOptimizedWeights[0]) {
            result =
                `${result}
```

web-client-classic/public/js/upload.js
```diff
@@ -663,16 +663,18 @@ export const upload = function () {
...
                 <input type=\'checkbox\' name=\'workbookSheets\' ${networkOptimizedWeights[0] ? 'checked="true"' : ""} value=\"${networkOptimizedWeights[1]}\" id=\'exportExcelWorkbookSheet-${networkOptimizedWeights[1]}\' class=\'export-checkbox\' ${networkOptimizedWeights[0] ? "" : "disabled"}/>
                 <label for=\'exportExcelWorkbookSheet-${networkOptimizedWeights[1]}\' id=\'exportExcelWorkbookSheet-${networkOptimizedWeights[1]}-label\' class=\'export-checkbox-label\' >
                     ${networkOptimizedWeights[1]}
                 </label>
             </li>
             `;
+        }
         let networkWeights = networks[2];
         result =
             result +
```
dondi, line 680: This is a concatenation too, but it isn’t involved in this branch. As you work on more changes, keep an eye out for concatenations like these and seek to migrate as you go

dondi (commented):
Functionally OK, but style-wise let’s start using template strings more comprehensively

dondi (approved):
Since this is a lone style note and should not affect functionality, I’ll go ahead and merge this but let’s plan on making this revision at the next opportunity

Issue #1393, opened 2026-03-18
Add warnings for Missing Genes + Value Combinations for Two Column Sheets


ntran18:
## Missing Gene + Value Combinations

-   [x] h_3-genes_6 edges_missing-first-gene-and-deg-rate-value.xlsx
-   [x] h_3-genes_6 edges_missing-first-gene-and-prod-rate-value.xlsx
-   [x] h_3-genes_6 edges_missing-first-gene-and-threshold_b-value.xlsx
-   [x] i_3-genes_6 edges_missing-middle-gene-and-deg-rate-value.xlsx
-   [x] i_3-genes_6 edges_missing-middle-gene-and-prod-rate-value.xlsx
-   [x] i_3-genes_6 edges_missing-middle-gene-and-threshold_b-value.xlsx
-   [x] j_3-genes_6 edges_missing-last-gene-and-deg-rate-value.xlsx
-   [x] j_3-genes_6 edges_missing-last-gene-and-prod-rate-value.xlsx
-   [x] j_3-genes_6 edges_missing-last-gene-and-threshold_b-value.xlsx

- Import -> Warning:
   - Degradation: GRNsight has detected that there are missing genes and values for degradation rates in the imported workbook's 'degradation_rates' sheet. A degradation rate will need to be supplied to use this workbook as an input file for GRNmap, but will not affect the display of the graph in GRNsight. The missing genes and values are: <list of genes' name>
   - Production: GRNsight has detected that there are missing genes and values for production rates in the imported workbook's 'production_rates' sheet. A production rate will need to be supplied to use this workbook as an input file for GRNmap, but will not affect the display of the graph in GRNsight. The missing genes and values are: <list of genes' name>
   - Threshold b: GRNsight has detected that there are missing genes and threshold b values in the imported workbook's 'threshold_b' sheet. A threshold b value will need to be supplied to use this workbook as an input file for GRNmap, but will not affect the display of the graph in GRNsight. The missing genes and values are: <list of genes' name>

- Export -> Warning:
  - If not asking to export that sheet -> Do nothing
  - If yes, use user imported data, no fetching from database, add the following warnings
    - Degradation: GRNsight has detected that there are missing genes and values for degradation rates in the exported workbook's 'degradation_rates' sheet because they were missing in the imported workbook. A degradation rate will need to be supplied to use this workbook as an input file for GRNmap, but will not affect the display of the graph in GRNsight. The missing genes and values are: <list of genes' name>
[...]

PR #1622, opened 2026-07-17
Updated the warning for missing genes and values

JustAlex3125:
Fixed the two-column warnings for files lettered h-j to fix issue #1393. Also, I think this takes third place for the smallest PR, possibly.

### DIFF
--- a/server/controllers/workbook-constants.js
+++ b/server/controllers/workbook-constants.js
@@ -416,7 +416,7 @@ module.exports = {
             const prefix = sheetName.includes("optimized");
             return {
                 warningCode: `MISSING_GENES_AND_VALUES_${sheetName.toUpperCase()}`,
-                errorDescription: prefix
+                errorDescription: !prefix
                     ? [
                           `GRNsight has detected that there are missing genes and ${valuesForEachTwoColSheet[sheetName]}`,
                           `in the imported workbook's ${sheetName} sheet.`,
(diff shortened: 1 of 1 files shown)


server/controllers/workbook-constants.js
dondi: I appreciate the smallness of the PR 😁 but in this case, we have another case of “forest/trees” swap—although adding the `!` is a fix in and of itself, note that it adds noise: you may as well just swap the conditional arms 😅

```suggestion
                errorDescription: prefix
? [
`There were no genes and values supplied in the ${sheetName} sheet in the imported workbook`,
grnMapInputSuppliedWarningMessage(sheetName, value),
].join(" "),
: [
`GRNsight has detected that there are missing genes and ${valuesForEachTwoColSheet[sheetName]}`,
`in the imported workbook's ${sheetName} sheet.`,
grnMapInputSuppliedWarningMessage(sheetName, value),
`The missing genes and values are: ${missingGenes}.`,
].join(" ")
```

Hope you’ll agree, it’s cognitively less demanding to read just `prefix` as the condition, then see the results when it’s true vs. false, rather than reading `!prefix` then making the mental switch that “oh yeah, the first option is actually when `prefix` _isn’t_ true…”

dondi: P.S. The suggestion might not produce the right formatting so please make sure to auto-format after the change is made

dondi (commented):
This one I think I’ll hold until it’s fixed—although it’s technically a style comment, it does involve a genuine logic change (swapping the value of a boolean condition) so it will be best to make the change first before merging

JustAlex3125:
@dondi Should be fixed now :) 

The idea of making the `prefix` read how the first one is when is true makes a lot of sense. Honestly was just trying to make the commit as fast and easy as possible 😇

dondi (approved):
Thanks for this! Yes, long-term readability takes precedence over short-term small diffs 😄

PR #1624, opened 2026-07-27
Jia 1617

jiacelyne:
I'm not sure why filenameWithExtension got put in the original code again, so I deleted it to make sure there arent two of the same variables. 

Pull Request Checklist
- [x] Print works as expected

### DIFF
--- a/web-client-classic/public/js/upload.js
+++ b/web-client-classic/public/js/upload.js
@@ -763,7 +763,8 @@ export const upload = function () {
                         $(".export-excel-workbook-sheet-option-subheader").remove();
                         handleExpressionSheetsFromSource(source);
                     }
-                });
+                }
+            );
         } else if (grnState.mode === NETWORK_PPI_MODE) {
             const source = "userInput";
             $("#exportExcelForm").remove();
--- a/web-client-classic/public/js/upload.js
+++ b/web-client-classic/public/js/upload.js
@@ -55,7 +55,7 @@ export const filenameWithExtension = function (mode, genes, edges, type, extensi
         filename = filename.substr(0, filename.length - currentExtension[0].length);
     }
     if (mode === NETWORK_GRN_MODE && extension === "xlsx") {
-        source = $("input[name=expressionSource]:checked")[0].value;
+        source = $("#expressionSourceDropdown").val();
         if (source === "none") {
             source = null;
         } else if (source === "userInput") {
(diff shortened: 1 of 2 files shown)


web-client-classic/public/js/upload.js
```diff
@@ -549,81 +507,35 @@ export const upload = function () {
     };
 
     const createHTMLforGRNForm = () => {
+        return `
+        <form id='exportExcelForm'>
```
dondi, line 511: More of a cleanup, not a blocker: with the full use of template strings here, you are no longer bound in terms of what quotes to use. You may have noticed a mix of quote styles—these are all vestiges of prior versions predating template strings

With template strings, we can now standardize on the quotes most typically associated with HTML tags: double-quotes `"`

```suggestion
        <form id="exportExcelForm">
```

web-client-classic/public/js/upload.js
```diff
@@ -549,81 +507,35 @@ export const upload = function () {
...
+            <div class='form-group export-form-group'>
+                <p id='exportExcelExpressionSources'></p>
+                <ul class='exportExcelWorkbookSheets' id='export-excel-workbook-sheet-list' style="list-style-type:none; padding-left:0;"></ul>
```
dondi, line 514: Notice how this line has a mix: we can standardize those

Further, however, you have inline styles: traditionally a no-no with HTML, to be used only in very specific circumstances, and this is not it. Ideally these go in a class—it can be `.exportExcelWorkbookSheets` since you’re assigning that already

```suggestion
                <ul class="exportExcelWorkbookSheets" id="export-excel-workbook-sheet-list"></ul>
```

(here, `list-style-type: none; padding-left: 0` go to the _.styl_ file, wherever `.exprtExcelWorkbookSheets` is defined

web-client-classic/public/js/upload.js
```diff
@@ -549,81 +507,35 @@ export const upload = function () {
...
-            <li class=\'export-excel-workbook-sheet-option export-excel-workbook-sheet-option-subheader\'>
-                <input type=\'checkbox\' name=\'workbookSheets\' checked=\"true\" value=\"select all\" id=\'exportExcelWorkbookSheet-All\' class=\'export-checkbox\' />
-                <label for=\'exportExcelWorkbookSheet-All\' id=\'exportExcelWorkbookSheet-All-label\' class=\'export-checkbox-label\' >
-                    Select All
-                </label>
-            </li>
-
-            <p class=\'export-excel-workbook-sheet-option-subheader\'> Network Sheets </p>
-            `;
+        <li class=\'export-excel-workbook-sheet-option export-excel-workbook-sheet-option-subheader\'>
```
dondi, line 530: Nice code reduction! And these lines also benefit from quote standardization

```suggestion
        <li class="export-excel-workbook-sheet-option export-excel-workbook-sheet-option-subheader">
```

web-client-classic/public/stylesheets/grnsight.styl
```diff
@@ -616,27 +616,73 @@ path.mousezone
   margin-left: auto;
 }
 
+.export-form-group
+  padding: 0 4px
+
+.exportExcelWorkbookSheets
+  margin: 0
```
dondi, line 623: Ah look—here it is!
```suggestion
  list-style-type: none
  margin: 0
  padding-left: 0
```

dondi (approved):
The gist of this looks good, with none of the comments changing functionality (well, not overtly) so I’ll go ahead and merge so this can be deployed

However, the style/standardization aspects are still worth doing. We can go over any questions during the meeting

dondi:
I’m keeping this branch open so that source code-level comments can be implemented. We can merge again after that round is done

PR #1639, opened 2026-08-15
Corrected blank expression sheet warning

JustAlex3125:
Corrected the export warning whenever a blank expression sheet is imported and then exported

### DIFF
--- a/web-client-classic/public/js/upload.js
+++ b/web-client-classic/public/js/upload.js
@@ -229,11 +229,12 @@ export const upload = function () {
                       return a - b;
                   })
                 : null;
-            const simTimepoints = expTimepoints
-                ? Array.from(Array(expTimepoints[expTimepoints.length - 1] + 1).keys()).filter(
-                      x => x % 5 === 0
-                  )
-                : null;
+            const simTimepoints =
+                expTimepoints && expTimepoints.length > 0
+                    ? Array.from(Array(expTimepoints[expTimepoints.length - 1] + 1).keys()).filter(
+                          x => x % 5 === 0
+                      )
+                    : null;
             const strain =
                 expression.length > 0 ? expression.map(x => removeExpressionSuffix(x)) : null;
             if (expTimepoints) {
@@ -409,6 +410,13 @@ export const upload = function () {
             } else if (isExpressionSheet(sheet)) {
                 finalExportSheets.expression[sheet] =
                     source === "userInput" ? grnState.workbook.expression[sheet] : null;
+                if (finalExportSheets.expression[sheet]) {
+                    for (let warning of finalExportSheets.expression[sheet].warnings) {
+                        if (warning.warningCode === "BLANK_EXPRESSION_SHEET") {
+                            finalExportSheets.warnings.push(warning);
+                        }
+                    }
+                }
             } else {
                 finalExportSheets.two_column_sheets[sheet] = grnState.workbook.twoColumnSheets
                     ? grnState.workbook.twoColumnSheets[sheet]
--- a/web-client-classic/public/js/upload.js
+++ b/web-client-classic/public/js/upload.js
@@ -231,9 +231,10 @@ export const upload = function () {
                 : null;
             const simTimepoints =
                 expTimepoints && expTimepoints.length > 0
-                    ? Array.from(Array(expTimepoints[expTimepoints.length - 1] + 1).keys()).filter(
-                          x => x % 5 === 0
-                      )
+                    ? Array.from(
+                          { length: expTimepoints[expTimepoints.length - 1] + 1 },
+                          (_, index) => index
+                      ).filter(timepoint => timepoint % 5 === 0)
                     : null;
             const strain =
                 expression.length > 0 ? expression.map(x => removeExpressionSuffix(x)) : null;
@@ -411,11 +412,9 @@ export const upload = function () {
                 finalExportSheets.expression[sheet] =
                     source === "userInput" ? grnState.workbook.expression[sheet] : null;
                 if (finalExportSheets.expression[sheet]) {
-                    for (let warning of finalExportSheets.expression[sheet].warnings) {
-                        if (warning.warningCode === "BLANK_EXPRESSION_SHEET") {
-                            finalExportSheets.warnings.push(warning);
-                        }
-                    }
+                    finalExportSheets.expression[sheet].warnings
+                        .filter(({ warningCode }) => warningCode === "BLANK_EXPRESSION_SHEET")
+                        .forEach(warning => finalExportSheets.warnings.push(warning));
                 }
             } else {
                 finalExportSheets.two_column_sheets[sheet] = grnState.workbook.twoColumnSheets


JustAlex3125:
Pull Request Checklist
- [x] Print works as expected

web-client-classic/public/js/upload.js
```diff
@@ -229,11 +229,12 @@ export const upload = function () {
...
                   })
                 : null;
-            const simTimepoints = expTimepoints
-                ? Array.from(Array(expTimepoints[expTimepoints.length - 1] + 1).keys()).filter(
-                      x => x % 5 === 0
-                  )
-                : null;
+            const simTimepoints =
+                expTimepoints && expTimepoints.length > 0
+                    ? Array.from(Array(expTimepoints[expTimepoints.length - 1] + 1).keys()).filter(
```
dondi: It took me a minute to parse out what you’re doing here, but I figured it out eventually 😅 Wouldn’t Python’s `range` have been nice huh?

As fun of a technique as this is, it’s hard to read the intent unless the code reader has seen it before. This produces the same result afaict, and reads a tiny bit better because it makes it clearer that you want an array of a particular length:
```suggestion
                    ? Array.from({ length: expTimepoints[expTimepoints.length - 1] + 1 }, (_, index) => index).filter(
```

`length` makes the intent more explicit, and the removal of `keys` avoids potential confusion with dictionaries (i.e., generic objects in JavaScript). `(_, index) => index` is then a function that returns the indices of the array (`index` is better associated with arrays than a “key”). So although it is longer text-wise, I think it takes less work to process mentally

web-client-classic/public/js/upload.js
```diff
@@ -229,11 +229,12 @@ export const upload = function () {
...
+                          x => x % 5 === 0
```
dondi: Also since at this point, we know what we’re dealing with (timepoints, afaict), may as well be more unequivocal that this is the variable at play here
```suggestion
                          timepoint => timepoint % 5 === 0
```

web-client-classic/public/js/upload.js
```diff
@@ -229,11 +229,12 @@ export const upload = function () {
...
+                      )
+                    : null;
```
dondi, line 238: …now having said all that, especially after my big soapbox speech about modern alternatives to `for` loops, admittedly I’d say this is one of those exceptions 😅 When operating on raw sequences of numbers, yeah the classic `for` loop is more efficient. Things like `forEach`, `map`, `filter`, `reduce`, and friends primarily assume that you are processing the data _within_ an array, and not the indices themselves. Here, ultimately ultimately, a `for` loop with a step of `index += 5` is ultimately the leanest way to do this

Kudos on the initiative at doing it this way though! The more you do software dev, I’d say you _will_ see this as the better approach more often, on the long term. You just happened to encounter an exception earlier than anticipated 😊

web-client-classic/public/js/upload.js
```diff
@@ -409,6 +410,13 @@ export const upload = function () {
             } else if (isExpressionSheet(sheet)) {
                 finalExportSheets.expression[sheet] =
                     source === "userInput" ? grnState.workbook.expression[sheet] : null;
+                if (finalExportSheets.expression[sheet]) {
+                    for (let warning of finalExportSheets.expression[sheet].warnings) {
+                        if (warning.warningCode === "BLANK_EXPRESSION_SHEET") {
+                            finalExportSheets.warnings.push(warning);
+                        }
+                    }
```
dondi: Now this one _does_ lend itself to the suite of array functions. If I’m reading this right, you’d like to push the expression sheet warnings to the top-level `warnings` array. Try this on for size:
```suggestion
                    finalExportSheets.expression[sheet].warnings.filter(
                        ({ warningCode }) => warningCode === "BLANK_EXPRESSION_SHEET"
                    ).forEach(
                        warning => finalExportSheets.warnings.push(warning)
                    )
```

dondi: Note how the `filter` function reads more explicitly (to me at least) that “I want to pick out certain warnings from the expression sheet” (with the condition captured by the function), with `forEach` then saying “push each of those to `finalExportSheets.warnings`”

In the end, there is no particularly right or wrong way to do this, but at your learning stage, it’s good to know what the _alternatives_ are, so that if the question comes up in the future, you will at least be familiar with the options and will have a general notion of each approach’s pros and cons

dondi (commented):
Some fun with array techniques here! Of the two blocks, I think the first one provides greater motivation for revision. The second one is almost a judgment call, so compare the two, understand the nuances, and make a decision for yourself. As stated in the comment, that is mentioned mainly to broaden your awareness of the range of approaches/techniques for doing what is operationally the same thing

JustAlex3125:
@dondi I think the Github differences picked up a lot more of the code than I actually changed. I only added the `&& expTimepoints.length > 0` on line 233. I think your suggestions make sense but I wasn't totally sure what was going on with the code either so I just left it alone. When I added that line of code the autoformator moved all those lines of code down a line which made it seem like I changed everything else. Do you want to make these changes now or just leave the code the way it is? I was focused on fixing the bug but I can change whatever else you want.

web-client-classic/public/js/upload.js
```diff
@@ -229,11 +229,12 @@ export const upload = function () {
                       return a - b;
                   })
                 : null;
-            const simTimepoints = expTimepoints
-                ? Array.from(Array(expTimepoints[expTimepoints.length - 1] + 1).keys()).filter(
-                      x => x % 5 === 0
-                  )
-                : null;
+            const simTimepoints =
+                expTimepoints && expTimepoints.length > 0
```
dondi, line 233: Yes I see what happened here now. The addition of `&& expTimepoints.length > 0` reformatted the conditional, and although that code itself is unchanged, the change in indentation made it show up like a change

In industry settings where I’ve seen code reviews happening, anything that the reviewer notices is fair game. I’ve seen a lot of reviews that have said “I realize that you didn’t change this line, but I noticed…”

The general rule with incidental changes like these are that they are more about style or readability, and should not change functionality. Changing that would go outside what the ticket needed to accomplish

In this case, the proposed change is purely readability without changing functionality. If you are unsure, try both versions of the code on some sample input and see if you get the same results. If you get the same results, I’d say it’s best to favor the [slightly] more readable version—or maybe even the `for` loop mentioned in the last comment

The reason the vibe is “may as well change this now” is that in many cases, if one doesn’t make an improvement now, then it might not be improved for a very long time. Another way to say it is “strike while the iron is hot”

Sorry this is a long-winded way of saying “please do it anyway” but I wanted to provide a rationale so that you know it wasn’t an arbitrary preference. It’s driven by what I’ve seen before and how I’ve seen some improvements just never get done because the devs are always turning their attention to something else

dondi (commented):
Thanks for the clarification—I wrote my response at “ground zero,” line 233 😊 Thinking about it more, doing this refactor also has learning value, as you gain firsthand experience with crafting multiple alternatives for performing the same operation

web-client-classic/public/js/upload.js
```diff
@@ -409,6 +410,13 @@ export const upload = function () {
             } else if (isExpressionSheet(sheet)) {
                 finalExportSheets.expression[sheet] =
                     source === "userInput" ? grnState.workbook.expression[sheet] : null;
+                if (finalExportSheets.expression[sheet]) {
+                    for (let warning of finalExportSheets.expression[sheet].warnings) {
+                        if (warning.warningCode === "BLANK_EXPRESSION_SHEET") {
+                            finalExportSheets.warnings.push(warning);
+                        }
+                    }
+                }
```
dondi, line 418: Oh, and _this_ diff isn’t a reformat, right? It reads as a genuine addition and not a substitution

JustAlex3125: That is correct this is apart of addition. Sorry I didn't catch this one I got messed up with the reformat as well.

dondi (approved):
Thanks for the updates! Worthwhile for both readability improvement and the additional exposure that you have now gotten to the various alternatives for processing arrays 👌🏽

PR #1640, opened 2026-08-17
Added ee expression test files to repository

JustAlex3125:
Added tests for expression files ee


JustAlex3125:
Pull Request Checklist
- [x] Print works as expected

test/expression-data-import-tests.js
```diff
@@ -174,4 +174,25 @@ describe("expression-data-import-tests", function () {
...
+                "test-files/expression-data-test-sheets/optimized_expression_sheet_missing_one_value.xlsx",
+                0
+            );
+        });
+        it("Should return a warning with one missing value in sigma sheet.", function () {
+            test.noWarnings(
+                "test-files/expression-data-test-sheets/sigma_sheet_missing_one_value.xlsx",
+                0
+            );
+        });
```
dondi, line 196: This is a good foundation, protecting us from regressions where warnings may then get generated. Some follow-ups though:
* Do we have a `noErrors` utility also? Because to further codify the existing behavior, I presume we’ll want that also
* Improving further on the coverage of these tests, can we capture this part as a test:

> GRNsight correctly handles this for node coloring and will export the worksheet with the correct missing value.

It’s when we can capture behaviors like this that test suites really wild up their value. Having this captured in code will allay future doubts about “um, was _this_ how the code was supposed to work?” (you have probably heard this come up already about older pre-existing functions 😄)

Short of having infinite accurate recall, it’s the presence of unit tests that represent the message from past developers to current ones: “Yes, that’s what we meant for it to do”

Acknowledged that this is pushing the envelope a bit. Give it a shot and let’s see how far we can get

dondi (commented):
Good start, and ideally we can go further but we can discuss the scope when we meet

dondi (approved):
Cool, we can go with the `noErrors` addition for now. As for the broader coverage, we can bring that up as a meeting topic

PR #1641, opened 2026-08-18
Updating two column warnings

JustAlex3125:
Finished updating and reviewing all two column warning tests.

### DIFF
--- a/server/controllers/validators/two-column-warnings.js
+++ b/server/controllers/validators/two-column-warnings.js
@@ -213,18 +213,16 @@ const applyTwoColumnSheetWarnings = (
     // Check missing genes
     if (missingGenes.length === genesInNetwork.length) {
         warningsToAdd.push(
-            constants.warnings.missingAllGenesAndValues(sheetName, /*isAllGenesMissing=*/ true)
+            constants.warnings.missingAllGenesAndValues({
+                sheetName: sheetName,
+                allGenesMissing: true,
+                missingGenes: genesMissingValue.join(", "),
+            })
         );
     } else if (missingGenes.length > 0) {
-        let geneIdWarningPresent;
-        for (let i = 0; i < warningsToAdd.length; i++) {
-            if (
-                warningsToAdd[i].warningCode.includes("MISSING_GENE_IDS_WITH_VALUES") ||
-                warningsToAdd[i].warningCode.includes("MISMATCH_GENE_IDS_WITH_VALUES")
-            ) {
-                geneIdWarningPresent = true;
-            }
-        }
+        const geneIdWarningPresent = warningsToAdd.some(({ warningCode }) =>
+            warningCode.includes("MISSING_GENE_IDS_WITH_VALUES")
+        );
         if (!geneIdWarningPresent) {
             warningsToAdd.push(
                 constants.warnings.missingGenesAndValuesWarningWhenImporting(
@@ -238,11 +236,11 @@ const applyTwoColumnSheetWarnings = (
     // Check missing values
     if (genesInSheet.length > 0 && genesMissingValue.length === genesInSheet.length) {
         warningsToAdd.push(
-            constants.warnings.missingAllGenesAndValues(
-                sheetName,
-                /*isAllGenesMissing=*/ false,
-                genesMissingValue.join(", ")
-            )
+            constants.warnings.missingAllGenesAndValues({
+                sheetName: sheetName,
+                allGenesMissing: false,
+                missingGenes: genesMissingValue.join(", "),
+            })
         );
     } else if (genesMissingValue.length > 0) {
         warningsToAdd.push(
--- a/server/controllers/validators/two-column-warnings.js
+++ b/server/controllers/validators/two-column-warnings.js
@@ -216,18 +216,30 @@ const applyTwoColumnSheetWarnings = (
             constants.warnings.missingAllGenesAndValues(sheetName, /*isAllGenesMissing=*/ true)
         );
     } else if (missingGenes.length > 0) {
-        warningsToAdd.push(
-            constants.warnings.missingGenesAndValuesWarningWhenImporting(
-                sheetName,
-                missingGenes.join(", ")
-            )
-        );
+        let geneIdWarningPresent;
+        for (let i = 0; i < warningsToAdd.length; i++) {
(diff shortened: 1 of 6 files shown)


JustAlex3125:
Pull Request Checklist
- [x] Print works as expected

server/controllers/validators/two-column-warnings.js
```diff
@@ -216,18 +216,33 @@ const applyTwoColumnSheetWarnings = (
...
-        );
+        let geneIdWarningPresent;
+        for (let i = 0; i < warningsToAdd.length; i++) {
+            if (
+                warningsToAdd[i].warningCode.includes("MISSING_GENE_IDS_WITH_VALUES") ||
+                warningsToAdd[i].warningCode.includes("MISMATCH_GENE_IDS_WITH_VALUES")
+            ) {
+                geneIdWarningPresent = true;
+            }
+        }
```
dondi: More array fun coming your way 😊 _This_ pattern also has a corresponding function: `some`. It’s basically, “is it true that there is something in this array that meets this condition?” Check it out:
```suggestion
        const geneIdWarningPresent = warningsToAdd.some(
            ({ warningCode }) => (
                warningCode.includes("MISSING_GENE_IDS_WITH_VALUES") ||
                warningCode.includes("MISMATCH_GENE_IDS_WITH_VALUES")
            )
        )
```

server/controllers/validators/two-column-warnings.js
```diff
@@ -216,18 +216,33 @@ const applyTwoColumnSheetWarnings = (
...
     }
 
     // Check missing values
     if (genesInSheet.length > 0 && genesMissingValue.length === genesInSheet.length) {
         warningsToAdd.push(
-            constants.warnings.missingAllGenesAndValues(sheetName, /*isAllGenesMissing=*/ false)
+            constants.warnings.missingAllGenesAndValues(
+                sheetName,
+                /*isAllGenesMissing=*/ false,
```
dondi: Oh nice comment! This hints that `missingAllGenesAndValues` might benefit from a refactor such that its arguments are clumped into one object. But that can be for a future revision

server/controllers/workbook-constants.js
```diff
@@ -438,24 +438,34 @@ module.exports = {
...
+                          `GRNsight has detected that there are missing genes and values for ${valuesForEachTwoColSheet[sheetName]}`,
                           `in the imported workbook's ${sheetName} sheet.`,
                           grnMapInputSuppliedWarningMessage(sheetName, value),
                           `The missing genes and values are: ${missingGenes}.`,
                       ].join(" "),
             };
         },
 
-        missingAllGenesAndValues: function (sheetName, isAllGenesMissing) {
+        missingAllGenesAndValues: function (sheetName, isAllGenesMissing, missingGenes) {
```
dondi: Regarding the aforementioned revision, this is what it would look like:
```suggestion
        missingAllGenesAndValues: function ({ sheetName, allGenesMissing, missingGenes }) {
```

The curly braces turn the three parameters into a single one with an object format. This in turn will force function calls to supply the names, like so:
```javascript
constants.warnings.missingAllGenesAndValues({
  sheetName: ....., // And you won’t need the `:` if the variable is already called `sheetName`
  allGenesMissing: true|false,
  missingGenes: .....
})
```

This will obviate the need for a comment

web-client-classic/public/js/two_column_sheets_warnings.js
```diff
@@ -27,10 +27,19 @@ const getChosenTwoColumnSheets = chosenSheets => {
...
     if (!importWarning) return null;
+
+    const additionalExportText = " because it was incorrect in the imported workbook.";
+    const oldImportWarningText = importWarning.errorDescription;
+    const newImportWarningText =
+        oldImportWarningText
+            .substring(0, oldImportWarningText.indexOf("sheet") + 5)
+            .replace(/\bimported\b/gi, "exported") +
+        additionalExportText +
+        oldImportWarningText.substring(oldImportWarningText.indexOf("sheet") + 6);
```
dondi: This amount of string manipulation makes me wonder if there’s a more concise way to say it. Let’s discuss it when we meet

server/controllers/validators/two-column-warnings.js
```diff
@@ -216,18 +216,33 @@ const applyTwoColumnSheetWarnings = (
...
+                genesMissingValue.join(", ")
+            )
```
dondi: Per the refactor comment below, this call would now look like this:
```suggestion
            constants.warnings.missingAllGenesAndValues({
                sheetName,
                allGenesMissing: false,
                missingGenes: genesMissingValue.join(", ")
            })
```

web-client-classic/public/js/two_column_sheets_warnings.js
dondi: Post-meeting, we have converged on a satisfactory approach: instead of doing string manipulation which is brittle in its reliance on assumptions of what the original import warning text was, we will _pre-encode_ in our import warnings constants what the corresponding export warning should look like:
```javascript
        missingAllGenesAndValues: function (sheetName, isAllGenesMissing) {
            const missingType = isAllGenesMissing ? "genes and values" : "values";
            const value = `a ${valuesForEachTwoColSheet[sheetName].replace(/s$/, "")}`;

            return {
                warningCode: `MISSING_ALL_${missingType.replace(/ /g, "_").toUpperCase()}_${sheetName.toUpperCase()}`,
                errorDescription: [
                    `There were no ${missingType} supplied in the "${sheetName}" sheet in the imported workbook.`,
                    grnMapInputSuppliedWarningMessage(sheetName, value),
                ].join(" "),
                exportErrorDescription: [
                    /* Place the corresponding error message here. */
                ],
            };
        },
```

With this in place, then this code’s approach fundamentally changes: given an import warning, extract the `exportErrorDescription` supplied with it and emit the new warning object with that value as the `errorDescription`

dondi (commented):
Looks good, comprising only of changes dealing with programming technique, readability, and robustness. Tested behavior should remain the same, which is how you’ll be able to tell that your changes were implemented successfully

JustAlex3125:
@dondi The changes requested along with the changes we discussed in our 8/18 meeting should be fixed with this commit. This should now be ready for beta

dondi (approved):
LGTM—that readability refactor feels like it pays dividends right away 👌🏽

PR #1650, opened 2026-09-10
Fix Issue 1293: Gene page gives an error when right-clicking on a protein node from a PPI

pichuT:
Fixed by modifying the contextmenu in graph.js to convert the protein name to a gene name when in PPI mode. The fix removes the "p" using .replace(/p$/i, "") then passes it to the gene page URL. This only applies when grnState.mode === NETWORK_PPI_MODE so GRN networks are unaffected.


pichuT:
Pull Request Checklist
- [x] Print works as expected
- [x] Right click on a node opens gene page, page is populated with correct data (currently broken, though)

web-client-classic/public/js/graph.js
```diff
@@ -1233,7 +1233,10 @@ export var drawGraph = function (workbook) {
                     href:
                         "info?" +
                         $.param({
-                            symbol: gene.name,
+                            symbol:
+                                grnState.mode === NETWORK_PPI_MODE
+                                    ? gene.name.replace(/p$/i, "")
```
dondi, line 1238: This looks functionally right, but I think it now “replicates the knowledge” of how protein names are derived from gene names in multiple places. As you get to know the code base better, see if we can colocate this code so that the convention of appending a `p` to the name can be in one place

dondi (approved):
Locally and functionally, lgtm, but note the bigger-picture comment about colocating knowledge within the code base

